PR #3, opened 2019-01-31
Add vendor and basic quickstart guide

LiliC:
Went with dep for vendoring, but can switch to [go modules]( https://github.com/golang/go/wiki/Modules#quick-start). I tried out go modules, but for that, the project needs to be outside of the `GOPATH`. 🤷‍♀️ Up to you really.

Added basic README, as it was diffucult to figure out how to run the example. :)

Note: Would also suggest for `conprof` to create the `data` directory or pass location to one through a flag. Same for config yaml.


brancz:
Awesome! Let's figure out dep->go mods later. lgtm :+1:

PR #5, opened 2019-04-16
Add key property to elements returned by .map

metalmatze:


metalmatze:
Here @brancz, first React PR in 4 years. Have fun :smile:

brancz:
I actually wildly changed this already, these elements don’t exist anymore. Sorry 😬

PR #11, opened 2019-07-10
Rewrite storage, add service-discovery, fix query params.

cube2222:
Hey,
Firstly, I brought back the service discovery and relabelling functionality from prometheus.

Secondly, I fixed some errors, where query params (focusing on nodes in the graph) wouldn't allow for further navigation in the pprof view.

Thirdly, I've tried running conprof with ~200 target instances to be scraped once per minute (planning to do it with a few thousand). It exploded, quickly rising to tens of GB of RES memory usage. The problem was that using tsdb made it necessary to keep the profiles in memory before writing them to the append log.

I rewrote it to use a simple file based storage, where the data is streamed from the target service right into the file, reducing memory usage to ~300 MB. Querying is still snappy.


brancz:
Sorry i missed the notification for this. Thanks a lot for the hard work, i will give this a review soon! :)

cube2222:
Hey, no worries!

As a disclaimer though, I've been running this rewrite for a while on a bunch of server (a few hundred), and the querying locks are definitely suboptimal when scanning multi-hour or multi-day timespans, I'll have to revisit it, because there's no point in locking everything, if you're only looking at historical data. (Lock per shard will probably fix it)

brancz:
I was just on a flight and I was thinking we’d run into that. We don’t rely on prometheus/gorilla compression at all for sample values (the plan was to use snappy or something similar within a block), so we can just make blocks 10min long minimum that way we told max 10 samples in memory. What do you think of trying that? Generally I’d say let’s separate the discovery changes into a separate PR from storage changes. I have some time tomorrow to hack on this so I might be able to try what I said above.

cube2222:
Hey, I don't think this will work :) I think any approach based on holding byte slices in memory won't be able to scale.

I just measured it with a clean data directory, an interval of 1 minute, and doing 10 second scrapes. I got to 13 GB RES in 1 minute. (rising steadily at constant speed)

My rewrite stays around 500 MB RES indefinitely (as far as I've measured for a few minutes) in this same situation.

When I have some time, I'll do a small rewrite of the locking. The writes are no problem (~ 2000 locks per minute is nothing really). The problem is there is one global lock which is taken while it scans all directories to match labels if a big period of time is given in a query. So I'll just rewrite it to use one lock per minute-shard. This would completely remove the problems with reads influencing writes, as old shards are never written to anyways, and would also not interfere in multiple queries executing at the same time. (The go rwlock implementation doesn't allow any more readers after a writer wants to lock, when another reader has it afaik, the scraper tries to write constantly, so only one reader is able to get the lock) So this should fix any problems in my implementation.

However, no hard feelings if you feel those changes are too big for conprof, I can keep it on a fork and just PR the discovery parts and bugfixes. 😉

brancz:
The biggest reason why I'd prefer not to go with that approach is because we used to have a 1 file per series approach in Prometheus (and this is even 1 file per sample), which quickly maxed out filesystem inodes and had other problems in regards to querying. The storage as it is has proven to work incredibly well (in Prometheus), so I'd prefer to explore other solutions to the problems you brought up within the existing storage. One thing that seems worth exploring: re-writing the WAL to fix the problems you brought up as there is really no reason to keep anything but maybe labels+timestamps in memory for querying and write profiles to disk in form of a WAL, and when the WAL is flushed to an immutable block, we put it into the format we already have. Just don't want to make the same mistakes we've made before :)

brancz:
The service discovery parts are extremely awesome, please feel free to open a separate PR for it!

cube2222:
Sure, I'll prepare them when I have a moment!

Thanks for the explanation. If I have some bigger amount of time in the future, maybe I'll explore how the existing storage could be optimized. (Store currently scraped profiles like I do, and asynchronously transform them into the prometheus format maybe?) No promises though ;)

cube2222:
Done, here's the PR: https://github.com/conprof/conprof/pull/12

Closing this for now.

brancz:
> (Store currently scraped profiles like I do, and asynchronously transform them into the prometheus format maybe?) No promises though ;)

Yes I was thinking in a similar direction! :)

PR #12, opened 2019-07-23
Adding back the service discovery subsystem. Fixing a query parameters reading bug.

cube2222:


brancz:
Just tested this with kubernetes discovery. Works like a charm! Soooo awesome!!! Thanks so much for the contribution!!!

PR #13, opened 2019-07-23
deployments: Generate manifests with jsonnet

brancz:
@metalmatze


deployments/dev.jsonnet
```diff
@@ -0,0 +1,27 @@
+local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
+
+(import 'conprof/conprof.libsonnet') {
+  conprof+:: {
+    config+:: {
+      imageRepos+:: {
```
metalmatze: If we introduce local scoped config objects. Could we also use `images` which contains the full image including the tag?

brancz: very good point, let's do this, image abstractions were always a mistake

PR #27, opened 2019-10-25
Use Go 1.13 and drone CI

brancz:
@metalmatze


.drone.yml
```diff
@@ -0,0 +1,48 @@
+---
+kind: pipeline
+name: build
+
+platform:
+  os: linux
+  arch: amd64
+
```
metalmatze: You're missing `steps:`

PR #35, opened 2019-11-27
Give each scrape config its own memory

### LINKED ISSUE #33: Per-profile config doesn't work
Attempting to set some profiles on and some off results in no profiling and no diagnostic.

E.g.
```
      profiling_config:
        pprof_config:
          allocs:
            enabled: false
          block:
            enabled: false
          cmdline:
            enabled: false
          goroutine:
            enabled: true
            path:    "/debug/pprof/goroutine"
          heap:
            enabled: true
            path:    "/debug/pprof/heap"
          mutex:
            enabled: false
          profile:
            enabled: false
          threadcreate:
            enabled: false
          trace:
            enabled: false
```

(I have a PR to fix this, will post soon)

bboreham:
Fixes #33 

Because everything in the config structs is a pointer we need to take care that they don't all point to the same global variable.  It seems neatest to do this by introducing two functions whose return value
escapes.

If they all point to the same place then whatever value is read last by `UnmarshalYAML()` will apply to all of them.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -27,9 +27,13 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-var (
-	trueValue           = true
-	DefaultScrapeConfig = ScrapeConfig{
+func trueValue() *bool {
+	a := true
+	return &a
+}
+
+func DefaultScrapeConfig() ScrapeConfig {
+	return ScrapeConfig{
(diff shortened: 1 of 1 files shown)


brancz:
Nice! Thanks! :+1:

PR #38, opened 2019-11-29
Don't re-use appender after calling Commit()

### LINKED ISSUE #34: Storage usage
Hey,
I ran conprof with this config:

```yaml
scrape_configs:
  - job_name: conprof
    kubernetes_sd_configs:
      - role: pod
    relabel_configs:
      - source_labels: [__meta_kubernetes_pod_annotation_conprof_scrape]
        action: keep
        regex: true
      - source_labels:
          - __meta_kubernetes_namespace
        target_label: namespace
      - source_labels:
          - __meta_kubernetes_pod_name
        target_label: pod
      - source_labels:
          - __meta_kubernetes_pod_container_name
        target_label: container
    scrape_interval: 1m
    scrape_timeout: 1m
```

and I attached a 500Gig persistent volume which it uses to write it's data to. The config targets 10 pods and it took only 3-4 hours until the 500Gig were completely full. Is this expected / what do you recommend to configure so that conprof doesn't need that much storage?

bboreham:
The [comment on tsdb.Appender](https://github.com/conprof/conprof/blob/ccbcf2069b2abbae7ab2dcb7304c88a2389c36df/vendor/github.com/conprof/tsdb/db.go#L85) says "It must be completed with a call to Commit or Rollback and must not be reused afterwards".

Fixes #34, I believe - the reused Appender would keep resetting the `minTime` on head to its oldest value which prevents anything from being removed.

### DIFF
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -444,7 +444,6 @@ func (sl *scrapeLoop) run(interval, timeout time.Duration, errc chan<- error) {
 	}
 
 	var last time.Time
-	app := sl.appendable.Appender()
 
 	ticker := time.NewTicker(interval)
 	defer ticker.Stop()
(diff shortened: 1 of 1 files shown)


brancz:
Thanks for taking the time to do these awesome contributions! :+1: :tada:

PR #43, opened 2020-04-20
Support custom path

### LINKED ISSUE #39: How do I configure conprof to run in a different URL
I would like to proxy conprof, so that the URL on the outside is like http://example.com/conprof.

Is there something like Prometheus `--web.external-url` flag?

geobeau:
Fix #39


brancz:
Nice!!! Thank you! :tada:

PR #45, opened 2020-04-28
Add arm64 build

povilasv:
Adds drone.io arm64 build

Also fixes `strconv.Atoi(fromString)` on 32 bit systems returns ant int32, casting to `int64(from)` then looses values and doesn't work

Ref https://github.com/conprof/conprof/issues/44

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -58,12 +58,12 @@ platform:
 steps:
 - name: build
   pull: always
-  image: golang:1.13
+  image: golang:1.14
   commands:
   - make
 
 - name: tags
-  image: golang:1.13
+  image: golang:1.14
   commands:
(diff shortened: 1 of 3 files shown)


.drone.yml
```diff
@@ -47,3 +47,97 @@ steps:
...
+name: build-arm64
+
+platform:
+  os: linux
+  arch: arm64
+
+steps:
+- name: build
+  pull: always
+  image: golang:1.13
```
brancz, line 61: Any reason not to use 1.14?

.drone.yml
```diff
@@ -47,3 +47,97 @@ steps:
...
+    branch:
+    - master
+    event:
+    - push
+---
+kind: pipeline
+name: manifest
+
+platform:
+ os: linux
```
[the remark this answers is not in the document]
povilasv: Fixed

.drone.yml
povilasv: just copy paste from old version of drone.yaml

povilasv:
I've just test 1.14, all good everything seems to work :)

brancz:
Thanks! :tada:

PR #50, opened 2020-05-01
[WIP] Add scrape go trace profile as bytes

povilasv:
First part of fixing go trace profiles, we scrape trace profiles and store them as bytes directly, instead of trying to parse them as pprof.

There is some UI work left to do

Ref https://github.com/conprof/conprof/issues/46


povilasv:
closing in favor of https://github.com/conprof/conprof/pull/59

PR #54, opened 2020-05-10
WIP: Expose profiles as Prometheus remote-read

brancz:
This converts pprof profile nodes over time into time-series samples and exposes them via Prometheus remote read to query.

I believe this is primarily useful for heap and allocation profiles to see where heap and or allocations grow over time (potentially to detect memory leaks faster, but maybe it turns out to be useful for other things).

@povilasv @metalmatze


metalmatze:
Amazing how little code that is to tie it up.
LGTM. Can you give some examples of what you can do with it?

povilasv:
also what do you think if we would use git submodules for pprof package? it would be easier to sync updates from remote?

brancz:
How would that work with import paths within pprof? Those wouldn't resolve correctly or attempt to access internal of the "actual" pprof package no?

povilasv:
I think go treats it as a regular directory:
```
git submodule add  https://github.com/google/pprof.git
```

```
cd pprof
ls                  
```
``` 
appveyor.yml  CONTRIBUTING.md  doc     fuzz    go.sum    LICENSE   profile  README.md  third_party
AUTHORS       CONTRIBUTORS     driver  go.mod  internal  pprof.go  proto    test.sh
```

```
cd ../
ls
```
```
all.go  conprof      examples  LICENSE   Makefile.common  pprof        README.md   screenshot.png  VERSION
api     deployments  go.mod    main.go   manifest.tmpl    pprofui      sampler.go  scripts         web
config  Dockerfile   go.sum    Makefile  pkg              pprofui.png  scrape      vendor          web.go
```

brancz:
And you can access the `internal` package that way? If so then this is indeed awesome!

povilasv:
So I tried it, it sort of clones the directory down, but as `pprof` imports :
```
	"github.com/google/pprof/internal/plugin"
	"github.com/google/pprof/profile"
```

Dude to that it doesn't really work.

I think maybe we can add some go module magic:
```
 replace github.com/google/pprof => ./pprof
```

will test this and let you know how it goes

povilasv:
Nope doesn't seem to work. I guess copy pasting directories is easier :) 

FYI you can disable license check in Makefile.common by adding ` ! -path './pprof/*'`: 
```
	@licRes=$$(for file in $$(find . -type f -iname '*.go' ! -path './vendor/*' ! -path './pprof/*' ) ; do \
```

yeya24:
Any idea to continue this pr? I played with it and looks good. One idea about this, looks like the `name` label has high cardinalities and doesn't really fit metrics. Can we implement the Loki querier interface and use it with Loki datasource?

Screenshot from 2021-03-25 20-11-39

brancz:
The PoC works, but we found that this has very significant performance problems to the point that query_range is basically completely useless. I think we will need some more in-depth storage changes where we prepare data to be read as metrics beforehand and not just at query time.

PR #55, opened 2020-05-10
Add internal/trace from golang

povilasv:
Currently adding that package to `internal/trace`, as I feel we shouldn't be exposing others author's packages for them. But if maintainers think it's more reasonable to make it non internal, then I can change the PR :) 

The only thing I actually added is README.md, so it makes sense to only review that :)

Ref https://github.com/conprof/conprof/issues/46


povilasv:
It would be awesome to get this reviewed/merged as a lot of other PRs that I would like to open depends on this.

brancz:
Could we add some automation on updating this? Otherwise stuff like this tends to get fragile quickly.

povilasv:
:+1: I'll do something like `make sync-trace` which git pull go repo and put it into right place

brancz:
sounds good to me

povilasv:
Done :)

internal/trace/gc.go
```diff
@@ -0,0 +1,825 @@
+// Copyright 2017 The Go Authors. All rights reserved.
```
brancz, line 1: I think github has a couple of magic headers that can be used to auto hide “generated” content. Do you mind adding one of those so the diff when we sync isn’t as huge to review as it is now?

povilasv: Done, looks like it's working :)

Docs: 
https://help.github.com/en/github/administering-a-repository/customizing-how-changed-files-appear-on-github

brancz: lovely

PR #56, opened 2020-05-10
[WIP] Add traceui package from go cmd/trace

povilasv:
This is quite edited version of https://golang.org/src/cmd/trace/ HTTP server. 
I'll try to contribute my changes upstream, for now I purpose to merge this as is.

Ref: https://github.com/conprof/conprof/issues/46

depends on: https://github.com/conprof/conprof/pull/55


syepes:
What a great addition! +1

povilasv:
Now this PR builds ;)

Sadly it's only partially copied from Go's repo. I will try to commit the patches upstream, but not sure if they will accept as it works fine for them in cli tool context, but doesn't work for us.

To move forward I propose to merge as is. Thoughts?

brancz:
Have you proposed the changes to the go repo? If so then I'm not blocking this, but I want to at least get that process started.

povilasv:
Not yet, will do and let you know

PR #58, opened 2020-05-13
jsonnet: Mount configmap if used

brancz:
Tested this out with the dev.jsonnet before committing :) 

@metalmatze @povilasv


deployments/jsonnet/conprof/conprof.libsonnet
```diff
@@ -189,6 +189,25 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
...
+                  },
+                } else v,
+                super.volumes
+              ),
+          },
+        },
+      },
+    },
+
     secret:: null,
```
metalmatze, line 211: Do we still need this one?

brancz: I think it's just convenience so that if you marshal the whole thing it automatically gets excluded. No strong opinion though if you prefer to keep it.

metalmatze: Sure, no strong opinion.

PR #59, opened 2020-05-16
Add Profile Type

povilasv:
REF https://github.com/conprof/conprof/issues/46

image

as  we are using special `__name__` label for profile types everything works, with current tsdb setup

Includes:
- Added UI rendering.
- Added Trace parsing, now Traces will be scrapped correctly instead of erroring out.
- Added Seconds argument to Trace/ Profile scrape jobs, which are set to their defaults in Go:

>Profile responds with the pprof-formatted cpu profile. Profiling lasts for duration specified in seconds GET parameter, or for 30 seconds if not specified. The package initialization registers it as /debug/pprof/profile. 
>Trace responds with the execution trace in binary form. Tracing lasts for duration specified in seconds GET parameter, or for 1 second if not specified. The package initialization registers it as /debug/pprof/trace.

### DIFF
--- a/scrape/scrape.go
+++ b/scrape/scrape.go
@@ -18,23 +18,24 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"sync"
 	"time"
 
+	"github.com/conprof/conprof/config"
+	"github.com/conprof/conprof/internal/trace"
+	"github.com/conprof/tsdb/labels"
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/google/pprof/profile"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/version"
-	"golang.org/x/net/context/ctxhttp"
-
-	"github.com/conprof/conprof/config"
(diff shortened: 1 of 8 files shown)


scrape/scrape.go
```diff
@@ -378,9 +378,24 @@ func (s *targetScraper) scrape(ctx context.Context, w io.Writer) error {
 		return fmt.Errorf("server returned HTTP status %s", resp.Status)
 	}
 
-	_, err = profile.Parse(io.TeeReader(resp.Body, w))
-	if err != nil {
-		return errors.Wrap(err, "failed to parse target's pprof profile")
+	switch profileType {
+	case ProfileTraceType:
+		// This is needed as trace.Parse fails with obscure error
+		// when you pass resp.Body
```
brancz, line 384: do we know why?

povilasv: I didn't dig deeply but the error is:

```
failed to parse target's trace profile: failed to read event 3 argument at offset 0 (failed to read trace at offset 13621: read 1, error EOF)
```

brancz: Odd, let’s try to come back to it. Could you open an issue for it?

povilasv: https://github.com/conprof/conprof/issues/60

PR #62, opened 2020-05-23
update jsonnet lib version

povilasv:


deployments/jsonnetfile.json
```diff
@@ -8,6 +8,15 @@
         }
       },
       "version": ""
+    },
+    {
```
brancz, line 12: this shouldn't be here, only in the `jsonnetfile.json` that the local dependency points to

povilasv: Ah, I couldn't find the docs anywhere. Could you link me to that? I'll add a README how to set it up and generate manifests.

brancz: What exactly where you looking for in the docs? Dependencies resolve transitively, so by seeing the local dependency, `jb` will resolve the transitive dependencies by reading all dependencies from the dependencies' `jsonnetfile.json` (and then does this recursively). Which means ksonnet should already be installed if you run `jb install` in the `deployments/` directory as it reads [this jsonnetfile.json](https://github.com/conprof/conprof/blob/master/deployments/jsonnet/conprof/jsonnetfile.json).

povilasv: installing ksonnet so that our makefile picks it up, as right now I can see that our makefile looks for stuff that's in `vendor`. So how do I configure ksonnet so that it's not in vendor and imported correctly?

 Ref: https://github.com/conprof/conprof/blob/master/deployments/Makefile#L6 

Install instructions in ksonnet only say how to install their cli https://github.com/ksonnet/ksonnet/blob/master/docs/build-install.md

brancz: Aha I see!

We should add a `vendor` make target, which depends on `jsonnetfile.lock.json`, and runs `jb install` in the `deployments/` directory. That should install things as expected (and update when needed).

deployments/Makefile
```diff
@@ -1,6 +1,11 @@
 JSONNET_FMT := jsonnetfmt -n 2 --max-blank-lines 2 --string-style s --comment-style s
 
-manifests: $(shell find . -name 'vendor' -prune -o -name '*.libsonnet' -print -o -name '*.jsonnet' -print)
+.PHONY: vendor
+vendor:
+	jb install ./jsonnet/conprof/
```
brancz, line 5: just `jb install` should be sufficient

PR #71, opened 2020-09-10
Add reload config endpoint

geobeau:
Add an endpoint on the API to support reloading of the configuration.

### DIFF
--- a/all.go
+++ b/all.go
@@ -28,7 +28,7 @@ import (
 )
 
 // registerAll registers the all command.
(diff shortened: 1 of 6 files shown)


brancz:
Presumably, you have tested this? :) 

In any case, this looks great! :+1: :tada:

geobeau:
Testing? Never heard of that! (I tested that it called the reload, hence the log https://github.com/conprof/conprof/blob/master/scrape/manager.go#L106 but didn't validate that the config was indeed reloaded)

Issue #70, opened 2020-09-08
Reload config endpoint - /-/reload


syepes:
It would be great to have the possibility to reload the configuration file, just like we can do with prometheus:

`curl -X POST http://localhost:8080/-/reload`

I see we have some methods, but it's not exposed:
https://github.com/conprof/conprof/blob/master/scrape/manager.go#L84

brancz:
Agreed! If someone wants to pick this up that would be amazing!

geobeau:
I can work on it if @syepes doesn't want!

syepes:
@geobeau I had a rapid look last night but could not figure out how to add the hook, if your more familiar with the code base I'll also be interested in seeing how you implement it. :-)

geobeau:
I made #71 which is a naive implementation. @brancz @syepes any advices on how to make it more elegant?

brancz:
I merged #71, maybe @syepes can try out the image that will be built from master and we close this issue once tested? :)

syepes:
I'll give this a try with your setup

syepes:
@geobeau | @brancz 
I just have given this a test drive and there is an issue.

The reloading [gets triggered](https://github.com/conprof/conprof/blob/master/scrape/manager.go#L111) but it's reloading the current (old) config that was previously loaded on the struct, I think that before running the actual reload we need to call [ApplyConfig](https://github.com/conprof/conprof/blob/master/scrape/manager.go#L152) and then [updateTsets](https://github.com/conprof/conprof/blob/master/scrape/manager.go#L145) with the necessary parameters.

brancz:
@geobeau do you think you could have a look at that?

geobeau:
Yes, sorry for the issue

geobeau:
Made another pull request. It looks like it's working fine now

brancz:
Merged #70, @syepes can you retest with the latest build?

syepes:
@geobeau  | @brancz 
Works like a charm! thanks

PR #72, opened 2020-09-11
Fix the reloader api

geobeau:
Before the reloader only reloaded a subpart of the service.
Now it reapply the config and trigger a reload of the discovery which reloads the scrapers

#70


brancz (approved):
nice!

PR #75, opened 2020-09-25
Added quickstart bash script for easy local run.

bwplotka:
Signed-off-by: Bartlomiej Plotka <bwplotka@gmail.com>

### DIFF
--- a/examples/quickstart.sh
+++ b/examples/quickstart.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+#
+# Starts ConProf `all` mode on :8080 which collects profiles from itself and stores in ./data
+
+trap 'kill 0' SIGTERM
+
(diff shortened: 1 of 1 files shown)


metalmatze:
Sweet. Next to `make build` I thought of a `make run` that I would want to push more across various projects, that would run scripts like that. I started in Observatorium recently and would love such a concept. What do you think? Essentially asking if you want to add a `make run` target :grin:

brancz:
What speaks against just putting `go get` plus

```
conprof all --config.file examples/config.yaml
```

In the readme? I fear that the script may feel intimidating to new users, when running it is really this simple.

bwplotka:
It's not that simple - you need to know what's the data dir and port. But agree, we can put that in readme (:

PR #79, opened 2020-09-28
*: Improve split architecture

brancz:
Previously the individual components could not actually be run
individually, but it was always meant to be possible. With this slight
refactoring and introduction of gRPC services this is now possible.

@povilasv @metalmatze


brancz:
I'm gonna go ahead and merge this, as it doesn't affect the monolithic usage at all and the new components are all unit tested.

PR #88, opened 2020-10-03
Fix API response and add instrumentation

### LINKED ISSUE #85: Bug: labels API and Series API not working
```
curl localhost:10902/api/v1/labels
{"status":"success"}
```

Only status and no data returned. Seems this bug is caused by https://github.com/conprof/conprof/blob/master/api/api.go#L138. Need more investigation.

Assign myself. Will handing the response similar to Thanos https://github.com/thanos-io/thanos/blob/master/pkg/api/api.go#L219

yeya24:
Fixes #85 

Change:
1. Fix metadata API response
2. Instrument the code using Thanos instrument middlewares
3. Make all API response in Prometheus format. So the `/api/v1/query_range` result would be like below. I removed the `series` part and moved them to `data` to be unified with Prometheus response format.

```
curl 'localhost:10902/api/v1/query_range?from=1601652239723&to=1601652241412&query=allocs' | jq
{
  "status": "success",
  "data": [
    {
      "labels": {
        "__address__": "localhost:10902",
        "__name__": "allocs",
        "__scheme__": "http",
        "instance": "localhost:10902",
        "job": "conprof",
        "profile_path": "/debug/pprof/allocs"
      },
      "labelsetEncoded": "e19fYWRkcmVzc19fPSJsb2NhbGhvc3Q6MTA5MDIiLCBfX25hbWVfXz0iYWxsb2NzIiwgX19zY2hlbWVfXz0iaHR0cCIsIGluc3RhbmNlPSJsb2NhbGhvc3Q6MTA5MDIiLCBqb2I9ImNvbnByb2YiLCBwcm9maWxlX3BhdGg9Ii9kZWJ1Zy9wcHJvZi9hbGxvY3MifQ==",
      "timestamps": [
        1601652239820,
        1601652240820
      ]
    }
  ]
}
```


api/api.go
```diff
@@ -115,67 +111,13 @@ func (a *API) QueryRange(w http.ResponseWriter, r *http.Request, _ httprouter.Pa
 			level.Error(a.logger).Log("err", err, "series", ls.String())
 		}
 
-		res.Series = append(res.Series, resSeries)
+		res = append(res, resSeries)
 	}
-	// TODO(bwplotka): Handle warnings.
 	if set.Err() != nil {
```
bwplotka, line 116: the way to avoid typos here is to do

```suggestion
	if err := set.Err(); err != nil {
```
And use `err` in return (:

api/api.go
```diff
@@ -271,84 +213,91 @@ func (a *API) Series(w http.ResponseWriter, r *http.Request, _ httprouter.Params
...
 
 func (a *API) Reload(w http.ResponseWriter, r *http.Request, _ httprouter.Params) {
 	a.reloadCh <- struct{}{}
 }
+
+type ApiFunc func(w http.ResponseWriter, r *http.Request, params httprouter.Params) (interface{}, []error, *thanosapi.ApiError)
+
+// TODO: add tracer
+// Instr returns a http HandlerFunc with the instrumentation middleware.
+func GetInstr(
```
bwplotka, line 277:
```suggestion
func Instr(
```

bwplotka: No need for Get

api/api_test.go
```diff
@@ -103,11 +103,11 @@ func TestAPIQueryRangeGRPCCall(t *testing.T) {
 		t.Fatalf("Unexpected status code, expected 200, got %d", resp.StatusCode)
 	}
 
-	expectedContentType := "application/json"
-	gotContentType := resp.Header.Get("Content-Type")
-	if gotContentType != expectedContentType {
-		t.Fatalf("Unexpected Content-Type, expected %s, got %s", expectedContentType, gotContentType)
-	}
+	//expectedContentType := "application/json"
```
bwplotka, line 106: Commented code?

flags.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+func modelDuration(flags *kingpin.FlagClause) *model.Duration {
+	value := new(model.Duration)
+	flags.SetValue(value)
+
+	return value
+}
+
+func regGRPCFlags(cmd extkingpin.FlagClause) (
```
bwplotka, line 29: All of this was made public on Thanos.. so cannot we use it?

bwplotka (commented):
:+1: Just commented code, I guess some todo for this PR still?

api/api_test.go
yeya24: Will update this after I finish the tests

flags.go
yeya24: Oh, right. I missed this! Updated.

yeya24:
Unit test cases added. Should be ready for review now.

brancz:
My apologies, there is a conflict with another change now. Could you rebase this?

api/api.go
```diff
@@ -271,84 +219,91 @@ func (a *API) Series(w http.ResponseWriter, r *http.Request, _ httprouter.Params
...
+
+// TODO: add tracer
+// Instr returns a http HandlerFunc with the instrumentation middleware.
+func Instr(
+	_ log.Logger,
+	ins extpromhttp.InstrumentationMiddleware,
+) func(name string, f thanosapi.ApiFunc) httprouter.Handle {
+	instr := func(name string, f thanosapi.ApiFunc) httprouter.Handle {
+		hf := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			thanosapi.SetCORS(w)
```
yeya24, line 289: The instr method already sets CORS, do you think it is better to remove this?

brancz: yeah I don't think configuring cors belong here, let's remove it here

brancz:
I fixed the last few things that were outstanding. Thank you so much for the contribution!

PR #99, opened 2020-11-03
Fix broken storage connection

aaronjheng:
The storage API can not be connected with empty token. This should fix it.


brancz:
Thanks for reporting! We actually need both. This was superseded by https://github.com/conprof/conprof/commit/ce50636e215a57f6a2712ef0b637d2aeed91a51a

PR #105, opened 2020-11-28
remove duplicate discoveryManager

thorfour:
There was a diplicate discoveryManager that was being created and being unused.


brancz:
Whopsie

lgtm :+1:

PR #106, opened 2020-11-28
scrape manager buffer reload channel

thorfour:
This prevents the scrape reloads from being skipped by the default
case in the select statement, and only takes that branch in the event
a reload is already pending in channel.


thorfour:
```

            select {
            case m.triggerReload <- struct{}{}:
            default:
            }
```

Is where the race condition happens, if the reloader() routine isn't ready to wait on triggerReload then the `default` branch is taken and a reload never happens. 

This manifested itself in conprof coming up and never scraping from the static targets because the initial reload was ignored.

brancz:
Oh wow, great find! lgtm :+1:

PR #121, opened 2020-12-11
api: Merge profiles in batches of configurable size

brancz:


all.go
```diff
@@ -36,13 +38,39 @@ func registerAll(m map[string]setupFunc, app *kingpin.Application, name string,
 	configFile := cmd.Flag("config.file", "Config file to use.").
 		Default("conprof.yaml").String()
 	retention := extkingpin.ModelDuration(cmd.Flag("storage.tsdb.retention.time", "How long to retain raw samples on local storage. 0d - disables this retention").Default("15d"))
+	maxMergeBatchSize := cmd.Flag("max-merge-batch-size", "Bytes loaded in one batch for merging. This is to limit the amount of memory a merge query can use. Defaults to 64Mb batch size.").
```
roidelapluie, line 41: Do you want to print the default? It would be displayed by kingpin.

all.go
```diff
@@ -36,13 +38,39 @@ func registerAll(m map[string]setupFunc, app *kingpin.Application, name string,
...
+			g,
+			mux,
+			probe,
+			reg,
+			logger,
+			*storagePath,
+			*configFile,
+			*retention,
+			reloadCh,
+			*maxMergeBatchSize,
```
roidelapluie, line 56: At some point this could be passed as a struct for readability I think

api.go
```diff
@@ -39,6 +41,8 @@ func registerApi(m map[string]setupFunc, app *kingpin.Application, name string)
 
 	storeAddress := cmd.Flag("store", "Address of statically configured store.").
 		Default("127.0.0.1:10901").String()
+	maxMergeBatchSize := cmd.Flag("max-merge-batch-size", "Bytes loaded in one batch for merging. This is to limit the amount of memory a merge query can use. Defaults to 64Mb batch size.").
```
roidelapluie, line 44: same comment

api.go
```diff
@@ -39,6 +41,8 @@ func registerApi(m map[string]setupFunc, app *kingpin.Application, name string)
...
+		Default(strconv.Itoa(1024 * 1024 * 64)).Uint64()
```
roidelapluie, line 45:
```suggestion
		Default(strconv.Itoa(1024 * 1024 * 64)).Bytes()
```

brancz: Ah nice, I didn't see that one!

PR #137, opened 2021-02-01
Simply use the oklog/run SignalHandler

metalmatze:
Less code to maintain and using what upstream provides :)


brancz:
Awesome! Thank you!

PR #139, opened 2021-02-02
Create configReloaders to synchronize config for api /api/v1/status/config

metalmatze:
A lot of the imports are automatically grouped and sorted by my editor, but should follow the guidelines of this project from what I can see.

I needed to do a bunch of plumbing to pass the configReloaders through all components properly and I think that the individual components registering the `ApplyConfig` is probably cleanest in terms of dependencies.


metalmatze:
Oops. I'll fix the tests soon.

brancz (approved):
Very nice!

PR #140, opened 2021-02-02
Synchronize all reloaders for the first time

metalmatze:
Otherwise the web handler won't have the proper config to serve until a reload is triggered once.
Currently the config handler actually panics. I didn't realize during dev earlier, as I constantly reloaded :grin:


brancz:
Thank youuuuuuu

PR #141, opened 2021-02-04
Create http router for api for reusability

metalmatze:
This should make things more maintainable in the future. 
While refactoring this I discovered that I only added the `api/v1/status/config` to _web_ but not _api_...

### DIFF
--- a/all.go
+++ b/all.go
@@ -61,7 +61,7 @@ func runAll(
 	g *run.Group,
 	mux httpMux,
 	p prober.Probe,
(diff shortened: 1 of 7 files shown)


brancz:
This looks good! I do wonder. Right now, if you just run the `sampler`, you won't get this endpoint. Could we change that in a follow up?

metalmatze:
Yes, that's exactly what it was. I think if we do that, then we almost end up with `all` already. :thinking:

PR #142, opened 2021-02-05
Add /api/v1/targets API

metalmatze:
I've also restructured how the Sampler and Web component are created to allow for better extensibility in the future. This was needed to be able to pass the `scrape.Manager` to the Web -> API.

Some rough edges that I'm happy to follow up in the future.

### DIFF
--- a/all.go
+++ b/all.go
@@ -14,6 +14,8 @@
 package main
 
 import (
+	"context"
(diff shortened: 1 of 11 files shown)


brancz:
Really awesome! :+1:

PR #144, opened 2021-02-10
Make configfile optional in sample command if targets provided.

### LINKED ISSUE #143: Config file still required with --target flag
The `--config.file` flag should not be necessary when `--target` is specified on the sampler. This seems like a bug.

metalmatze:
Closes #143


brancz:
Niiiice! 

lgtm :+1:

kakkoyun:
@metalmatze Thanks for the fix :)

PR #158, opened 2021-02-26
api: add helper to skip tests if graphviz is missing

mdlayher:
Signed-off-by: Matt Layher <mdlayher@gmail.com>

I ran into this when attempting to package conprof on NixOS. The SVG renderer tests should use dot if it's available, but gracefully degrade if not.


brancz:
Thank you!

PR #167, opened 2021-03-11
Enable /api/v1/targets endpoint on sampler

metalmatze:
As per request by @lilic

### DIFF
--- a/sampler.go
+++ b/sampler.go
@@ -34,6 +34,7 @@ import (
 	"google.golang.org/grpc/credentials"
 	"gopkg.in/alecthomas/kingpin.v2"
 
+	conprofapi "github.com/conprof/conprof/api"
 	"github.com/conprof/conprof/config"
 	"github.com/conprof/conprof/pkg/store"
 	"github.com/conprof/conprof/pkg/store/storepb"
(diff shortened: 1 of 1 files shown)


sampler.go
```diff
@@ -112,21 +113,34 @@ func registerSampler(m map[string]setupFunc, app *kingpin.Application, name stri
...
 		if err := s.Run(context.TODO(), g, reloadCh); err != nil {
 			return nil, err
 		}
 
+		api := conprofapi.New(logger, reg, NopQuerier{}, reloadCh, 64*1024*1024,
+			func(ctx context.Context) conprofapi.TargetRetriever {
+				return scrapeManager
+			},
+		)
+		const apiPrefix = "/api/v1/"
```
lilic, line 141: Nitpick: Is this constant something you can define in the conprofapi package?

brancz:
Amazing! lgtm :tada: :+1:

lilic (commented):
Thanks so much, this is super helpful! 🎉

sampler.go
metalmatze: Yes. I'll take care of that along with some other things to refactor in there.

lilic: Nice! :niceee:

PR #168, opened 2021-03-11
Refactor api.New() to take options for optional config

metalmatze:
/cc @lilic 

Sadly the prefix isn't able to just be done on the outside, as the current routers don't support routing. For that I'd suggest using go-chi/chi in the future.
Still a lot cleaner already.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -78,48 +78,91 @@ type API struct {
 	maxMergeBatchSize int64
 	targets           func(context.Context) TargetRetriever
 	globalURLOptions  GlobalURLOptions
+	prefix            string
 
 	mu     sync.RWMutex
 	config *config.Config
 }
 
+type Option func(*API)
+
 func New(
 	logger log.Logger,
 	registry *prometheus.Registry,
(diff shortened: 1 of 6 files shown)


brancz:
Looks like some e2e tests are failing. Otherwise lgtm!

api/api.go
```diff
@@ -78,48 +78,91 @@ type API struct {
...
-		db:                db,
-		reloadCh:          reloadCh,
-		maxMergeBatchSize: maxMergeBatchSize,
-		targets:           targets,
+	a := &API{
+		logger:   logger,
+		registry: registry,
+		prefix:   "/api/v1/",
+		reloadCh: make(chan struct{}),
 		globalURLOptions: GlobalURLOptions{ // TODO pass into from flags
```
lilic, line 100: Side note: Ah maybe this could be a nice first issue for someone to do, what do you if we open an issue for this?

lilic (commented):
Nice!

brancz (approved):
Very nice!

PR #171, opened 2021-03-15
api: Seek to profile timestamp instead of requiring equality

brancz:
@metalmatze


metalmatze (approved):
LGTM

Just wondering: Where there cases where the timestamp wasn't specific enough?

PR #172, opened 2021-03-15
api: Pass query time range of +/- 5 minutes

brancz:
@metalmatze


metalmatze (approved):
Very nice!
At lot better with a range indeed :+1:

PR #173, opened 2021-03-17
Upgrade conprof/db to use zstd and new chunks

### LINKED ISSUE #76: tsdb: Compress Profile chunks
While during scrape we download profiles that are potentially gzipped already and write them directly to the chunk, there is huge potential to compress profiles within chunk as well. Multiple profiles around a similar time period should be very compressible.

We could try something like:

* https://facebook.github.io/zstd/
* Some snappy chunked

Definitely benchmark and we result. The fact that profiles are gzipped might be a blocker for multi-profile compression.

metalmatze:
We'll also write uncompressed raw profiles to it.

Closes #76


thorfour (approved):
hot diggity

bwplotka (approved):
Sweet! LGTM, just some bytes error still to be handled in some places 💪🏽

brancz (approved):
Nice!

PR #190, opened 2021-04-20
LazyTSDBOpener to unblock HTTP handler for metrics and profile when replaying WAL

### LINKED ISSUE #37: Start http listener before opening the DB
I restarted `conprof` and it has been silent for 25 minutes; if it was listening on `/debug/pprof` I could find out what it is doing.  Also it should respond to `/metrics`.

metalmatze:
Closes #37 
Other components might need updating as well.

cc @bboreham

### DIFF
--- a/all.go
+++ b/all.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"time"
 
+	"github.com/conprof/db/storage"
 	"github.com/conprof/db/tsdb"
 	"github.com/conprof/db/tsdb/wal"
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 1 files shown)


metalmatze:
Given that conprof uses a noopAppender while starting we actually need to profile conprof startup with another conprof.
HA pair of conprof - anyone? 😂

brancz:
We should be ok on Polar Signals to try this out as the sampler is separate from the storage (part of which would be restarting).

all.go
```diff
@@ -97,69 +101,112 @@ func runAll(
...
+			)
+			return err
+		}()
+		if err != nil {
+			level.Error(logger).Log("msg", "failed to start conprof components", "err", err)
+			os.Exit(2)
+		}
+	}()
 
 	p.Ready()
```
brancz, line 177: We should only call ready when the DB is actually open.

metalmatze: Updated! :+1:

brancz:
Generally looking good, but it looks like e2e tests are failing.

metalmatze:
The e2e test are failing due to start up of HTTP and TSDB now being concurrent...
I need to take a look again to make it somewhat predictable for the tests.

PR #202, opened 2021-06-01
Remote symbolization

brancz:
This patch adds a new gRPC service to check for existence and to upload debug symbols. These symbols are in turn uploaded to an object storage implementation in [symbolicator](https://getsentry.github.io/symbolicator/)'s ["unified" format](https://getsentry.github.io/symbolicator/api/#sources). Conprof can be configured with a symbol server URL, which will be called when there is an unsymbolized location in a pprof profile when serving that profile. The requests conform to the symbolicator JSON API.

Currently, this gRPC service was added to the `all` and `store` command because it didn't seem like it deserves its own process, but I'm happy to move it to its own process/subcommand as it's technically a distinct responsibility from actual profile storage.

Note: This uses a patched version of symbolicator right now since there is still an outstanding patch to land, but we can already go ahead with reviewing the conprof changes.

@metalmatze @thorfour


api/api.go
```diff
@@ -470,6 +482,12 @@ func (a *API) Query(r *http.Request) (interface{}, []error, *ApiError) {
 		}
 	}
 
+	// Attempt to symbolize all unsymbolized data.
+	err := a.symbolizeProfile(r.Context(), profile)
+	if apiErr != nil {
```
yeya24, line 487: `if err != nil`

brancz: Good catch!

all.go
```diff
@@ -129,13 +138,21 @@ func runAll(
 		return nil, err
 	}
 
+	var sym *symbol.Symbolizer = nil
```
thorfour, line 141: the zero value of that pointer should be nil. Unnecessary assignment.

api.go
```diff
@@ -79,15 +82,23 @@ func runApi(
 	db storage.Queryable,
 	maxMergeBatchSize int64,
 	queryTimeout model.Duration,
+	symbolServerURL string,
 ) error {
 	logger = log.With(logger, "component", "api")
 
+	var s *symbol.Symbolizer = nil
```
thorfour, line 89: another unnecessary assignment

symbol/symbolizer.go
```diff
@@ -0,0 +1,107 @@
...
+		locationAddr := fmt.Sprintf("0x%x", p.Location[locationIdx].Address)
+		locationAddrToIdx[locationAddr] = locationIdx
+		r.Stacktraces[0].Frames = append(r.Stacktraces[0].Frames, Frame{
+			InstructionAddr: locationAddr,
+		})
+	}
+
+	for _, mapping := range mappings {
+		r.Modules = append(r.Modules, Module{
+			Type:      "elf",
```
thorfour, line 70: is this safe to hardcode here?

symbol/symbolizer.go
```diff
@@ -0,0 +1,107 @@
...
+	res, err := s.c.Symbolicate(ctx, r)
+	if err != nil {
+		return err
+	}
+
+	functionIdx := map[string]int{}
+	for _, stacktrace := range res.Stacktraces {
+		for _, frame := range stacktrace.Frames {
+			var f *profile.Function
+			fIdx, ok := functionIdx[fmt.Sprintf("%s:%s", frame.Function, frame.AbsPath)]
```
thorfour, line 86: If these are strings `strings.Join` or `+` operator are faster options. Same for below

thorfour (commented):
Small nits and questions. So far looks good to me

symbol/symbolizer.go
brancz: Unfortunately, pprof does not have an indicator for what kind of binary the mapping originates from, so the idea was to start with hardcoding this for now, and figure out how to align it with pprof and or symbolicator. The options are, we either propose an extension to the format or squish it into the `comment` field, or make symbolicator auto-discover the type from the build ID if it knows it (though I'm not sure this is feasible as this is currently a required field).

brancz:
After sleeping over it, I decided that I do want the symbol management server to be a separate process. I'll get the changes ready and push them onto the PR.

all.go
```diff
@@ -118,48 +157,86 @@ func runAll(
...
+
+	confContentYaml, err := objStoreConfig.Content()
 	if err != nil {
 		return nil, err
 	}
 
-	p.Ready()
+	var symStore storepb.SymbolStoreServer
+	if len(confContentYaml) > 0 {
+		bkt, err := client.NewBucket(logger, confContentYaml, reg, comp.String())
```
metalmatze, line 208: Reading this it's not really clear what kind of bucket this is. Maybe a comment could help.

metalmatze:
Looks pretty dope! :+1: 
It would be good to document somewhere how hard of a dependency this will be for conprof. 
From my understanding this is just an additional component right?

brancz:
Fair point, I'm adding more details in flags for now, and we can add more elaborate documentation when we overhaul the documentation in general.

PR #204, opened 2021-06-02
Serve downloads directly from pprofui

brancz:
Follow up to https://github.com/conprof/conprof/issues/64

This updates the pprof dependency (both the "manual" vendor and the go mod), and removes our own custom download endpoint and now serves downloads directly from the standard pprof UI.

@metalmatze @thorfour 

cc @yeya24 @sandangel


metalmatze (approved):
Makes a lot of sense!

PR #206, opened 2021-06-04
symbol: Add minimal symbol reading gRPC API

brancz:
@metalmatze @thorfour


symbol/symbol_store.go
```diff
@@ -19,25 +19,35 @@ import (
...
+	bucket   objstore.Bucket
+	logger   log.Logger
+	bu       *binutils.Binutils
+	cacheDir string
 }
 
-func NewSymbolStore(logger log.Logger, bucket objstore.Bucket) *SymbolStore {
+func NewSymbolStore(logger log.Logger, bucket objstore.Bucket, cacheDir string) *SymbolStore {
+	bu := &binutils.Binutils{}
+	level.Debug(logger).Log("msg", "initializing binutils", "binutils", bu.String())
```
yeya24, line 45: I am not familiar with `binutils`, but what's the use case of printing `bu.String()`. It has not been initialized yet.

brancz: it's a little convoluted but `bu.String()` initializes the binutils object if it hasn't already been .. basically this prints the external tools that were discovered and will be used .. I left this here because I had trouble understanding the difference in environments from my local machine and github actions, and this is how I found it, so I felt it was useful to keep for future troubleshooting.

PR #403, opened 2021-11-02
Create pprof download of profiles

metalmatze:
Reading profiles from storage and then converting them back to pprof format and open them as download! :tada:

This needs a rebase to work with the new UUID location IDs, but should be good to go otherwise.


proto/parca/query/v1alpha1/query.proto
```diff
@@ -333,6 +340,11 @@ message QueryResponse{
     }
 }
 
+// QueryPprofResponse holds a pb.gz pprof profile for downloading
+message QueryPprofResponse {
```
brancz, line 344: The `Query` endpoint already has the ability to request the kind of response the user is requesting. I expected pprof to be just a different option there next to flamegraph/callgraph/top etc.

metalmatze: Oh, yes. That would make a lot of sense. :smile: :100:

pkg/storage/pprof.go
```diff
@@ -89,6 +89,7 @@ func generatePprof(ctx context.Context, locationStore Locations, ip InstantProfi
 		id := n.LocationID()
 		_, seenLocation := locationByID[id]
 		if !seenLocation {
+			// TODO(metalmatze): Improve this by calling once with a slice of IDs
```
kakkoyun, line 92: 👍

kakkoyun (approved):
LGTM 🚀 Awesome job!

PR #410, opened 2021-11-04
pkg/storage: Fix inlined items in flamegraph

brancz:
The `TestGenerateInlinedFunctionFlamegraph` test failed previously with the Cumulative value being 0 of the non-leaf inlined functions, causing icicle graphs to look odd whenever there was inlining happening.

@metalmatze


metalmatze:
Amazing!
Great that we also have a test for the future now! :100:

PR #424, opened 2021-11-06
Top table 305

thorfour:
Attempt at adding support for the top report type. 

TODO: needs unit tests. Wanted more input on the approach.


metalmatze:
Rebasing to current `main` should already get you the different render options for flame graphs and pprof downloads, so adding the top should be "just" another render.

pkg/query/query.go
```diff
@@ -143,16 +144,27 @@ func (q *Query) Query(ctx context.Context, req *pb.QueryRequest) (*pb.QueryRespo
...
+
 	switch req.Mode {
 	case pb.QueryRequest_MODE_SINGLE_UNSPECIFIED:
-		return q.singleRequest(ctx, req.GetSingle())
+		p, err = q.selectSingle(ctx, req.GetSingle())
 	case pb.QueryRequest_MODE_MERGE:
-		return q.mergeRequest(ctx, req.GetMerge())
+		p, err = q.mergeRequest(ctx, req.GetMerge())
 	case pb.QueryRequest_MODE_DIFF:
-		return q.diffRequest(ctx, req.GetDiff())
```
brancz, line 152: Rendering was intentionally done in these methods because for example flamegraphs need different data when diffing compared to a pprof export (which only contains one value, the diff, while diff flamegraphs consist of the "second" selection and are colored using the diff values).

The decision we were faced with was to either have these functions take care of rendering as well or have singleRequest and mergeRequest return nil "diff Profiles" to be passed to the flamegraph.

brancz (commented):
If I'm understanding the strategy correctly, I think nodes can occur multiple times in this report, which I believe is not how the pprof top report works. On a broader level I feel we should understand whether the strategy that we are exploring in https://github.com/parca-dev/parca/pull/435 is the one we will continue with long term, as that impacts the implementation of this significantly.

PR #433, opened 2021-11-08
pkg/storage/metastore: Add badger based meta store implementation

brancz:
I'm sure there are still plenty of problems with this, especially linting, but wanted to open this early. Also, I haven't actually tested this further other than the (very elaborate) unit tests.


pkg/storage/metastore/badger.go
```diff
@@ -0,0 +1,687 @@
...
+	pb "github.com/parca-dev/parca/gen/proto/go/parca/metastore/v1alpha1"
+)
+
+// BadgerMetastore is an implementation of the metastore using the badger KV
+// store.
+type BadgerMetastore struct {
+	db *badger.DB
+}
+
+var _ ProfileMetaStore = &BadgerMetastore{}
```
metalmatze, line 34: The compiler should generally be able to catch this when assigning the badger store to the metastore interface in the main.go within that switch case, shouldn't it? Generally not a fan of this pattern in Go.

brancz: Ah, this is left from the start of the implementation. I agree once it's integrated it's pointless.

brancz:
There are looooots of things that still need to be done, but I felt this was a good point to start.

Next I would move all types in the metastore to the protobuf types, and use [vtprotobuf](https://github.com/planetscale/vtprotobuf) for optimized allocations and memory pooling, as allocations continue to be the biggest source of CPU time in metadata. Also unrelated to query performance but moving to the byte slices for keys allows us to also introduce buffers for reusing key buffers.

pkg/symbolizer/symbolizer.go
```diff
@@ -30,21 +30,21 @@ import (
 
 type Symbolizer struct {
 	logger    log.Logger
-	locations metastore.LocationStore
+	metaStore metastore.ProfileMetaStore
```
metalmatze, line 33: Cosmetic: I'd prefer to have a separate interface here. It seems as if all this symbolizer needs would be:
```go
type SymbolizerMetaStore interface {
  GetSymbolizableLocations()
  Symbolize()
}
```
Happy to have that in some follow-up at some point. Just wanted to point out.

metalmatze (commented):
A few mostly cosmetic things I point out. Definitely not block if you want to merge it and then follow up with them. You decide.

pkg/symbolizer/symbolizer.go
kakkoyun: Maybe we cab also avoid stutter?

pkg/storage/metastore/badger.go
```diff
@@ -0,0 +1,687 @@
...
+		IsFolded: l.IsFolded,
+	}
+
+	if l.Mapping != nil {
+		loc.MappingId = l.Mapping.ID[:]
+	}
+
+	buf, err := proto.Marshal(loc)
+	if err != nil {
+		return uuid.UUID{}, err
```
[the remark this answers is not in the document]
brancz: Turns out the uuid package actually has a `uuid.Nil` https://pkg.go.dev/github.com/google/uuid#pkg-variables. I'll replace these as well as the emptyUUID occurrences with that!

pkg/storage/metastore/badger.go
```diff
@@ -0,0 +1,750 @@
...
+	})
+}
+
+// GetLinesByLocationIDs returns the lines for the given location IDs.
+func (m *BadgerMetastore) GetLinesByLocationIDs(ctx context.Context, ids ...uuid.UUID) (
+	map[uuid.UUID][]Line,
+	[]uuid.UUID,
+	error,
+) {
+	locations := map[uuid.UUID][]Line{}
```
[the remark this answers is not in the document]
brancz: I'm guessing this was a copy-paste mistake. Renaming to `linesByLocation`.

pkg/storage/flamegraph_flat_test.go
```diff
@@ -15,136 +15,205 @@ package storage
...
-	uuidf4 = uuid.MustParse(idf4)
-	uuidf5 = uuid.MustParse(idf5)
-)
+type LinearUUIDGenerator struct {
+	i uint64
+}
+
+func NewLinearUUIDGenerator() metastore.UUIDGenerator {
+	return &LinearUUIDGenerator{}
+}
```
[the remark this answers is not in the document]
brancz: Agreed.

pkg/symbolizer/symbolizer.go
brancz: Usually, I would agree, but the `metastore.GetSymbolizableLocations` function expects a ProfileMetaStore as the second argument, and yeah there we could trim the interface down very slightly but it's almost identical at that point. I do think we should probably abstract the metastore further and rename the current metastore implementations to "drivers", then this suggestion will make sense again I think. I'd prefer to leave that for the future though.

metalmatze (approved):

pkg/symbolizer/symbolizer.go
metalmatze: Sounds good to me!

metalmatze:
Comparing the flat storage with in-memory SQLite and badger:
```
Appends-24    18.2ms ± 1%    31.5ms ± 1%   ~     (p=0.100 n=3+3)

Appends-24    9.84MB ± 0%   13.94MB ± 0%   ~     (p=0.100 n=3+3)

Appends-24      140k ± 0%      216k ± 0%   ~     (p=0.100 n=3+3)
```

PR #435, opened 2021-11-09
pkg/storage: Implement GenerateFlamegraphFlat that works based on []Sample

metalmatze:
To get rid of the ProfileTree going forward, we need to be able to generate a flame graph based on just the samples of a profile and its locations.

More tests will be either in this PR or following PRs to match the existing `flamegraph_test.go`, up to the reviewer.

### DIFF
--- a/pkg/storage/flamegraph_flat.go
+++ b/pkg/storage/flamegraph_flat.go
@@ -1,0 +1,104 @@
+package storage
+
+import (
+	"context"
+	"fmt"
+	"sort"
+
+	"github.com/google/uuid"
+	pb "github.com/parca-dev/parca/gen/proto/go/parca/query/v1alpha1"
+	"go.opentelemetry.io/otel/trace"
+)
+
+type InstantFlatProfile interface {
+	ProfileMeta() InstantProfileMeta
+	Samples() []*Sample
+}
+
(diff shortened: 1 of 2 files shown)


pkg/storage/flamegraph_flat.go
```diff
@@ -0,0 +1,125 @@
...
+			height = int32(len(s.Location))
+		}
+
+		// Reverse walking the location as stacked location are like 3 > 2 > 1 > 0 where 0 is the root.
+		for i := len(s.Location) - 1; i >= 0; i-- {
+			location := locationsMap[s.Location[i].ID] // use the fully populated location
+			nextID := location.ID
+
+			index := sort.Search(len(cur.Children), func(i int) bool {
+				cmp := uuidCompare(uuid.MustParse(cur.Children[i].GetMeta().GetLocation().GetId()), nextID)
```
brancz, line 79: This looks very expensive but I'm happy to wait for profiling data to confirm (alternatively we can also have a second function `uuidCompareBytes`)

brancz (commented):
Happy to start building on this. One thing that we'll need to keep in mind is we'll need a second InstantFlatProfile that will provide the Diff values. Happy to discuss that further in an issue though.

PR #446, opened 2021-11-15
Break query loop when timestamp 0

metalmatze:
We had a few reports that their logs were spammed with "timestamp is 0" and to mitigate this, we can rather `break` than `continue`. The downside might be that you won't see any metrics if it's the first sample or in between, that's `0`. So far when I saw this happening it was always the last sample tough, so we can stop anyway.


metalmatze:
Supersedes https://github.com/parca-dev/parca/pull/445

PR #447, opened 2021-11-15
pkg/storage: Querying flat profiles

metalmatze:
Implement querying the new, more simple, `[]*Sample` from the QueryRange and Query.singleProfile handlers.

For now, we pass a `profileTrees` bool into some structs to decide whether or not to use the new storage or old storage way of representing trees. Eventually, we'll clean it up by removing the profileTree entirely.

### DIFF
--- a/pkg/parca/parca.go
+++ b/pkg/parca/parca.go
@@ -65,6 +65,7 @@ type Flags struct {
 
 	StorageTSDBRetentionTime    time.Duration `default:"6h" help:"How long to retain samples in storage."`
 	StorageTSDBExpensiveMetrics bool          `default:"false" help:"Enable really heavy metrics. Only do this for debugging as the metrics are slowing Parca down by a lot." hidden:"true"`
+	StorageTSDBProfileTrees     bool          `default:"true" help:"Enable profile tree storage in the TSDB" hidden:"true"`
 
 	SymbolizerDemangleMode  string `default:"simple" help:"Mode to demangle C++ symbols. Default mode is simplified: no parameters, no templates, no return type" enum:"simple,full,none,templates"`
 	SymbolizerNumberOfTries int    `default:"3" help:"Number of tries to attempt to symbolize an unsybolized location"`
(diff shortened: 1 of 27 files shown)


metalmatze:
Benchmarks are looking good. And I already know what to do next for appends to get this even lower than the previous benchmarks.

appends:
```
Appends-24    12.5ms ± 3%    16.8ms ± 1%   ~     (p=0.100 n=3+3)

Appends-24    6.26MB ± 0%    9.84MB ± 0%   ~     (p=0.100 n=3+3)

Appends-24      122k ± 0%      140k ± 0%   ~     (p=0.100 n=3+3)
```
iterators: (look at this! :tada: :rocket:)
```
Iterator-24     104µs ± 5%      62µs ± 1%   ~     (p=0.100 n=3+3)

Iterator-24    28.0kB ± 0%     0.6kB ± 0%   ~     (p=0.100 n=3+3)

Iterator-24     1.70k ± 0%     0.00k ± 0%   ~     (p=0.100 n=3+3)
```

brancz:
Do we have an idea why appends are slower? I would have expected that due to skipping trees we should be quite a bit faster?

metalmatze:
Most of the time in appends is spent in `makeStacktraceKey` and that's exactly what we're going to tackle next:
https://share.polarsignals.com/9aa74a3/

Most significantly, we didn't update the benchmarks after introducing UUIDs. So what we're really comparing above is profileTrees without UUIDs to the new flat samples with UUIDs too.
Let me provide you with a fairer comparison in a minute.

metalmatze:
tree with UUIDs vs flat with UUIDs
```
Appends-24    14.3ms ± 2%    16.8ms ± 1%   ~     (p=0.100 n=3+3)

Appends-24    8.98MB ± 0%    9.84MB ± 0%   ~     (p=0.100 n=3+3)

Appends-24      122k ± 0%      140k ± 0%   ~     (p=0.100 n=3+3)
```

brancz (approved):
lgtm nice work!

PR #453, opened 2021-11-22
*: Move metastore to proto types

brancz:
Even with a massive diff this change just changes the core metastore types to be protobuf based. With the amount of work that has now gone into the badger-based storage, I also made it the default now.


env-proto.sh
```diff
@@ -17,9 +17,9 @@ go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.27.1
...
-VERSION="0.51.1" && \
+VERSION="1.0.0-rc8" && \
 BINARY_NAME="buf" && \
-  curl -sSL \
+  sudo curl -sSL \
     "https://github.com/bufbuild/buf/releases/download/v${VERSION}/${BINARY_NAME}-$(uname -s)-$(uname -m)" \
     -o "${BIN_DIR}/${BINARY_NAME}" && \
-  chmod +x "${BIN_DIR}/${BINARY_NAME}"
\ No newline at end of file
+  sudo chmod +x "${BIN_DIR}/${BINARY_NAME}"
```
metalmatze: Is there really no way of doing this without calling `sudo` from a build script? :flushed: :thinking:

pkg/storage/flamegraph.go
```diff
@@ -166,7 +168,7 @@ func GenerateFlamegraph(
 		if it.NextChild() {
 			child := it.At()
 			id := child.LocationID()
-			l, found := locs[id]
+			l, found := locs[string(id[:])]
```
kakkoyun, line 171: It would be nice to have comments for these types of voodoo magics.

env-proto.sh
brancz: My bad, I think this was just something I had to do in my env, don't think this is universally necessary.

metalmatze: Ok! :blush: Thanks!

pkg/storage/flamegraph.go
metalmatze: It's going to be ALL across Parca for the time until we can actually move to [16]byte for the UUIDs across the board, I think.

brancz: I didn't want to conflate the PR further, I was thinking of subsequently moving the `.LocationID()` functions to be plain `[]byte`. What's happening here is the `uuid.UUID`, which is a `[16]byte`, is being converted to a plain `[]byte`. Since this will be in a quick follow up, would it be ok with you to skip it?

kakkoyun: Yes, totally ok.

kakkoyun (approved):

PR #462, opened 2021-11-23
*: Use vtprotobuf for Go protobuf generation

brancz:


metalmatze (approved):
Simply and straight forward :100:

metalmatze:
Benchmarks are looking good:

```
Appends-24    21.5ms ± 3%    20.4ms ± 0%  -5.11%  (p=0.024 n=6+3)

Appends-24    10.3MB ± 0%    10.2MB ± 0%  -1.13%  (p=0.024 n=6+3)

Appends-24      137k ± 0%      137k ± 0%    ~     (p=1.000 n=6+3)
```
```
Iterator-24    66.9µs ± 9%    62.1µs ± 5%   ~     (p=0.200 n=3+3)

Iterator-24      496B ± 1%      492B ± 0%   ~     (p=0.400 n=3+3)

Iterator-24      3.00 ± 0%      3.00 ± 0%   ~     (all equal)
```

PR #465, opened 2021-11-24
pkg/storage: Implement diffs for flat profiles

metalmatze:
Comparing apples and oranges once more, but diffing for flat profiles is looking good:
```
name            old time/op    new time/op    delta
Diff/simple-24     118µs ± 8%       5µs ±10%  -95.72%  (p=0.000 n=10+10)

name            old alloc/op   new alloc/op   delta
Diff/simple-24     117kB ± 0%       2kB ± 0%  -98.04%  (p=0.000 n=10+10)

name            old allocs/op  new allocs/op  delta
Diff/simple-24     2.43k ± 0%     0.01k ± 0%  -99.79%  (p=0.000 n=10+10)
```

### DIFF
--- a/pkg/storage/diff.go
+++ b/pkg/storage/diff.go
@@ -56,7 +56,21 @@ func (d *DiffProfile) ProfileMeta() InstantProfileMeta {
 }
 
 func (d *DiffProfile) Samples() map[string]*Sample {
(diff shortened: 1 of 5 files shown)


brancz (approved):
Woah nice!

PR #466, opened 2021-11-24
pkg/storage: Use UUIDs for stack traces

metalmatze:
Stack traces, or samples, are now using UUIDs as keys. The badger store (and other SQL stores to be implemented) are storing the actual information for the stack traces and can be retrieved by its UUID.

Appends are looking slightly worse but storing the stack traces in the metastore will be important for persistence in the future.
```
Appends-24    21.0ms ± 1%    28.3ms ± 2%  +34.75%  (p=0.008 n=5+5)

Appends-24    10.2MB ± 0%    11.3MB ± 0%  +10.66%  (p=0.008 n=5+5)

Appends-24      137k ± 0%      162k ± 0%  +18.21%  (p=0.008 n=5+5)
```

Interestingly enough CPU time got better for querying yet memory got worse. 
```
Iterator-24    66.9µs ± 9%    62.3µs ±13%     ~     (p=0.250 n=3+5)

Iterator-24      496B ± 1%      672B ± 5%  +35.53%  (p=0.036 n=3+5)

Iterator-24      3.00 ± 0%      4.00 ± 0%  +33.33%  (p=0.036 n=3+5)
```

Follow up: 
Implement `GetStacktraceByID` and remove the location `map[[16]byte][]*metastore.Location` from the series and instead use the metastore's data.


pkg/storage/interface.go
```diff
@@ -138,7 +138,7 @@ func CopyInstantProfileTree(pt InstantProfileTree) *ProfileTree {
 type InstantProfile interface {
 	ProfileTree() InstantProfileTree
 	ProfileMeta() InstantProfileMeta
-	Samples() map[string]*Sample
+	Samples() map[[16]byte]*Sample
```
brancz, line 141: I prefer this to be a string. A copy happens anyways, so `[16]byte` is no better in my opinion and locks the interface into having to be 16 bytes.

metalmatze: We now could also have this be a `Samples() map[uuid.UUID]*Sample` too, what about that one? 
Happy to use the string nonetheless. I guess in the case of a string we don't want to use `uuid.String()` still but cast like before?

brancz: Yeah, I think using the opaque `[]byte` (and consequently `string` as map keys, this is a bit of a hack but allows for various proven optimization techniques) would be best as we're not locking ourselves into a specific mechanism or length.

PR #467, opened 2021-11-24
Handling of version (ui, --version, removal of date+builtBy)

fpuc:
UI: The build version is listed in the Header

image
image

CLI: Added flag `--version` to dump version information

### DIFF
--- a/cmd/parca/main.go
+++ b/cmd/parca/main.go
@@ -28,8 +28,6 @@ import (
 var (
 	version string = "dev"
 	commit  string = "dev"
-	date    string = "dev"
-	builtBy string = "dev"
 )
 
 func main() {
@@ -50,8 +48,6 @@ func main() {
 	level.Debug(logger).Log("msg", "parca initialized",
 		"version", version,
 		"commit", commit,
-		"date", date,
-		"builtBy", builtBy,
 		"config", fmt.Sprint(flags),
 	)
 
--- a/cmd/parca/main.go
+++ b/cmd/parca/main.go
@@ -26,16 +26,22 @@ import (
 )
 
 var (
-	version string
-	commit  string
-	date    string
-	builtBy string
+	version string = "dev"
+	commit  string = "dev"
+	date    string = "dev"
+	builtBy string = "dev"
 )
 
 func main() {
 	ctx := context.Background()
 	flags := &parca.Flags{}
+
 	kong.Parse(flags)
+
+	if flags.Version {
+		fmt.Printf("parca, version %s (commit: %s)\n", version, commit)
+		return
+	}
 
 	serverStr := figure.NewColorFigure("Parca", "roman", "cyan", true)
 	serverStr.Print()
@@ -51,7 +57,8 @@ func main() {
 
 	registry := prometheus.NewRegistry()
 
-	err := parca.Run(ctx, logger, registry, flags)
+	err := parca.Run(ctx, logger, registry, flags, version)
+
 	if err != nil {
 		level.Error(logger).Log("msg", "Program exited with error", "err", err)
 		os.Exit(1)
--- a/pkg/parca/parca.go
+++ b/pkg/parca/parca.go
@@ -62,6 +62,7 @@ type Flags struct {
 	Port               string   `default:":7070" help:"Port string for server"`
 	CORSAllowedOrigins []string `help:"Allowed CORS origins."`
 	OTLPAddress        string   `help:"OpenTelemetry collector address to send traces to."`
+	Version            bool     `help:"Show application version."`
 
 	StorageTSDBRetentionTime    time.Duration `default:"6h" help:"How long to retain samples in storage."`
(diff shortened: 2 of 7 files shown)


pkg/server/server.go
```diff
@@ -150,16 +154,70 @@ func (s *Server) ListenAndServe(ctx context.Context, logger log.Logger, port str
...
+
+		paths := []string{fmt.Sprintf("/%s", path)}
+
+		if paths[0] == "/index.html" {
+			paths = append(paths, "/")
+		}
+
+		for _, path := range paths {
+			uiHandler.HandleFunc(path, func(w http.ResponseWriter, r *http.Request) {
+				http.ServeContent(w, r, d.Name(), time.Now(), bytes.NewReader(outputBytes)) // TODO(modtime) ? buildtime
```
fpuc, line 205: here is one todo left -> we could pass down the time of the build here, for client caching.

metalmatze: Nice idea, however, I'd actually be in favor of removing build time altogether. 
Build times aren't good for reproducible builds.
I've opened an issue here: https://github.com/parca-dev/parca/issues/468

fpuc: :+1: for pointing that out - dropped also the `date` from the `--version` flag

fpuc:
Also added another commit targeting #468 PTAL

pkg/parca/parca.go
```diff
@@ -74,7 +75,7 @@ type Flags struct {
 }
 
 // Run the parca server
-func Run(ctx context.Context, logger log.Logger, reg *prometheus.Registry, flags *Flags) error {
+func Run(ctx context.Context, logger log.Logger, reg *prometheus.Registry, flags *Flags, uiData map[string]string) error {
```
brancz, line 78: Any reason why we wouldn't make `uiData` a struct?

brancz (commented):
This is super awesome!

pkg/parca/parca.go
fpuc: agree, let's go with type safety -> changed

cmd/parca/main.go
```diff
@@ -22,36 +22,42 @@ import (
...
 		"commit", commit,
-		"date", date,
-		"builtBy", builtBy,
 		"config", fmt.Sprint(flags),
 	)
 
 	registry := prometheus.NewRegistry()
 
-	err := parca.Run(ctx, logger, registry, flags)
+	err := parca.Run(ctx, logger, registry, flags, &server.UIData{
```
brancz, line 57: My apologies, I wasn't clear in my previous comment, I also meant that at this point it shouldn't even be "UIData" but just version information, and where rendering happens it can be transformed into a different datastructure (that currently happens to be identical to a "VersionInfo" struct for example).

fpuc: Ah thanks - thats a good idea, changed

fpuc:
PTAL

pkg/parca/parca_test.go
```diff
@@ -34,7 +34,7 @@ func benchmarkSetup(ctx context.Context, b *testing.B) (pb.ProfileStoreServiceCl
 	done := make(chan struct{})
 	go func() {
 		defer close(done)
-		err := Run(ctx, logger, reg, &Flags{ConfigPath: "testdata/parca.yaml", Port: ":9090"})
+		err := Run(ctx, logger, reg, &Flags{ConfigPath: "testdata/parca.yaml", Port: ":9090"}, nil)
```
brancz, line 37: We can't pass `nil` here. Let's just do `test-version` or something like that.

brancz (commented):
Other than the compile error, this lgtm! Thank you so much for taking care of this!

pkg/server/server.go
```diff
@@ -150,16 +154,80 @@ func (s *Server) ListenAndServe(ctx context.Context, logger log.Logger, port str
...
 
+	// Strip the subpath
 	uiFS, err := fs.Sub(ui.FS, "packages/app/web/dist")
 	if err != nil {
 		return fmt.Errorf("failed to initialize UI filesystem: %w", err)
 	}
 
+	// There is currently no way to go between `http.FileServer(http.FS(uiFS))` and execute
+	// templates an http.ServeMux with registered paths and executes templates is used.
+	uiHandler := http.ServeMux{}
```
kakkoyun, line 165: I think we should create a new method to init `uiHandler` and extract the logic there.

kakkoyun (commented):
Awesome work!
Commented.

pkg/parca/parca_test.go
fpuc: :heavy_check_mark:  TY

pkg/server/server.go
fpuc: :+1: changed

fpuc:
ready for next round :-)

brancz (approved):
Love it! Giving @kakkoyun last call.

kakkoyun (approved):
LGTM. Thanks for the contribution.

metalmatze:
Thanks @fpuc :tada:

PR #469, opened 2021-11-25
pkg/storage: Create GenerateFlatPprof to download pprof profiles

metalmatze:
Making pprof downloads with the new flat profiles work.


brancz (approved):
Nice!

PR #478, opened 2021-12-02
pkg/storage: Properly inline functions in flat flame graphs

metalmatze:
fixed

So far I also cannot see any gaps in any profiles showing up anymore.

### DIFF
--- a/pkg/storage/flamegraph_flat.go
+++ b/pkg/storage/flamegraph_flat.go
@@ -69,8 +69,6 @@ func GenerateFlamegraphFlat(ctx context.Context, tracer trace.Tracer, metaStore 
 	}
 
 	for _, s := range p.Samples() {
-		var child *pb.FlamegraphNode
-
 		if int32(len(s.Location)) > height {
 			height = int32(len(s.Location))
 		}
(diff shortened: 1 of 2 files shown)


pkg/storage/flamegraph_flat.go
```diff
@@ -86,32 +84,54 @@ func GenerateFlamegraphFlat(ctx context.Context, tracer trace.Tracer, metaStore
 			})
 
 			if index < len(cur.Children) && bytes.Equal(cur.Children[index].GetMeta().GetLocation().GetId(), nextID[:]) {
-				child = cur.Children[index]
+				// The node already exists in the flamegraph, we can simply add the value and diff value and continue with this child's children next.
```
brancz, line 87: I think this still misses the case where we have two stack traces that have the same inlined location in the "middle" of the stack but different leafs. Because the second stack added will not add the cumulative to all of the functions in the flamegraph that belong to this location.

metalmatze: Makes sense. I'll do that one as a follow-up if that's ok?

brancz: Since this is behind a feature flag, I'm happy with that.

brancz (approved):

brancz (approved):

Issue #464, opened 2021-11-24
Versioned metastore keys


brancz:
@kakkoyun correctly mentioned that it's possible that our understanding of what makes a record unique might change over time, so our keys for locations, functions and mappins should be verioned. Eg. these prefixes should contain a `v1` versioned string somewhere: 

https://github.com/parca-dev/parca/blob/c086e4d35270f1fd0903f93f0842f3d8737e7131/pkg/storage/metastore/kv.go#L22

cbrgm:
Merged in #484  🎉

brancz:
Thank you very much @cbrgm! :tada:

PR #484, opened 2021-12-07
Use versioned metastore keys (#464)

cbrgm:
This PR will make sure keys for locations, functions and mappings are versioned.  `locationsKeyPrefix` as an example will now be `v1/locations/by-key/` instead of `locations/by-key/` with a version value of `v1`.

Related to #464 (Versioned metastore keys)


pkg/storage/metastore/kv.go
```diff
@@ -15,15 +15,29 @@ package metastore
...
+	v1prefix = "v1"
+)
 
-const stacktraceIDPrefix = "stacktrace/by-id/"
+const (
+	//const stacktraceKey = "stacktrace/by-key/"
+	stacktraceID = "stacktrace/by-id/"
+	locationsKey = "locations/by-key/"
+	functionKey  = "functions/by-key/"
+	mappingKey   = "mappings/by-key/"
```
metalmatze, line 32: Since we only have one version we don't need to compute these really. It's mostly a matter of figuring out the layout of the key:

Should we either pick
```suggestion
	stacktraceID = "stacktrace/by-id/v1"
	locationsKey = "locations/by-key/v1"
	functionKey  = "functions/by-key/v1"
	mappingKey   = "mappings/by-key/v1"
```
or
```suggestion
	stacktraceID = "v1/stacktrace/by-id/"
	locationsKey = "v1/locations/by-key/"
	functionKey  = "v1/functions/by-key/"
	mappingKey   = "v1/mappings/by-key/"
```

What do you think?

metalmatze (changes requested):
Thank you for that amazing contribution! :tada: 
While we can compute these keys, I don't think it's necessary and we simply should figure out how to version this going forward.

I'm curious to learn what people think about one or the other.

pkg/storage/metastore/kv.go
brancz: I prefer the second option

metalmatze (approved):
:100:

metalmatze:
Oops, seems like you need to update some test cases.
https://github.com/parca-dev/parca/blob/main/pkg/storage/metastore/metastore_test.go#L26-L228
This should work by copying the printed bytes from the failed test into the test case.

cbrgm:
Sorry, I have adjusted the test cases and ran the make target `make go/test` once again locally. Should work now.

brancz:
Nice! Thank you for taking care of this! :tada:

PR #485, opened 2021-12-08
Correct inline existing functions in flamegraphs

metalmatze:
The key difference to the previous implementation now is, that we always get the nodes for a location first.
Then, then we loop over these and in turn also inline the functions properly. That way handling the inlined functions that aren't yet part of the flame graph are exactly the same as handling inline functions that already exist in the flame graph.


brancz (approved):
Nice :100: this was a tough one, nice!!

Issue #488, opened 2021-12-09
Use pprof labels in flame graphs


metalmatze:
We don't really do anything with pprof labels in the UI right now. We only use them to create unique stacktrace IDs.
What should we do with them in the UI?

brancz:
I don't think I've seen any other flamegraph implementations with label support. I think they are intended to be aggregated away. If we can come up with a useful way to visualize them though, I'm all for it!

PR #489, opened 2021-12-09
pkg/storage: Use metastore for storing and retreiving stacktraces by UUID

metalmatze:
This should allow us to remove the labels from the storage entirely going forward, where they don't really belong anyway.
In the end we don't even use the labels in the flame graph, maybe we should start doing something with them? https://github.com/parca-dev/parca/issues/488

Appends:
```
Appends-24    28.3ms ± 2%    27.7ms ± 2%    ~     (p=0.056 n=5+5)

Appends-24    11.3MB ± 0%    11.4MB ± 0%  +0.55%  (p=0.008 n=5+5)

Appends-24      162k ± 0%      164k ± 0%  +1.59%  (p=0.008 n=5+5)
```
Iterator:
```
Iterator-24    62.3µs ±13%    73.8µs ±15%  +18.51%  (p=0.032 n=5+5)

Iterator-24      672B ± 5%      765B ± 7%  +13.84%  (p=0.008 n=5+5)

Iterator-24      4.00 ± 0%      4.00 ± 0%     ~     (all equal)
```

### DIFF
--- a/pkg/storage/series_iterator.go
+++ b/pkg/storage/series_iterator.go
@@ -394,11 +394,7 @@ func (m MemSeriesInstantFlatProfile) Samples() map[string]*Sample {
 	samples := make(map[string]*Sample, len(m.sampleIterators))
 	for k, it := range m.sampleIterators {
 		samples[k] = &Sample{
-			Value:    it.At(),
-			Location: m.locations[k],
-			Label:    nil, // TODO
-			NumLabel: nil, // TODO
-			NumUnit:  nil, // TODO
+			Value: it.At(),
 		}
 	}
 	return samples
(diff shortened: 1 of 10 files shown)


pkg/storage/series_iterator.go
```diff
@@ -394,11 +394,7 @@ func (m MemSeriesInstantFlatProfile) Samples() map[string]*Sample {
 	samples := make(map[string]*Sample, len(m.sampleIterators))
 	for k, it := range m.sampleIterators {
 		samples[k] = &Sample{
-			Value:    it.At(),
-			Location: m.locations[k],
-			Label:    nil, // TODO
-			NumLabel: nil, // TODO
-			NumUnit:  nil, // TODO
```
metalmatze, line 401: These are now populated from the meta store.

brancz:
lgtm nice!

PR #493, opened 2021-12-13
*: Fix env.sh when installing minikube

derekparker:
There was a `mv` command omitted in the script which caused it to try
running the `minikube-linux-amd64` binary instead of moving it into
$PATH. This caused an error because the binary would not be found.

Add `mv` command to fix this.


metalmatze (approved):

kakkoyun (approved):
:100:

brancz (approved):

PR #494, opened 2021-12-13
*: Specify shell in Makefile

derekparker:
The default shell Make will try to use in /bin/sh which does not
implement the `source` command, causing make targets such as `dev/up` to
fail.


kakkoyun (approved):
:tada: :rocket:

brancz (approved):

PR #495, opened 2021-12-15
Tailwind v3 upgrade

manojVivek:
This PR adds a commit on top of #491, to port the `tailwind.config.js` to eliminate a deprecation warning during the startup.

### DIFF
--- a/ui/packages/app/web/tailwind.config.js
+++ b/ui/packages/app/web/tailwind.config.js
@@ -1,8 +1,5 @@
 module.exports = {
-  purge: {
-    enabled: process.env.NODE_ENV === 'production',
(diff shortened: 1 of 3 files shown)


ui/packages/app/web/tailwind.config.js
```diff
@@ -1,8 +1,5 @@
 module.exports = {
-  purge: {
-    enabled: process.env.NODE_ENV === 'production',
```
manojVivek, line 3: The `enabled` option is removed from the config schema as v3, as tailwind is not using `PurgeCSS` any more. 
Ref: https://tailwindcss.com/docs/upgrade-guide#configure-content-sources

brancz (approved):
Nice!

PR #496, opened 2021-12-15
area/ui: Unify value formatting

### LINKED ISSUE #409: Unify number formatting
We currently use two separate mechanisms to format numbers with SI units:

* [this](https://github.com/parca-dev/parca/blob/4327220e84eb14db4162f479c91f5506a4fd2340/ui/packages/shared/profile/src/IcicleGraph.tsx#L463-L465) one in use by the IcicleGraph hover tooltip
* [this](https://github.com/parca-dev/parca/blob/main/ui/packages/app/web/src/libs/unit.ts) one in use by the MetricsGraph axis and tooltip

yomete:
This PR unifies the two separate mechanisms we use to format numbers right now. The single function for formatting numbers can now be found in `@parca/functions` and currently allows for formatting values in bytes, count, and time.

This also fixes the CPU nanoseconds value not being formatted when hovering over an icicle graph item.

Resolves #409


brancz:
:racing_car: 

lgtm! :tada:

PR #500, opened 2021-12-16
ui: Fix undefined format and return unformatted number as string

metalmatze:
Please take a look if this is a good fix, @yomete.


yomete (approved):
LGTM! thanks for the fix.

PR #501, opened 2021-12-17
Added 'assetPrefix' and a corresponding handler in the go server to allow the app to be accessed over an URL prefix

manojVivek:


brancz:
Nice! @yomete will this still work together with having multiple nextjs pages?

yomete:
> Nice! @yomete will this still work together with having multiple nextjs pages?

I just opened a PR for this [here](https://github.com/parca-dev/parca/pull/508) actually

PR #507, opened 2021-12-21
pkg/storage: Fix bug so samples seek to correct index

metalmatze:
We currently don't seek the correct index for samples in chunks. 
This has the effect that we basically every only read the first sample of each chunk - resulting in not updating icicle graphs until a timestamp in another chunk is selected.

I was already wondering why this was happening in the UI... :man_shrugging: :smile:


brancz (approved):
Whops

PR #508, opened 2021-12-21
area/ui: Add additional Next.js pages to the binary pkg

yomete:
This is a proof of concept for how we can we can create additional Next.js pages and embed them in the built binary package.

After the page has been created in the Next.js app and exported using `make ui`, there should be a directory created for the page with a corresponding `index.html` file.

Finally, the path to the page's directory is added to the embed directive in `ui.go` and the `pkg/server/server.go` file.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -252,6 +252,10 @@ func (s *Server) uiHandler(uiFS fs.FS) (*http.ServeMux, error) {
 			paths = append(paths, "/")
 		}
 
+		if paths[0] == "/table/index.html" {
+			paths = append(paths, "/table")
+		}
+
 		for _, path := range paths {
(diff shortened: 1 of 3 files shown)


metalmatze:
The reason why we need these different `index.html` per sub-folder is because these have their specific props per page like:
```html
<script id="__NEXT_DATA__" type="application/json">{"props":{"pageProps":{}},"page":"/","query":{},"buildId":"static","nextExport":true,"autoExport":true,"isFallback":false,"scriptLoader":[]}</script>
```

So, I'm thinking maybe it might even be possible to generate this JSON object with the per page props on the fly in the Go backend. 
It would be great if we could find the code from NextJS on how they generate this on their backend.

PR #525, opened 2022-01-06
*: Extract profile normalization into separate package

brancz:
This code really has nothing to do with the storage, so it should be
separate from it.

@metalmatze @thorfour


thorfour:
LGTM

PR #528, opened 2022-01-06
Update all Go dependencies to latest versions

metalmatze:
This updates all our Go dependencies to their latest versions, not only patch version but also minor and some even major version updates.
Prometheus is updated to v2.32.1.


kakkoyun (approved):
LGTM

metalmatze:
I'm fixing those unit tests.

PR #537, opened 2022-01-10
Add version and commit to Parca build in container

metalmatze:
We added the version to binaries compiled outside containers but were still missing the version and commit for the container binaries.
This PR add VERSION and COMMIT to Parca so that it shows up in the UI too.


kakkoyun (approved):
LGTM

PR #543, opened 2022-01-12
Initialize Lerna in UI directory

yomete:
This PR initializes Lerna in the UI directory which will allow us to publish the Parca UI components. I've also gone ahead to publish the components:

[@parca/parser](https://www.npmjs.com/package/@parca/parser)
[@parca/dynamicsize](https://www.npmjs.com/package/@parca/dynamicsize)
[@parca/client](https://www.npmjs.com/package/@parca/client)
[@parca/icons](https://www.npmjs.com/package/@parca/icons)
[@parca/profile](https://www.npmjs.com/package/@parca/profile)

Next steps will be to edit the `README.md` files in each pkg with proper documentation and also automate publishing to npm after a PR merge using GitHub actions.


brancz:
I'm curious, would it make sense to align the module versions with the Parca version? I fear we might create confusion otherwise. Thoughts?

yomete:
I could reinitialize Lerna and set its initial version to be `0.7.0` (which is the latest tag for Parca) and the components' versions should begin from `0.7.1` going forward. But because we want to do a publish to npm every time we update any of the components, there's a chance the versions will be different at some point in time. Which isn't bad imo.
 
Unless we want to change that and rather include the publishing in the general Parca release that's done manually.

brancz:
Got it, yeah I'm happy with that!

ui/lerna.json
```diff
@@ -0,0 +1,13 @@
+{
+  "version": "0.1.2",
+  "npmClient": "yarn",
+  "tagVersionPrefix": "v-ui",
+  "packages": [
```
manojVivek, line 5: Should we try `useWorkspaces: true` here so that it uses the existing yarn workspace and its config? 

If we do that, then we might have to add `private: true` to the `app/*` packages' `package.json`, but it shouldn't be a problem, I guess.

manojVivek:
Looking great otherwise! 👍🏼

ui/lerna.json
yomete: `useWorkspaces: true` was initially part of the options but I had to remove it because Lerna also published everything in the "packages/app/*" dir.

yomete:
> Got it, yeah I'm happy with that!

Perfect! I might have to close this PR and create another but i'll make that change

ui/lerna.json
manojVivek: Yeah it would, but `private: true` on those packages should prevent that.

manojVivek: Also, if we don't include `app/*` packages in `lerna`'s config, then the version bump of the packages in `shared/*` won't get updated on `app/*` projects' dependencies, I guess. Was that bit working fine?

Just curious, I could be wrong too.

yomete: Interesting. I\ll give that a try and let you know

PR #545, opened 2022-01-12
Start a new top table report type

metalmatze:
We need to follow up in a couple of places:
* [x] Make sure to take all lines into account (inline functions)
* [x] Merge nodes/samples by their function name if location addresses match (in one example lines went from ~4600 => ~375)

Other than that we need to obviously build the UI parts :)


pkg/query/query.go
```diff
@@ -312,6 +312,14 @@ func (q *Query) renderReport(ctx context.Context, p profile.InstantProfile, typ
 		return &pb.QueryResponse{
 			Report: &pb.QueryResponse_Pprof{Pprof: buf.Bytes()},
 		}, nil
+	case pb.QueryRequest_REPORT_TYPE_TOP:
+		top, err := GenerateTopTable(ctx, q.metaStore, p)
```
brancz, line 316: Where is `GenerateTopTable` defined? I can't find it? :thinking:

metalmatze: Seems like I missed to commit the untracked files :see_no_evil:

brancz:
Is this ready for review?

metalmatze:
We still have a few TODOs left to do from the backend and UI side. If you want you can obviously give feedback already! :relaxed:

brancz:
I'll hold off until you deem it ready :) let me know when that's the case, looking forward to this!

brancz:
What's the status on this?

metalmatze:
I need to finish the aggregation work. I wanted to do this in a stream but didn't end up streaming. I'll finish it regardless then.

yomete:
I just pushed my latest UI changes to make the table look nice. it has all the agreed features e.g sorting. @metalmatze it'd be nice to also receive the units from the backend so we can use that to format the numbers in the table.

Lastly, did we say we wanted to have additional columns for `Flat%` and `Cum%`? Let me know so i can add the columns and what their keys will be :)

metalmatze:
Very nice! 
I'll work on the backend side to finish this up. 
Let's make this PR work as is and then follow up with another PR to add the percentages if we think it's better with than without?

yomete:
> Very nice! I'll work on the backend side to finish this up. Let's make this PR work as is and then follow up with another PR to add the percentages if we think it's better with than without?

sounds good!

metalmatze:
Another round of reviews would be perfect now! :relaxed:

brancz (approved):
Amazing!!

PR #558, opened 2022-01-17
Initialize Lerna in UI directory

yomete:
This PR initializes Lerna in the UI directory which will allow us to publish the Parca UI components. I've also gone ahead to publish the components:

[@parca/parser](https://www.npmjs.com/package/@parca/parser)
[@parca/dynamicsize](https://www.npmjs.com/package/@parca/dynamicsize)
[@parca/client](https://www.npmjs.com/package/@parca/client)
[@parca/icons](https://www.npmjs.com/package/@parca/icons)
[@parca/profile](https://www.npmjs.com/package/@parca/profile)

I also modified the initial version for all pkgs to be `0.7.x` so as to be on par with the main Parca tags/releases.

Next steps will be to edit the README.md files in each pkg with proper documentation <del>and also automate publishing to npm after a PR merge using GitHub actions.</del>

<del>I had to do a name change for the shared packages (parca -> parcaui) because I botched up setting up our organization on npm 😓 Apparently deleted organization names cannot be retrieved. So all of our UI components are published under the `parcaui` and the actual app(s) is still under @parca/web which is fine since we don't intend on publishing that.</del>

<del>P.S I already reached out to npm's support team to see if we can get the parca organization name back. If this is successful, I'll create another PR to change the names back.</del>


yomete:
okay, @brancz @manojVivek this is ready now.

brancz:
Something that we always tried to avoid as much as possible is having to do any kind of changes when publishing releases, so I wonder if there is any way that we could read the version to publish from the "current" git tag instead of having to specify it in the package.json before every release?

yomete:
When we publish any pkg to the npm registry, Lerna automatically goes into all of our packages (including @parca/web) that has that newly published pkg as a dependency and modifies the package.json file with the newly updated version of that pkg.

So we shouldn't have to specify it in the package.json before every release. Me setting the initial version for all pkgs to be 0.7.x so as to be on par with the main Parca tags/releases was a one time setup thing.

 I'm also going to setup the automation so that it's literally just one command to publish.

I tried to investigate to see if we could setup Lerna to read the version to publish from the "current" Parca git tag but that's going to create conflicts as we're now gonna have UI releases and Parca (binary) releases.

brancz:
If I understand correctly, the version will never be checked into the repo after this initial setup?

yomete:
Oh it will everytime we make updates to any of the components. we don't want that? 

(unless I'm misinterpreting your question)

So as an example, to test publishing, I updated the @parcaui/profile component by deleting some LOCs [here](https://github.com/parca-dev/parca/pull/558/commits/5f1dbd8e1ff628d52e0023106ca582ee691a4f2c) and then ran `lerna publish` locally, just to see how it would handle updating the pkg.

That committed and created this [tag](https://github.com/parca-dev/parca/pull/558/commits/9097466ddf80ac9c6e96510ad2918488efcb955b). All the updates (version updates, changelog) was handled by Lerna and of course it also published to the npm registry.

brancz:
If at all possible I would prefer if we didn't need to do anything but create a git tag/release and GitHub actions do all the rest automatically. In my experience, having to do _anything_ manually ends up costing a lot of time over the years and it's worth it to put in the time to automate the release process in every way possible, as that way we can 100% focus on producing value instead of spending time on releasing.

yomete:
Definitely agreed. I just added (apologies for the several commits 😓 ) a new workflow job to the `ui.yml` which will automate the process. 

I did a couple of test runs on my forked repo, as you can see here, after making [this](https://github.com/yomete/parca/commit/8b899416a4b80d4ffd6724a8988b21c88dab0aca) change to the @parcaui/dynamicsize component, GH Actions [picked](https://github.com/yomete/parca/runs/4871588000?check_suite_focus=true) it up immediately and published all the pkgs that needed to be published.

The job (Publish UI components to NPM) is currently failing because it needs a `NPMTOKEN` secret, which I've added to Bitwarden :)

manojVivek (approved):
This is looking great @yomete 🎉! 

TIL: It is seamless to make a commit and push it in Github Actions. It used to be not so straightforward in Jenkins and Teamcity.

brancz:
Love it! lgtm!! :tada

PR #559, opened 2022-01-17
[WIP] Add support for multi-arch container builds

mrueg:
At this time, this adds additional support for arm64-based containers.
This can be useful for cloud-based platforms like graviton or to run
parca at home on your SBC cluster.

TODO: Still needs verification that the CI path works and actual testing on an arm64 board.

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -18,8 +18,17 @@ jobs:
         shell: bash
         run: echo "GITHUB_BRANCH_NAME=$(echo ${GITHUB_REF#refs/heads/} | tr / -)" >> $GITHUB_ENV
 
+      # To build multi-arch images, see also: https://github.com/redhat-actions/buildah-build#multi-arch-builds
+      - name: Install qemu dependency
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y qemu-user-static
+
       - name: Build container
         run: make container
+
+      - name: Check images created
+        run: buildah images | grep 'ghcr.io/parca-dev/parca'
 
       - name: Login to registry
         if: ${{ github.event_name != 'pull_request' }}
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,9 +27,10 @@ RUN mkdir /.cache && chown nobody:nogroup /.cache && touch -t 202101010000.00 /.
 
 ARG VERSION
 ARG COMMIT
+ARG ARCH=amd64
 ENV CGO_ENABLED=0
 ENV GOOS=linux
-ENV GOARCH=amd64
+ENV GOARCH=$ARCH
 
 WORKDIR /app
 
@@ -46,6 +47,8 @@ COPY --chown=nobody:nogroup ./ui/ui.go ./ui/ui.go
 COPY --chown=nobody:nogroup --from=ui-builder /app/packages/app/web/dist ./ui/packages/app/web/dist
 RUN go build -ldflags "-X main.version=${VERSION} -X main.commit=${COMMIT}" -trimpath -o parca ./cmd/parca
 RUN go install github.com/grpc-ecosystem/grpc-health-probe@latest
+# Predicatable path for copying over to final image
+RUN if [ "$(go env GOHOSTARCH)" != "$(go env GOARCH)" ]; then mv "$(go env GOPATH)/bin/$(go env GOOS)_$(go env GOARCH)/grpc-health-probe" "$(go env GOPATH)/bin/grpc-health-probe"; fi
 
 # this image is what docker.io/alpine:3.14.1 on August 13 2021
 FROM docker.io/alpine@sha256:be9bdc0ef8e96dbc428dc189b31e2e3b05523d96d12ed627c37aa2936653258c
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@ else
 	COMMIT := $(shell echo $(GITHUB_SHA) | cut -c1-8)
 endif
 VERSION ?= $(if $(RELEASE_TAG),$(RELEASE_TAG),$(shell $(CMD_GIT) describe --tags 2>/dev/null || echo '$(BRANCH)$(COMMIT)'))
+ALL_ARCH ?= amd64 arm arm64
 OUT_DOCKER ?= ghcr.io/parca-dev/parca
 
 .PHONY: build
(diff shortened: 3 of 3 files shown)


kakkoyun (commented):
This looks good! Thanks for your contribution :)

CI failures seem unrelated to this PR. We will check them out.

As you've suggested we need to figure out testing this with our CI. Dropping some links for future reference:
* https://github.blog/changelog/2019-12-03-github-actions-self-hosted-runners-on-arm-architectures/
* https://github.com/uraimo/run-on-arch-action

kakkoyun:
@mrueg Could you rebase? CI should be fixed on the main branch now.

.github/workflows/container.yml
```diff
@@ -18,6 +18,11 @@ jobs:
         shell: bash
         run: echo "GITHUB_BRANCH_NAME=$(echo ${GITHUB_REF#refs/heads/} | tr / -)" >> $GITHUB_ENV
 
+      - name: Install qemu dependency
```
brancz, line 21: Could we add a comment on why is this necessary?

Dockerfile
```diff
@@ -46,6 +47,8 @@ COPY --chown=nobody:nogroup ./ui/ui.go ./ui/ui.go
 COPY --chown=nobody:nogroup --from=ui-builder /app/packages/app/web/dist ./ui/packages/app/web/dist
 RUN go build -ldflags "-X main.version=${VERSION} -X main.commit=${COMMIT}" -trimpath -o parca ./cmd/parca
 RUN go install github.com/grpc-ecosystem/grpc-health-probe@latest
+# Predicatable path for copying over to final image
+RUN if [ "$(go env GOHOSTARCH)" != "$(go env GOARCH)" ]; then mv "$(go env GOPATH)/bin/$(go env GOOS)_$(go env GOARCH)/grpc-health-probe" "$(go env GOPATH)/bin/grpc-health-probe"; fi
 
 # this image is what docker.io/alpine:3.14.1 on August 13 2021
 FROM docker.io/alpine@sha256:be9bdc0ef8e96dbc428dc189b31e2e3b05523d96d12ed627c37aa2936653258c
```
brancz, line 54: I kind of expected that we need to swap out the base image for the appropriate one with the target architecture? If not then I think I would need a comment for future Frederic to understand why :joy:

Dockerfile
```diff
@@ -46,6 +47,8 @@ COPY --chown=nobody:nogroup ./ui/ui.go ./ui/ui.go
 COPY --chown=nobody:nogroup --from=ui-builder /app/packages/app/web/dist ./ui/packages/app/web/dist
 RUN go build -ldflags "-X main.version=${VERSION} -X main.commit=${COMMIT}" -trimpath -o parca ./cmd/parca
 RUN go install github.com/grpc-ecosystem/grpc-health-probe@latest
+# Predicatable path for copying over to final image
+RUN if [ "$(go env GOHOSTARCH)" != "$(go env GOARCH)" ]; then mv "$(go env GOPATH)/bin/$(go env GOOS)_$(go env GOARCH)/grpc-health-probe" "$(go env GOPATH)/bin/grpc-health-probe"; fi
```
brancz, line 51: Nice!

brancz (commented):
cc @Sylfrena since you were starting to look into this as well

Dockerfile
mrueg: It's a multi-arch container image already, see also: `buildah manifest inspect docker.io/alpine:3.15` Based on the hostarch it will pick the right top layer

.github/workflows/container.yml
mrueg: Done! :)

Makefile
```diff
@@ -76,7 +77,9 @@ proto/vendor:
 
 .PHONY: container
 container:
-	buildah build-using-dockerfile --timestamp 0 --layers --build-arg VERSION=$(VERSION) --build-arg COMMIT=$(COMMIT) -t $(OUT_DOCKER):$(VERSION) .
+	for arch in $(ALL_ARCH); do \
+	buildah build-using-dockerfile --build-arg VERSION=$(VERSION) --build-arg COMMIT=$(COMMIT) --build-arg ARCH=$$arch --arch $$arch --timestamp 0 --manifest $(OUT_DOCKER):$(VERSION); \
```
mrueg, line 81: One caveat here is that this removes caching of layers. If devs use "make container" as part of their workflow, they might want to readjust (or add a different single-arch target).

Dockerfile
brancz: Woah, I didn't even know this worked with the exact SHA. Today I learned ...

Makefile
brancz: Good call, can we maybe have a "container-dev" make target for the dev use case?

Dockerfile
mrueg: You might be right. Inspecting the sha gives a different output, I guess I'll should be able to find it out on an arm64 board over the weekend.

mhausenblas:
I'm very interested in this. Use case: running Parca on my Raspberry Pi cluster. I believe that's something along the line of
`GOARCH=arm GOARM=7` and ` GOARCH=arm64` from what I can glean from https://zchee.github.io/golang-wiki/GoArm/

mrueg:
> I'm very interested in this. Use case: running Parca on my Raspberry Pi cluster. I believe that's something along the line of `GOARCH=arm GOARM=7` and ` GOARCH=arm64` from what I can glean from https://zchee.github.io/golang-wiki/GoArm/

I've added additional support for 32-bit arm nodes as well (although we can probably discuss if this is worth building).

mhausenblas:
Well, super cool @mrueg and can't wait to see this PR merged and the images available :)

brancz:
@mrueg we publish container images from the main branch on every merge, do you want to just give this a try and see if it works and do follow-up PRs if there is anything to fix based on @mhausenblas's feedback?

mrueg:
Sure, we can do that as well!
Once things work I'll create a PR to replicate it on parca-agent.

Dockerfile
kakkoyun: A note to the future, we should get rid of `grpc-health-probe` in the long run. With the introduction of grpc-based health probes in k8s v1.23, this shouldn't be a huge problem anymore https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-grpc-liveness-probe

Of course, there will be a tread-off between simplicity and broader platform support. Up for discussion.

brancz:
Let's see how it goes

mrueg:
```buildah manifest inspect ghcr.io/parca-dev/parca:main-633b0938 
tried manifest is of type application/vnd.oci.image.manifest.v1+json (not a list type)
: manifest is of type application/vnd.oci.image.manifest.v1+json (not a list type)``` 
looks like push needs some more changes.

PR #573, opened 2022-01-21
Makefile: Push manifest instead of image

mrueg:
@brancz  This should allow to push all architectures.

currently it only pushed one image.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,11 +87,11 @@ container:
 
 .PHONY: push-container
 push-container:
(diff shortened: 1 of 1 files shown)


brancz (approved):
Thank you so much for the follow up!

metalmatze:
I assume this was just waiting for CI and thus not merged. I'm going ahead now. :) 
Thanks, @mrueg!

mrueg:
whoops, that required a second arg. following up again.

PR #580, opened 2022-01-26
pkg/columnstore: Make iterators return arrow frames

brancz:
Not really sure how to tell arrow to use a certain encoding :thinking: but this is a start


brancz:
Merged directly into the `col2` branch for a start.

Issue #499, opened 2021-12-16
Flaky test `Test_QueryRange_Limited`


brancz:
I've noticed the `Test_QueryRange_Limited` test flakes every now and then on PRs. This should be fixed.

metalmatze:
I was trying to find a flaky test log output on GitHub Actions. Is there still some available? I wasn't able to find anything under Go in our Github Actions.
Could you link to one? Thanks.

brancz:
https://github.com/parca-dev/parca/runs/4507396630?check_suite_focus=true

brancz:
https://github.com/parca-dev/parca/runs/4591987969?check_suite_focus=true

metalmatze:
Thanks for the links. Something is up with this test. While working on the deletion of profile trees it got my attention because it took 30s to complete. I'll definitely take a closer look on what's going on.

metalmatze:
Still flaky...

importhuman:
Think this can be closed as #771 is closed

metalmatze:
Nice. Yes, if it ever shows up again let's open a new one.

PR #582, opened 2022-01-27
pkg/query: Fix adding a unit to the flamegraph

### LINKED ISSUE #578: Broken cumulative value formatting in profile popup
It seems that for some reason our formatting of cumulative values in the popup is broken.

Screenshot from 2022-01-25 18-54-09

It's probably some wiring in this package: https://github.com/parca-dev/parca/tree/main/ui/packages/shared/profile

metalmatze:
Fixes #578

For some reason, we didn't set the unit field anymore. :man_shrugging: 
It might be easier to construct the struct in the workflow I chose now and similar mistakes could be easier to catch. Either way, I added updated a unit test too to catch it in the future.

### DIFF
--- a/pkg/query/flamegraph_flat.go
+++ b/pkg/query/flamegraph_flat.go
@@ -20,8 +20,8 @@ import (
 	"sort"
 
 	pb "github.com/parca-dev/parca/gen/proto/go/parca/query/v1alpha1"
-	"github.com/parca-dev/parca/pkg/profile"
 	"github.com/parca-dev/parca/pkg/metastore"
+	"github.com/parca-dev/parca/pkg/profile"
 	"go.opentelemetry.io/otel/trace"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/query/flamegraph_flat.go
```diff
@@ -116,12 +116,16 @@ func GenerateFlamegraphFlat(ctx context.Context, tracer trace.Tracer, metaStore
...
-	flamegraph.Root.Diff = rootNode.Diff
-	flamegraph.Root.Children = rootNode.Children
+	flamegraph := &pb.Flamegraph{
+		Root: &pb.FlamegraphRootNode{
+			Cumulative: rootNode.Cumulative,
+			Diff:       rootNode.Diff,
+			Children:   rootNode.Children,
+		},
+		Total:  rootNode.Cumulative,
+		Unit:   p.ProfileMeta().SampleType.Unit,
```
metalmatze, line 126: This line was missing before.

metalmatze:
The CI errors are related to https://github.com/parca-dev/parca/issues/499 
Merging here and then finally trying to fix that flaky test.

PR #586, opened 2022-01-28
Close profile card in compare view

### LINKED ISSUE #211: Close profile diffing panel
It'd be great to have a close button for diffing. 
To demonstrate what I mean, I've put an ugly red X into this screenshot. The actual icon should be a lot nicer.
Probably, it will look like closing the second panel, but in the background, we should redirect back to the single query panel with the query from the left.

diff-close

avinashupadhya99:
Closes #211 

https://user-images.githubusercontent.com/52544819/151407091-d6f30235-060a-4ad7-8a91-e279b75a43ab.mp4

The `compare_a` parameter continues to exist in the single profile view. If necessary, I can write another function that will filter the params based on the prefix(`compare_` in this case).


thorfour:
YES! I love this change!

brancz:
Looooooove it! lgtm! Thank you so much for contributing this!

metalmatze:
Thank you so much @avinashupadhya99 :tada:

PR #588, opened 2022-01-29
Use arrow types

thorfour:
After reading through all the arrow converters we had I was a little confused why we'd want to convert into our own "arrow" type instead of just converting directly into a proper arrow record since I thought the goal was to have an open type as our interface between parca and the data store. 

So I took a stab at a direct conversion instead of intermediary.

@brancz please TAL


brancz:
You're totally right! Feel free to merge this into the col2 branch and ignore linting for now. I mistakenly thought the record type was just a convenience type, but I completely failed to look at the rest of the arrow package until now where I would have seen it being used all over the place :see_no_evil: 

Thank you very much for questioning the approach, that saved us a whole lot of trouble down the line! :100:

PR #598, opened 2022-02-02
Add targets page

monicawoj:
Resolves 215

This PR relates to creating a "Targets" page in the Parca UI. As part of this PR, a reusable `Pill` component to highlight information as a tag colored based on the specified variant, was also created. 

An additional task that might make sense to work on at a later time related to the content of this PR would be to create a reusable `Table` component with additional features that would make for a better user experience than just scrolling (such as pagination, filtering by target "health"/status, etc). The UI may likely undergo some tweaking as well - very open to suggestions!


metalmatze:
Skimming over the file changes, I'm very excited about this! :tada:

monicawoj:
> Skimming over the file changes, I'm very excited about this! 🎉

I'll be doing a self-review and making some adjustments before it's ready for a "real" review, but glad to hear that! :D

ui/packages/app/web/src/components/Targets/LabelsCell.tsx
```diff
@@ -0,0 +1,53 @@
+import React, {useState} from 'react';
+import {ChevronDoubleDownIcon, ChevronDoubleUpIcon} from '@heroicons/react/solid';
+import Pill from '../ui/Pill';
+
+export interface ITargetEndpoint {}
```
monicawoj, line 5: what is the exact structure of the TargetEndpoint? fill in as appropriate

ui/packages/app/web/src/components/Targets/LabelsCell.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+const LabelsCell = ({key, labels, discoveredLabels}) => {
+  const [areDiscoveredLabelsVisible, setAreDiscoveredLabelsVisible] = useState<boolean>(false);
+
+  return (
+    <td key={key} className="px-6 py-4 whitespace-nowrap text-sm text-gray-500 flex flex-col w-96">
+      <div className="flex flex-wrap">
+        {labels.length &&
+          labels.map(item => {
```
monicawoj, line 14: add a type 'item'

ui/packages/app/web/src/components/Targets/LabelsCell.tsx
```diff
@@ -0,0 +1,53 @@
...
+            return (
+              <div className="pb-1 pr-1">
+                <Pill key={item['name']} variant="info">{`${item.name}="${item.value}"`}</Pill>
```
monicawoj, line 17: if possible, use `item.name` instead of item['name']

monicawoj, line 17: save Pill variant as a separate enum instead of using a string as the value (i.e. PillVariant.Info instead of "info" )

ui/packages/app/web/src/components/Targets/LabelsCell.tsx
```diff
@@ -0,0 +1,53 @@
...
+              </div>
+            );
+          })}
+        {areDiscoveredLabelsVisible &&
+          discoveredLabels.length &&
+          discoveredLabels.map(item => {
```
monicawoj, line 23: type 'item'

ui/packages/app/web/src/components/Targets/LabelsCell.tsx
```diff
@@ -0,0 +1,53 @@
...
+            return (
+              <div className="pb-1 pr-1">
+                <Pill key={item['name']} variant="info">{`${item.name}="${item.value}"`}</Pill>
```
monicawoj, line 26: item.name

ui/packages/app/web/src/components/Targets/LabelsCell.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export interface ITargetEndpoint {}
+
+const LabelsCell = ({key, labels, discoveredLabels}) => {
+  const [areDiscoveredLabelsVisible, setAreDiscoveredLabelsVisible] = useState<boolean>(false);
+
+  return (
+    <td key={key} className="px-6 py-4 whitespace-nowrap text-sm text-gray-500 flex flex-col w-96">
+      <div className="flex flex-wrap">
+        {labels.length &&
```
monicawoj, line 13: instead of rendering two separate lists (for labels and discoveredLabels) just render one list and conditionally control the items in that list (sometimes just 'labels', sometimes 'labels' and 'discoveredLabels')

ui/packages/app/web/src/components/Targets/LabelsCell.tsx
```diff
@@ -0,0 +1,53 @@
...
+        }
+      >
+        <span className="mr-1">{`${
+          areDiscoveredLabelsVisible ? 'Hide' : 'Show'
+        } Discovered Labels`}</span>
+        {areDiscoveredLabelsVisible ? (
+          <ChevronDoubleUpIcon className="h-5 w-5" aria-hidden="true" />
+        ) : (
+          <ChevronDoubleDownIcon className="h-5 w-5" aria-hidden="true" />
+        )}
```
monicawoj, line 47: may be more readable to have the if-else be on the element level so as to avoid double use of `areDiscoveredLabelsVisible`

ui/packages/app/web/src/components/Targets/TargetsTable.tsx
```diff
@@ -0,0 +1,104 @@
...
+              })}
+            </tr>
+          );
+        })}
+      </tbody>
+    </table>
+  );
+};
+
+export default TargetsTable;
```
monicawoj, line 104: would be good to extract the "Table" to a reusable component in the 'ui' directory, but that might make more sense once we decide a table solution/further specifications (for example, pagination? filtering?)

ui/packages/app/web/src/components/ui/Navbar.tsx
```diff
@@ -16,6 +17,11 @@ const Navbar = ({
   isDarkMode: boolean;
   setDarkMode: (mode: boolean) => void;
 }) => {
+  const getPageByHref = (href: string): {name: string; href: string; external: boolean} =>
+    links.find(link => link.href === href) || links[0];
```
monicawoj, line 21: instead of `links[0]` we should define a default 'home' page

ui/packages/app/web/src/components/Targets/utils.ts
```diff
@@ -0,0 +1,105 @@
+import moment from 'moment';
```
brancz, line 1: I think we started converting things to date-fns, is there a specific functionality in moment that date-fns doesn't support?

brancz (commented):
Very nice!! :100: 

I have one question about the usage of moments, and I think there are still a couple of linters failing. Not sure why those are only failing in the go builds and not in the JS GitHub actions.

ui/packages/app/web/src/components/Targets/utils.ts
```diff
@@ -0,0 +1,105 @@
...
+  const label = HealthStatus[numericValue];
+  const colorVariants = {
+    Unspecified: 'neutral',
+    Good: 'success',
+    Bad: 'danger',
+  };
+  return {label, colorVariant: colorVariants[label]};
+};
+
+export const secondsInNanos = (seconds: number) => seconds * 1e9;
```
yomete, line 19: I wonder if we can reuse this utility function [here](https://github.com/parca-dev/parca/blob/main/ui/packages/shared/functions/src/index.ts#L41) to format the values? 🤔

yomete (changes requested):
nice work on this!
 
Just a few things to change. I noticed even in dark mode, the table has a white background.
image

Also, you'd probably have to make some additional changes so as to make this page visible in the Go binary. I already have a PR [here](https://github.com/parca-dev/parca/pull/508) that shows how to do it. happy to discuss further!

ui/packages/app/web/src/components/Targets/utils.ts
monicawoj: @brancz I'm all for using `date-fns`! just wanted to confirm this migration from moment is the case before making changes though.  Just asking because I still see several instances of 'moment' being used throughout the code.

brancz: Whops, then that was definitely an oversight, I thought we had done the migration. Then let's use date-fns here and maybe open an issue for converting the other places where we use moment?

ui/packages/app/web/src/components/Targets/utils.ts
monicawoj: The issue is that the formatting function `valueFormatter` currently only formats to a single unit (3.5 minutes vs. 3m 30s). I believe formatting a value to multiple units might be more understandable for larger values or durations. For example, it might be more understandable to see '6d 4h ago' instead of '6.17d ago'. What do you think @yomete?

ui/packages/app/web/src/components/Targets/TargetsTable.tsx
metalmatze: Yep, that approach sounds reasonable! :+1:

ui/packages/app/web/src/components/Targets/utils.ts
monicawoj: turns out that I was able to avoid using an external date library (Moment or date-fns) altogether 😄

ui/packages/app/web/src/components/Targets/utils.ts
monicawoj: I modified the value formatter function in our @parca/functions file and created a separate function as well for this use case. In particular, I think a separate function is necessary if we want to communicate durations in a more human-readable manner - that is, instead of a single unit, communicating several (3m 30s instead of 3.5m)

monicawoj:
> make this page visible in the Go binary

checks are still failing because I still have to make the page visible in the Go binary as Yomi mentioned!

ui/packages/app/web/src/components/Targets/utils.ts
yomete: just seen it. love it!

metalmatze:
Turns out we still needed to register the grpc-web API endpoints with a sub router that then gets mounted onto `/api`:
https://github.com/parca-dev/parca/pull/598/commits/128f86168b95a576a88cc46fa46c87f975e158ed#diff-104a11712684e520f0affa95e6a053faeb713306a10f69ee99b9729cc9dd1996R142

metalmatze (approved):

PR #600, opened 2022-02-03
Only run GitHub Actions when related files changed

metalmatze:
Especially for golangci-lint, proto and jsonnet we don't need to run these actions, if none of the related files changed.
As for the UI and Go itself I'm not 100% sure but willing to try. Maybe doing a change to the README would be super fast then? Not sure if those are worth it in the end. WDYT?


kakkoyun (approved):
LGTM. I loved it.

PR #601, opened 2022-02-04
pkg/columnstore: Add filtering via expressions

brancz:
@thorfour this took way longer than I thought and I implemented way less than I thought I could in the first take.

Aggregations are next. It's possible that we will still need to add projections as well.


brancz:
merged into branch

PR #610, opened 2022-02-09
Move the Parca UI to use Create React App as a toolchain 

### LINKED ISSUE #434: Allow configurable base path in UI
In our setup, we have a shared nginx which we add services to. I've added `/parca` to it, but the problem is the UI serves all assets from the root `/`.

I know nothing about next.js, but this looks like it might work? https://nextjs.org/docs/api-reference/next.config.js/basepath
How we would pass that as an argument from the main Go binary to next.js I have no clue about :slightly_smiling_face:

yomete:
This is an ongoing PR that moves the frontend of the Parca project from Next.js to Create React App. This is being done so as to support configurable base paths and allow anyone run the Parca binary from any subfolder in their workspace.

To do:
- [x] Add CRA to ui/packages/app/web by installing react-scripts.  
- [x] Restructure the ui/packages/app/web directory
    - [x] Create a ui/packages/app/web/src/index.tsx file to be the JavaScript entry point.
    - [x] Add an index.html file to the public folder to serve as the page template.
- [x] Replace the dev/build/test scripts in package.json to use <del>react-scripts</del> craco
- [x] Look out for NextJS dependencies/APIs we might be using and use it’s CRA/community alternative
- [x] Ensure that this works with the Go binary
    - [x] NextJS uses dist as the build folder (for production builds) while CRA uses build so we should look out for that too
    - [x] we’ll also need to modify the Go server that currently houses the Parca UI to respond to client side requests by always serving index.html (as for a CRA production build, a static file server is [not](https://create-react-app.dev/docs/deployment#serving-apps-with-client-side-routing) enough for client side routing (which we do in Parca).)

Resolves #434


pkg/parca/parca.go
```diff
@@ -63,6 +63,7 @@ type Flags struct {
 	CORSAllowedOrigins []string `help:"Allowed CORS origins."`
 	OTLPAddress        string   `help:"OpenTelemetry collector address to send traces to."`
 	Version            bool     `help:"Show application version."`
+	PathPrefix         string   `default:"" help:"Path prefix for the UI"`
```
metalmatze, line 66: :tada: :tada: :tada:

ui/package.json
```diff
@@ -6,11 +6,7 @@
     "lint": "eslint --ext .ts,.tsx,.js packages/*",
     "fix": "eslint --fix --ext .ts,.tsx,.js packages/*",
     "test": "jest --coverage --config jest.config.js",
-    "dev": "next dev",
-    "build": "next build",
```
metalmatze, line 10: Do we really want to remove these entirely? :thinking:

ui/packages/app/web/src/components/ui/Navbar.tsx
```diff
@@ -77,10 +78,10 @@ const Navbar = ({
                 </div>
               </div>
               <div className="text-gray-500 dark:text-gray-0 absolute inset-y-0 right-0 flex items-center pr-2 sm:static sm:inset-auto sm:ml-6 sm:pr-0">
-                {'{{.Version}}'}
+                {window.APP_VERSION}
```
metalmatze, line 81: Oh yes. So much better :sunglasses:

ui/ui.go
```diff
@@ -15,11 +15,8 @@ package ui
 
 import "embed"
 
-//go:embed packages/app/web/dist
-//go:embed packages/app/web/dist/targets
-//go:embed packages/app/web/dist/_next
-//go:embed packages/app/web/dist/_next/static/chunks/pages/*.js
-//go:embed packages/app/web/dist/_next/static/*/*.js
+//go:embed packages/app/web/build
```
metalmatze, line 18: Even more yay! :tada:

metalmatze (commented):
Amazing work! :tada: :rocket:

ui/package.json
manojVivek: Yeah, these are just on the root `package.json`. Don't really serve any purpose there.

brancz (approved):
Amazing!! lgtm 🎉

PR #611, opened 2022-02-09
pkg/debuginfo: Add debuginfod client support

Sylfrena:
Signed-off-by: Sumera Priyadarsini <sylphrenadin@gmail.com>

### DIFF
--- a/pkg/debuginfo/debuginfod.go
+++ b/pkg/debuginfo/debuginfod.go
@@ -1,0 +1,44 @@
+package debuginfo
+
+import (
+	"io"
+	"net/http"
+	"path"
+)
+
+var publicServer = "https://debuginfod.systemtap.org"
+
+type DebugInfodClient interface {
+	GetDebugInfo(buildid string) (io.ReadCloser, error)
+}
+
+type HttpDebuginfodClient struct {
+	UpstreamServer string //url
+}
+type ObjectStorageDebugInfodClientCache struct {
+	client DebugInfodClient
+}
+
+func NewHttpDebugInfoClient(serverUrl string) HttpDebuginfodClient {
+	return HttpDebuginfodClient{UpstreamServer: serverUrl}
+}
+
+func NewObjectStorageDebugInfodClientCache(h HttpDebuginfodClient) *ObjectStorageDebugInfodClientCache {
+	return &ObjectStorageDebugInfodClientCache{client: &h}
+}
+
+func (c *ObjectStorageDebugInfodClientCache) GetDebugInfo(buildid string) (io.ReadCloser, error) {
+	return c.client.GetDebugInfo(buildid)
+}
+
+func (c *HttpDebuginfodClient) GetDebugInfo(buildID string) (io.ReadCloser, error) {
+	serverUrl := path.Join(c.UpstreamServer, buildID, "debuginfo")
+
+	resp, err := http.Get(serverUrl)
+	if err != nil {
+		//level.Debug(logger).Log("msg", "object not found in public server", "object", buildID, "err", err)
+		return nil, ErrDebugInfoNotFound
+	}
+
+	return resp.Body, nil
+}
--- a/pkg/debuginfo/debuginfod.go
+++ b/pkg/debuginfo/debuginfod.go
@@ -14,70 +14,108 @@
 package debuginfo
 
 import (
+	"fmt"
 	"io"
 	"net/http"
 	"net/url"
 	"path"
+	"time"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/thanos-io/thanos/pkg/objstore"
+	"github.com/thanos-io/thanos/pkg/objstore/client"
 	"golang.org/x/net/context"
+	"gopkg.in/yaml.v2"
 )
 
 type DebugInfodClient interface {
-	GetDebugInfo(buildid string) (io.ReadCloser, error)
+	GetDebugInfo(ctx context.Context, buildid string) (io.ReadCloser, error)
 }
 
-type HttpDebuginfodClient struct {
+type HttpDebugInfodClient struct {
+	logger         log.Logger
 	UpstreamServer *url.URL
 }
 type ObjectStorageDebugInfodClientCache struct {
-	ctx    context.Context
 	logger log.Logger
 	client DebugInfodClient
 	bucket objstore.Bucket
 }
 
-func NewHttpDebugInfoClient(serverUrl string) (*HttpDebuginfodClient, error) {
+func NewHttpDebugInfoClient(logger log.Logger, serverUrl string) (*HttpDebugInfodClient, error) {
 	parsedUrl, err := url.Parse(serverUrl)
 	if err != nil {
 		return nil, err
 	}
-	return &HttpDebuginfodClient{UpstreamServer: parsedUrl}, nil
+	return &HttpDebugInfodClient{
+		logger:         logger,
+		UpstreamServer: parsedUrl,
+	}, nil
 }
 
-func NewObjectStorageDebugInfodClientCache(ctx context.Context, logger log.Logger, h *HttpDebuginfodClient) *ObjectStorageDebugInfodClientCache {
+func NewObjectStorageDebugInfodClientCache(logger log.Logger, config *Config, h DebugInfodClient) (*ObjectStorageDebugInfodClientCache, error) {
+	cfg, err := yaml.Marshal(config.Bucket)
+	if err != nil {
+		return nil, fmt.Errorf("marshal content of debuginfod object storage configuration: %w", err)
+	}
+
+	bucket, err := client.NewBucket(logger, cfg, nil, "parca")
+	if err != nil {
+		return nil, fmt.Errorf("instantiate debuginfod object storage: %w", err)
+	}
 	return &ObjectStorageDebugInfodClientCache{
-		ctx:    ctx,
 		logger: logger,
 		client: h,
-	}
+		bucket: bucket,
+	}, nil
 }
 
-func (c *ObjectStorageDebugInfodClientCache) GetDebugInfo(buildId string) (io.ReadCloser, error) {
-	debugInfo, err := c.client.GetDebugInfo(buildId)
+func (c *ObjectStorageDebugInfodClientCache) GetDebugInfo(ctx context.Context, buildId string) (io.ReadCloser, error) {
+	path := buildId + "/debuginfod-cache/debuginfo"
+
+	if exists, _ := c.bucket.Exists(ctx, path); exists {
+		debuginfoFile, err := c.bucket.Get(ctx, path)
+		if err != nil {
+			level.Debug(c.logger).Log("msg", "object file present in debuginfod cache", "build_id", buildId)
+			level.Error(c.logger).Log("msg", "failed to download from debuginfod cache", "err", err)
+			return nil, err
+		}
+		return debuginfoFile, nil
+	}
+
+	debugInfo, err := c.client.GetDebugInfo(ctx, buildId)
 	if err != nil {
 		return nil, ErrDebugInfoNotFound
 	}
 
-	path := buildId + "/debuginfo"
-
-	err = c.bucket.Upload(c.ctx, path, debugInfo)
+	err = c.bucket.Upload(ctx, path, debugInfo)
 	if err != nil {
 		level.Error(c.logger).Log("msg", "failed to upload to debuginfod cache", "err", err)
 		return nil, err
 	}
-	return debugInfo, nil
+	debugInfoReader, err := c.bucket.Get(ctx, path)
+	if err != nil {
+		level.Error(c.logger).Log("msg", "failed to download from debuginfod cache", "err", err)
+		return nil, err
+	}
+
+	return debugInfoReader, nil
 }
 
-func (c *HttpDebuginfodClient) GetDebugInfo(buildID string) (io.ReadCloser, error) {
+func (c *HttpDebugInfodClient) GetDebugInfo(ctx context.Context, buildID string) (io.ReadCloser, error) {
 	buildIdUrl := *c.UpstreamServer
 	buildIdUrl.Path = path.Join(buildIdUrl.Path, buildID, "debuginfo")
 
-	resp, err := http.Get(buildIdUrl.String())
+	ctx, cancel := context.WithTimeout(ctx, 1*time.Millisecond)
+	defer cancel()
+
+	req, _ := http.NewRequestWithContext(ctx, "GET", buildIdUrl.String(), nil)
+
+	client := http.DefaultClient
+	resp, err := client.Do(req)
 	if err != nil {
-		//level.Debug(logger).Log("msg", "object not found in public server", "object", buildID, "err", err)
+		level.Debug(c.logger).Log("msg", "object not found in public server", "object", buildID, "err", err)
 		return nil, ErrDebugInfoNotFound
 	}
 
--- a/pkg/debuginfo/debuginfod.go
+++ b/pkg/debuginfo/debuginfod.go
@@ -33,25 +33,26 @@ type DebugInfodClient interface {
 	GetDebugInfo(ctx context.Context, buildid string) (io.ReadCloser, error)
 }
 
-type HttpDebugInfodClient struct {
+type HTTPDebugInfodClient struct {
 	logger          log.Logger
 	UpstreamServer  *url.URL
 	timeoutDuration time.Duration
 }
+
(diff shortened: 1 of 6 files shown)


pkg/debuginfo/store.go
```diff
@@ -71,12 +73,14 @@ type Store struct {
...
 
 // NewStore returns a new debug info store
-func NewStore(logger log.Logger, symbolizer *symbol.Symbolizer, config *Config) (*Store, error) {
+func NewStore(logger log.Logger, symbolizer *symbol.Symbolizer, config *Config, debuginfodUrl string) (*Store, error) {
 	cfg, err := yaml.Marshal(config.Bucket)
 	if err != nil {
 		return nil, fmt.Errorf("marshal content of object storage configuration: %w", err)
 	}
 
+	debuginfodClientCache := NewObjectStorageDebugInfodClientCache(NewHttpDebugInfoClient(debuginfodUrl))
```
brancz, line 82: I think this client should be instantiated by the caller of this function instead. The store should not need to care that what it uses is a cached client, it should just accept a parameter that implements the `DebugInfodClient` interface.

pkg/debuginfo/store.go
```diff
@@ -228,9 +233,12 @@ func (s *Store) fetchObjectFile(ctx context.Context, buildID string) (string, er
...
 	if _, err := os.Stat(mappingPath); os.IsNotExist(err) {
 		r, err := s.bucket.Get(ctx, path.Join(buildID, "debuginfo"))
+
 		if s.bucket.IsObjNotFoundErr(err) {
-			level.Debug(s.logger).Log("msg", "object not found", "object", buildID, "err", err)
-			return "", ErrDebugInfoNotFound
+			level.Debug(s.logger).Log("msg", "object not found in parca object storage", "object", buildID, "err", err)
+			//return "", ErrDebugInfoNotFound
+			//if publicDebuginfodBuildIDExists()
+			r, err = s.debuginfodClientCache.GetDebugInfo(buildID)
```
brancz, line 241: I think this deserves its own `if err != nil` as the error wrapping below does not represent the same situation as this potential error.

pkg/parca/parca.go
```diff
@@ -71,6 +71,10 @@ type Flags struct {
 	SymbolizerNumberOfTries int    `default:"3" help:"Number of tries to attempt to symbolize an unsybolized location"`
 
 	Metastore string `default:"badgerinmemory" help:"Which metastore implementation to use" enum:"sqliteinmemory,badgerinmemory"`
+
+	UpstreamDebuginfodServer string `default:"https://debuginfod.systemtap.org" help:"Upstream private/public server for debuginfod files. Defaults to https://debuginfod.systemtap.org
+	."`
+	//TODO: change this to a url type
```
brancz, line 77: I think here it can be a string and parse it within the client instantiation.

pkg/debuginfo/debuginfod.go
```diff
@@ -0,0 +1,44 @@
+package debuginfo
+
+import (
+	"io"
+	"net/http"
+	"path"
+)
+
+var publicServer = "https://debuginfod.systemtap.org"
```
brancz, line 9: Since the flag defines the default, this seems unused. I think we can safely remove it.

pkg/debuginfo/debuginfod.go
```diff
@@ -0,0 +1,44 @@
...
+func NewHttpDebugInfoClient(serverUrl string) HttpDebuginfodClient {
+	return HttpDebuginfodClient{UpstreamServer: serverUrl}
+}
+
+func NewObjectStorageDebugInfodClientCache(h HttpDebuginfodClient) *ObjectStorageDebugInfodClientCache {
+	return &ObjectStorageDebugInfodClientCache{client: &h}
+}
+
+func (c *ObjectStorageDebugInfodClientCache) GetDebugInfo(buildid string) (io.ReadCloser, error) {
+	return c.client.GetDebugInfo(buildid)
```
brancz, line 31: I think this still needs the part that actually caches the returned debug info files, right?

pkg/debuginfo/debuginfod.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func (c *HttpDebuginfodClient) GetDebugInfo(buildID string) (io.ReadCloser, error) {
+	serverUrl := path.Join(c.UpstreamServer, buildID, "debuginfo")
```
brancz, line 35: This is why we want to parse the upstream server URL when we create the client, as `path.Join` should only be called on the actual path, instead of the whole URL.

brancz (commented):
Very nice start!

Sylfrena:
@brancz I am going to fix the tests, but apart from that, wdyt?

pkg/debuginfo/debuginfod.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/go-kit/log/level"
+	"github.com/thanos-io/thanos/pkg/objstore"
+	"golang.org/x/net/context"
+)
+
+type DebugInfodClient interface {
+	GetDebugInfo(buildid string) (io.ReadCloser, error)
+}
+
+type HttpDebuginfodClient struct {
```
kakkoyun, line 32:
```suggestion
type HttpDebugInfodClient struct {
```

pkg/debuginfo/store.go
```diff
@@ -71,7 +73,7 @@ type Store struct {
 }
 
 // NewStore returns a new debug info store
-func NewStore(logger log.Logger, symbolizer *symbol.Symbolizer, config *Config) (*Store, error) {
+func NewStore(logger log.Logger, symbolizer *symbol.Symbolizer, config *Config, debuginfodClientCache *ObjectStorageDebugInfodClientCache) (*Store, error) {
```
kakkoyun, line 76:
```suggestion
func NewStore(logger log.Logger, symbolizer *symbol.Symbolizer, config *Config, debuginfodClientCache DebugInfodClient) (*Store, error) {
```

I think it's better to use the interface here.

pkg/debuginfo/store.go
```diff
@@ -63,6 +63,8 @@ type CacheConfig struct {
 type Store struct {
 	debuginfopb.UnimplementedDebugInfoServiceServer
 
+	debuginfodClientCache *ObjectStorageDebugInfodClientCache
```
kakkoyun, line 66:
```suggestion
	debuginfodClientCache DebugInfodClient
```
I think it's better to use the interface here.

pkg/debuginfo/store.go
```diff
@@ -228,9 +231,15 @@ func (s *Store) fetchObjectFile(ctx context.Context, buildID string) (string, er
 	// Check if it's already cached locally; if not download.
 	if _, err := os.Stat(mappingPath); os.IsNotExist(err) {
 		r, err := s.bucket.Get(ctx, path.Join(buildID, "debuginfo"))
+
 		if s.bucket.IsObjNotFoundErr(err) {
-			level.Debug(s.logger).Log("msg", "object not found", "object", buildID, "err", err)
-			return "", ErrDebugInfoNotFound
+			level.Debug(s.logger).Log("msg", "object not found in parca object storage", "object", buildID, "err", err)
+
+			r, err = s.debuginfodClientCache.GetDebugInfo(buildID)
```
kakkoyun, line 238:
```suggestion
			r, err = s.debuginfodClientCache.GetDebugInfo(ctx, buildID)
```

As I pointed out above, context should be passed per request in here.

pkg/debuginfo/debuginfod.go
```diff
@@ -0,0 +1,85 @@
...
+	UpstreamServer *url.URL
+}
+type ObjectStorageDebugInfodClientCache struct {
+	ctx    context.Context
```
kakkoyun, line 36: I think context should be per request rather than per instance of the client.

pkg/debuginfo/debuginfod.go
```diff
@@ -0,0 +1,85 @@
...
+
+func NewHttpDebugInfoClient(serverUrl string) (*HttpDebuginfodClient, error) {
+	parsedUrl, err := url.Parse(serverUrl)
+	if err != nil {
+		return nil, err
+	}
+	return &HttpDebuginfodClient{UpstreamServer: parsedUrl}, nil
+}
+
+func NewObjectStorageDebugInfodClientCache(ctx context.Context, logger log.Logger, h *HttpDebuginfodClient) *ObjectStorageDebugInfodClientCache {
```
kakkoyun, line 50:
```suggestion
func NewObjectStorageDebugInfodClientCache(logger log.Logger, h *HttpDebuginfodClient) *ObjectStorageDebugInfodClientCache {
```

pkg/debuginfo/debuginfod.go
```diff
@@ -0,0 +1,85 @@
...
+	return &ObjectStorageDebugInfodClientCache{
+		ctx:    ctx,
+		logger: logger,
+		client: h,
+	}
+}
+
+func (c *ObjectStorageDebugInfodClientCache) GetDebugInfo(buildId string) (io.ReadCloser, error) {
```
kakkoyun, line 58:
```suggestion
func (c *ObjectStorageDebugInfodClientCache) GetDebugInfo(ctx context.Context, buildId string) (io.ReadCloser, error) {
```

pkg/debuginfo/debuginfod.go
```diff
@@ -0,0 +1,85 @@
...
+	debugInfo, err := c.client.GetDebugInfo(buildId)
+	if err != nil {
+		return nil, ErrDebugInfoNotFound
+	}
+
+	path := buildId + "/debuginfo"
+
+	err = c.bucket.Upload(c.ctx, path, debugInfo)
```
kakkoyun, line 66:
```suggestion
	err = c.bucket.Upload(ctx, path, debugInfo)
```

pkg/debuginfo/debuginfod.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil, err
+	}
+	return debugInfo, nil
+}
+
+func (c *HttpDebuginfodClient) GetDebugInfo(buildID string) (io.ReadCloser, error) {
+	buildIdUrl := *c.UpstreamServer
+	buildIdUrl.Path = path.Join(buildIdUrl.Path, buildID, "debuginfo")
+
+	resp, err := http.Get(buildIdUrl.String())
```
kakkoyun, line 78: Better to utilize the context here as well.

For example: 
https://gist.github.com/superbrothers/dae0030c151d1f3c24311df77405169b

kakkoyun (commented):
Looks great. Good job.

I have some suggestions to improve context handling.

pkg/debuginfo/debuginfod.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil, err
+	}
+	return &HttpDebuginfodClient{UpstreamServer: parsedUrl}, nil
+}
+
+func NewObjectStorageDebugInfodClientCache(ctx context.Context, logger log.Logger, h *HttpDebuginfodClient) *ObjectStorageDebugInfodClientCache {
+	return &ObjectStorageDebugInfodClientCache{
+		ctx:    ctx,
+		logger: logger,
+		client: h,
```
brancz, line 54: I think you forgot to set the `bucket` here.

pkg/debuginfo/debuginfod.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+}
+
+func (c *ObjectStorageDebugInfodClientCache) GetDebugInfo(buildId string) (io.ReadCloser, error) {
+	debugInfo, err := c.client.GetDebugInfo(buildId)
+	if err != nil {
+		return nil, ErrDebugInfoNotFound
+	}
+
+	path := buildId + "/debuginfo"
```
brancz, line 64: I think this deserves a prefix such as `/debuginfod-cache/`.

pkg/debuginfo/debuginfod.go
```diff
@@ -0,0 +1,85 @@
...
+func NewObjectStorageDebugInfodClientCache(ctx context.Context, logger log.Logger, h *HttpDebuginfodClient) *ObjectStorageDebugInfodClientCache {
+	return &ObjectStorageDebugInfodClientCache{
+		ctx:    ctx,
+		logger: logger,
+		client: h,
+	}
+}
+
+func (c *ObjectStorageDebugInfodClientCache) GetDebugInfo(buildId string) (io.ReadCloser, error) {
+	debugInfo, err := c.client.GetDebugInfo(buildId)
```
brancz, line 59: We should first check whether we already have this build ID in object storage, otherwise, we'll download it from upstream and upload it to object storage every time.

pkg/debuginfo/debuginfod.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	path := buildId + "/debuginfo"
+
+	err = c.bucket.Upload(c.ctx, path, debugInfo)
+	if err != nil {
+		level.Error(c.logger).Log("msg", "failed to upload to debuginfod cache", "err", err)
+		return nil, err
+	}
+	return debugInfo, nil
```
brancz, line 71: Uploading the `debufInfo` will have read the entire `io.Reader`, therefore this reader is already exhausted. We need to read it at this point from object storage.

Sylfrena:
I am not sure I understand why exactly the tests are failing o.o

brancz:
that one is just a flaky test, feel free to re-run every time you see it (RE https://github.com/parca-dev/parca/issues/499)

pkg/debuginfo/debuginfod.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	return &DebugInfodClientObjectStorageCache{
+		logger: logger,
+		client: h,
+		bucket: bucket,
+	}, nil
+}
+
+func (c *DebugInfodClientObjectStorageCache) GetDebugInfo(ctx context.Context, buildID string) (io.ReadCloser, error) {
+	path := buildID + "/debuginfod-cache/debuginfo"
```
brancz, line 78:
```suggestion
	path := "/debuginfod-cache/" + buildID + "/debuginfo"
```

pkg/symbolizer/symbolizer_test.go
```diff
@@ -411,23 +412,32 @@ func setup(t *testing.T) (*grpc.ClientConn, *debuginfo.Store, metastore.ProfileM
...
+		},
+		Cache: &debuginfo.CacheConfig{
+			Type: debuginfo.FILESYSTEM,
+			Config: &debuginfo.FilesystemCacheConfig{
+				Directory: cacheDir,
+			},
+		},
+	}
+
+	httpDebugInfodClient, err := debuginfo.NewHTTPDebugInfodClient(logger, "https://debuginfod.systemtap.org", 4*time.Millisecond)
```
brancz, line 430: 4 milliseconds timeout seems very very low for a download that might be several 100s of megabytes large. I would rather make this something like 2 or 5 minutes.

pkg/debuginfo/debuginfod.go
```diff
@@ -0,0 +1,127 @@
+// Copyright 2021 The Parca Authors
```
brancz, line 1: this is a new file so the year should be 2022

brancz (commented):
I think this is suuuuuuuper close. Very nice job! 💯

pkg/symbolizer/symbolizer_test.go
Sylfrena: so, I should change the default value for the `DebugInfodHTTPRequestTimeout` flag as well?

kakkoyun: I think this is just for the tests, no?

pkg/parca/parca.go
```diff
@@ -71,6 +71,9 @@ type Flags struct {
 	SymbolizerNumberOfTries int    `default:"3" help:"Number of tries to attempt to symbolize an unsybolized location"`
 
 	Metastore string `default:"badgerinmemory" help:"Which metastore implementation to use" enum:"sqliteinmemory,badgerinmemory"`
+
+	UpstreamDebuginfodServer     string        `default:"https://debuginfod.systemtap.org" help:"Upstream private/public server for debuginfod files. Defaults to https://debuginfod.systemtap.org."`
+	DebugInfodHTTPRequestTimeout time.Duration `default:"5m" help:"Timeout duration for HTTP request to upstream debuginfod server. Defaults to 5m"`
```
kakkoyun, line 76: :+1:

pkg/symbolizer/symbolizer_test.go
brancz: oh whops my bad I don't know how I didn't see that this was a test

PR #622, opened 2022-02-14
Add CRA support to the go server and cleaned up some next-js config files.

manojVivek:
- [x] Ensure go server works with the cra build
- [x] Added an option to support custom path prefix for parca UI
- [x] clean-up unwanted nextjs config files
- [x] ~~Create a generic `useRouter` hook that the UI components can use to support both `next/react` router.~~ Not feasible, as `react-router` hooks throw an error when accessed without a `ReactRouter` context.
- [x] Refactor the UI components to take generic props for router functionality and let the parent handle the router specific invocation, as the above is not feasible.


manojVivek:
@yomete The idea that we discussed, creating a generic `useRouter` hook to enable the UI components to work seamlessly with `next/react` router, didn't work as expected. 

I tried something like the following: 
```javascript
import {useRouter as useNextRouter} from 'next/router';
import {useLocation as useRRLocation} from 'react-router-dom`;

const useRouter = () => {
  const nextRouter = useNextRouter();
  const rrLocation = useRRLocation();

  if (nextRouter !== null) {
    return {queryParams: nextRouter.queryparams, navigateTo: ...};
  }

  if (rrLocation !== null {
    return {queryParams: rrLocation.search, navigateTo: ...};
  }
  console.warn('No supported router detected in the app');
  return null;
}

export default useRouter;
```

The `next/router` worked nicely with this when tried on both `nextjs` app and `CRA` app. So, that was fine. But the `react-router-dom` router didn't work when tried on the `nextjs` app. It threw an error that it cannot be used without the `ReactRouter` context and there was no option to suppress the error and return a `null` instead, like how the `next/router` handles the situation. 

So unfortunately, we won't be able to follow this approach. Alternatively, I have refactored the UI components to take the router data and functions as generic prop items so that the caller takes care of handling the router-specific invocation. Here is the commit that does it: https://github.com/parca-dev/parca/pull/622/commits/6c564129ca8cefd8086cd9cd033ad0a67a813f34

Please let me know your thoughts.

yomete:
> @yomete The idea that we discussed, creating a generic `useRouter` hook to enable the UI components to work seamlessly with `next/react` router, didn't work as expected.
> 
> I tried something like the following:
> 
> ```js
> import {useRouter as useNextRouter} from 'next/router';
> import {useLocation as useRRLocation} from 'react-router-dom`;
> 
> const useRouter = () => {
>   const nextRouter = useNextRouter();
>   const rrLocation = useRRLocation();
> 
>   if (nextRouter !== null) {
>     return {queryParams: nextRouter.queryparams, navigateTo: ...};
>   }
> 
>   if (rrLocation !== null {
>     return {queryParams: rrLocation.search, navigateTo: ...};
>   }
>   console.warn('No supported router detected in the app');
>   return null;
> }
> 
> export default useRouter;
> ```
> 
> The `next/router` worked nicely with this when tried on both `nextjs` app and `CRA` app. So, that was fine. But the `react-router-dom` router didn't work when tried on the `nextjs` app. It threw an error that it cannot be used without the `ReactRouter` context and there was no option to suppress the error and return a `null` instead, like how the `next/router` handles the situation.
> 
> So unfortunately, we won't be able to follow this approach. Alternatively, I have refactored the UI components to take the router data and functions as generic prop items so that the caller takes care of handling the router-specific invocation. Here is the commit that does it: [6c56412](https://github.com/parca-dev/parca/commit/6c564129ca8cefd8086cd9cd033ad0a67a813f34)
> 
> Please let me know your thoughts.

I think it's a good solution providing what you described with react-router-dom not working with the useRouter hook. We might have to move the `ProfileExplorer` component to be a `shared` component in the future and that might make things interesting (with the cloud product using nextjs) but we can tackle that when it's time.

yomete:
I think it's a good solution providing what you described with react-router-dom not working with the useRouter hook. We might have to move the ProfileExplorer component to be a shared component in the future and that might make things interesting (with the cloud product using nextjs) but we can tackle that when it's time.

yomete:
love it! let me know when it's ready for review :)

manojVivek:
It's ready now. 

Once this is merged, we just need to sync with `main` and take care of the new commits.

yomete:
ok so we merge this into the `move-nextjs-to-cra` branch, rebase with the `main` branch and then we can can ask others for review. sounds good?

monicawoj:
> Including and using `ProfileExplorer` in a NextJS app is still totally possible after [this commit](https://github.com/parca-dev/parca/commit/6c564129ca8cefd8086cd9cd033ad0a67a813f34), with something like the following.
> 
> ```js
> import {useRouter} from 'next/router';
> 
> const Page = () => {
>   const router = useRouter();
> 
>   return (
>   <ProfileExplorer
>     queryClient={queryClient}
>     queryParams={Object.fromEntries(Object.entries(router.query))}
>     navigateTo={(pathname, search) => router.push({pathname, search})} />
>   )
> }
> ```
> 
> Just that it is not as smooth as what I was expecting with the generic `useRouter` hook. 😁

@manojVivek / @yomete - just out of curiosity, why do we need to do this calculation `Object.fromEntries(Object.entries(router.query))` for the query params instead of just providing `router.query`?

manojVivek:
> ok so we merge this into the `move-nextjs-to-cra` branch, rebase with the `main` branch and then we can can ask others for review. sounds good?

Yes, let's do it. 👍🏼

manojVivek:
> just out of curiosity, why do we need to do this calculation `Object.fromEntries(Object.entries(router.query))` for the query params instead of just providing `router.query`?

I'm not 100% sure as well, I just quoted it from the usage here: https://github.com/parca-dev/parca/pull/610/files#diff-fb2db6a18b0f0efbd7cfcadd2095cfe5ced565a8fb7b340bab40f9c7b61229b2L36

Probably to create a clone of the router.query object? 🤔

yomete:
good catch @monicawoj it's actually the same output as just using `router.query`. I can make a commit to make it just be `const queryParams = router.query`

PR #633, opened 2022-02-14
fix(grafana): Fix datasource template

maxbrunet:
The template name had likely been overwritten by the Grafana export.

I have done that so many times, I have written scripts a couple years ago, now in a Gist if that helps anyone:
https://gist.github.com/maxbrunet/e96dee35f7e5fcd73a98929f4bc9c2bb


metalmatze (approved):
Makes sense. Thanks! :+1:

PR #628, opened 2022-02-14
pkg/server: JS bundle files template error fix

manojVivek:
Fixes the syntax error that happens with the js bundle files, while building parca binary.

PR #635, opened 2022-02-15
pkg/server: Temp fix for the version string on header

manojVivek:
A follow-up on https://github.com/parca-dev/parca/pull/628#issuecomment-1039891237. 
I changed the pattern to `*_app-*`, as the `.html` suffix filter is breaking the version string on Nextjs app. We can use this as a temporary fix until CRA migration PR is merged.


manojVivek:
> Since we now merged the changes in #598 this needs a small rebase.

Done.

metalmatze (approved):

PR #659, opened 2022-02-22
pkg/columnstore: Add comments after reading/debugging through the code

metalmatze:
Debugged my way through the code. Most of it is super readable. Kudos!
The thing that's most unclear to me as of right now, is how a Part and a Granule relate. It'd be great to add a comment describing why we need the Parts.


pkg/columnstore/part.go
```diff
@@ -9,11 +9,13 @@ type Row struct {
 	Values []interface{}
 }
 
+// Part is a sub set of a Granule? TODO: How do they relate?
```
thorfour, line 12: A part is an ordered set of rows that sort into this granule, but haven't yet been merged into the sorted portion of the Granule.

pkg/columnstore/part.go
```diff
@@ -9,11 +9,13 @@ type Row struct {
...
 type Part struct {
 	columns     []Iterable
 	Cardinality int
 
-	// transaction id that this part was indserted under
+	// transaction id that this part was inserted under.
+	// TODO: Rename to insertTX?
```
thorfour, line 18: correct

metalmatze:
Closing since the target branch has been rebased and this was only comments anyway.

PR #667, opened 2022-02-23
Remove granule lock

thorfour:
~There's still a small race condition on index swap that we either need to retry or unmark the granule as pruned.~


pkg/columnstore/granule.go
```diff
@@ -12,22 +13,21 @@ import (
...
-	card int
+	card uint64
 
 	schema *Schema
 
 	granulesCreated prometheus.Counter
 
 	// pruned indicates if this Granule is longer found in the index
-	pruned bool
+	pruned uint64
```
metalmatze, line 30: Before `pruned` was a simple bool for which the comment made sense. I don't really get what a number between 0 and maxUInt64 now says. Can we update the comment? Thanks!

pkg/columnstore/granule.go
```diff
@@ -62,62 +63,67 @@ func NewGranule(granulesCreated prometheus.Counter, schema *Schema, parts ...*Pa
...
+	newcard := atomic.AddUint64(&g.card, uint64(p.Cardinality))
 	it := p.Iterator()
 
 	if it.Next() {
-		r := Row{Values: it.Values()}
-		if g.schema.RowLessThan(r.Values, g.least.Values) {
-			g.least = r
+		for {
+			r := &Row{Values: it.Values()}
+			least := atomic.LoadPointer((*unsafe.Pointer)(unsafe.Pointer(&g.least)))
```
metalmatze, line 75: Nice, lots of unsafe. :smile: 
Would be good to document why we need the magic here.

pkg/columnstore/granule.go
```diff
@@ -62,62 +63,67 @@ func NewGranule(granulesCreated prometheus.Counter, schema *Schema, parts ...*Pa
...
 
 	// How many granules we'll need to build
-	count := g.parts[0].Cardinality / n
+	count := 0
+	var it *PartIterator
+	g.parts.Iterate(func(p *Part) bool {
+		count = p.Cardinality / n
+		it = p.Iterator()
+		return false
+	})
```
metalmatze, line 122: Love the closure approach here! :tada:

brancz (approved):
very nice!

PR #668, opened 2022-02-23
chore: Update Readme file for UI

yomete:
I removed all instances of NextJS and instructions that were specific to NextJS. I also updated to reflect that we now use Create React App.

### DIFF
--- a/ui/README.md
+++ b/ui/README.md
@@ -1,15 +1,13 @@
 # Parca UI
 
-This is a [Next.js](https://nextjs.org/) project that utilizes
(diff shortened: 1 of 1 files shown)


yomete:
fyi @manojVivek @monicawoj I updated the readme here to reflect changes to our UI setup.

monicawoj (approved):
🚀 lgtm - very helpful!

manojVivek:
Should we also add an empty `build` folder inside `web`, to prevent the error that Monica faced when she built the go binary without building the `ui` or is that alright?

yomete:
Alright. I'll add that 👍🏾

PR #672, opened 2022-02-24
Fix zindex issue with tooltips and MatcherInput dropdown

### LINKED ISSUE #664: Tooltips are getting hidden beneath the buttons in the merged view
And maybe we should also choose a solid color(without any transparency) as a background for the tooltips. It makes it hard to read the tooltip text when it is over another text element. 

Screenshot 2022-02-23 at 10 12 04 AM

yomete:
Resolves #664


yomete:
@manojVivek @monicawoj please take a look when you can

metalmatze:
I'd be great to sent this against the `release-0.8` branch so we can include it into v0.8.1.

brancz:
Looks like we already merged https://github.com/parca-dev/parca/pull/666 into main which would also need to go out as a bugfix, I think we should just cut 0.8.1 from main or just go straight to 0.9

PR #674, opened 2022-02-24
Update CONTRIBUTING.md to include bug fixes and Go formatting

metalmatze:


CONTRIBUTING.md
```diff
@@ -85,6 +85,21 @@ This is roughly what the contribution workflow should look like:
...
+```bash
+git checkout release-0.8
+git checkout --branch my-fix
+# start working on your fixes
+git add .
+git commit -m "Fix xyz"
+git push origin my-fix
+```
+
+Then open your Pull Request and make sure to select `release-0.8` as the base to send your changes on to.
```
brancz, line 100: I'm not against this, but at the same time, I think moving forward is more important than maintaining patch releases. So I think if the only patch release we would be putting a fix in is the next patch release of the current latest minor release, we might as well just merge it into `main` and cut the next release and decide based on what's there whether to call it the next minor release or the next patch release.

metalmatze: Yep, not against doing that at all. 
The only problem I can see is introducing a new feature in between and then not being able to "just" send out a bug fix without that new bigger feature. Depends on the feature and situation for sure.

brancz: Agreed

PR #675, opened 2022-02-24
symbolizer: Make sure vendored debug/elf used

kakkoyun:


brancz:
phew, good catch

PR #676, opened 2022-02-24
area/ui: Fix selecting multiple labels breaking the UI

### LINKED ISSUE #662: Selecting multiple labels breaks the UI
Selecting on multiple labels (while holding shitft and then clicking) breaks the UI at the second addition.
Example URL: 
https://demo.parca.dev/?currentProfileView=icicle&expression_a=parca_agent_cpu_samples_count{namespace=%22parca%22,%20container=%22parca-agent%22}&from_a=1645542723562&to_a=1645546323562&merge_a=false&time_selection_a=relative:hour|1

In the console I can see:
```
TypeError: e.startsWith is not a function
    at main.5e7506f4.js:2:1110785
    at e.value (main.5e7506f4.js:2:1110861)
    at tg (main.5e7506f4.js:2:1205259)
    at si (main.5e7506f4.js:2:919481)
    at Qs (main.5e7506f4.js:2:971781)
    at xu (main.5e7506f4.js:2:959022)
    at Du (main.5e7506f4.js:2:958950)
    at Ru (main.5e7506f4.js:2:958813)
    at bu (main.5e7506f4.js:2:955779)
    at du (main.5e7506f4.js:2:953127)
```
Maybe it's due to the URL params not being properly url-encoded? Or would that be another issue?

yomete:
Fixes #662

Two things were responsible for the bug here. The query parameter  `expression_a` was being interpreted as an array whenever there are multiple labels and it was not being encoded. This PR checks if the query param is an array, if it is, it converts it to a string and then encodes them in the fns for selecting profiles and queries.

### DIFF
--- a/ui/packages/app/web/src/components/ProfileExplorer.tsx
+++ b/ui/packages/app/web/src/components/ProfileExplorer.tsx
@@ -17,10 +17,7 @@ const ProfileExplorer = ({
   queryParams,
   navigateTo,
 }: ProfileExplorerProps): JSX.Element => {
-  /* eslint-disable */
-  // Disable eslint due to params being snake case
   const {
-    expression_a,
     from_a,
     to_a,
     merge_a,
@@ -28,7 +25,6 @@ const ProfileExplorer = ({
     time_a,
     time_selection_a,
     compare_a,
-    expression_b,
     from_b,
     to_b,
     merge_b,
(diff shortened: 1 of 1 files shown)


yomete:
would be nice to get a review on this @manojVivek @monicawoj

PR #677, opened 2022-02-24
area/ui: Fix z-index issues with tooltips

### LINKED ISSUE #664: Tooltips are getting hidden beneath the buttons in the merged view

yomete:
Fixes #664


metalmatze:
Cool! Re-running failed tests.

yomete:
Actually the "Publish" check is expected to fail, it only works when a PR is merged into `main`. I'm even surprised it got picked up for this PR.

brancz:
Probably a CI configuration, let's not worry about the CI configuration as part of this PR then.

PR #679, opened 2022-02-28
ui/datePicker: Date Range Picker uses UTC as the timezone

### LINKED ISSUE #663: Time picker uses local browser time vs UTC in graphs
For now, we probably only want to support UTC across all parts of the UI. Hence, we should update the date time picker to use UTC as well. 
Threw me off when entering time in absolute after hovering the metrics.

manojVivek:
Fixes #663.

Added a small note as well to indicate it is in UTC:
Screenshot 2022-02-28 at 10 48 35 AM


brancz (approved):
Perfect!

PR #680, opened 2022-02-28
feat: remove threadcreate profiles

### LINKED ISSUE #670: remove threadcreate profiles from default scraping
threadcreate profiles have been broken in Go since 2013, and it doesn't seem like they'll be fixed in any shape or form anytime soon.

For the sake of not confusing users, we should remove threadcreate from being scraped by default. (it also triggered confusion causing us to ignore empty stack traces, eg. #666)

https://github.com/golang/go/issues/6104

me-diru:
fixes #670

Signed-off-by: Rohit Dandamudi <rohitdandamudi.1100@gmail.com>

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -35,12 +35,11 @@ import (
 )
 
 const (
-	pprofMemoryTotal       string = "memory_total"
-	pprofBlockTotal        string = "block_total"
-	pprofGoroutineTotal    string = "goroutine_total"
-	pprofMutexTotal        string = "mutex_total"
-	pprofProcessCPU        string = "process_cpu"
-	pprofThreadcreateTotal string = "threadcreate_total"
+	pprofMemoryTotal    string = "memory_total"
+	pprofBlockTotal     string = "block_total"
+	pprofGoroutineTotal string = "goroutine_total"
+	pprofMutexTotal     string = "mutex_total"
+	pprofProcessCPU     string = "process_cpu"
 )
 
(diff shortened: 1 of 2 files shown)


me-diru:
there are a few references to threadcreate in the frontend as well, not sure if there should be removed as well.

pkg/config/config.go
```diff
@@ -35,12 +35,11 @@ import (
 )
 
 const (
-	pprofMemoryTotal       string = "memory_total"
-	pprofBlockTotal        string = "block_total"
-	pprofGoroutineTotal    string = "goroutine_total"
-	pprofMutexTotal        string = "mutex_total"
-	pprofProcessCPU        string = "process_cpu"
-	pprofThreadcreateTotal string = "threadcreate_total"
```
me-diru, line 43: removed this as well. as it is not used anymore.

me-diru: why does it show that I removed all the variables 😅

brancz (approved):
I think this way is fine. We don't want to prevent people from scraping them if they really want to but remove them from the default, so this is exactly what we wanted. Thank you so much for taking care of this!

pkg/config/config.go
brancz: whitespace changes...

metalmatze:
Amazing work! :rocket: 
Seems like an unrelated unit test is flaky... 
Merging this as everything else is completely fine!

manojVivek:
@all-contributors please add @me-diru for code.

PR #691, opened 2022-03-01
ui/packages/shared/profile: Use similar function from flamegraph to build node label

metalmatze:
The label (name) of each top table row now uses the same function as the flame graph:
labels


brancz (approved):
Love it!

PR #693, opened 2022-03-02
Add buf GitHub Actions

metalmatze:
This will make sure our proto is all good on PRs. When there are changes on main those will be automatically uploaded to buf's BSR.


kakkoyun (approved):
LGTM

PR #696, opened 2022-03-02
Fix loading animation

### LINKED ISSUE #260: Fix loading animations
Only on first load do we show a loading animation, not on subsequent ones, this should be fixed.

monicawoj:
resolves #260 

As mentioned in the issue, I believe the best approach for a good user experience would be to show the loader only if a request is taking longer than a user-noticable time. In this PR, I suggest doing so after 500ms (half a second), but that can be lengthened/shortened as needed.


yomete:
lgtm!

PR #699, opened 2022-03-02
pkg/profilestore: Add a Parca scraper-only mode that forwards via gRPC

### LINKED ISSUE #629: Add Parca storage-less scraper
Next to having the Parca agent creating pprof profiles from eBPF and then sending it off via gRPC we should also support a mode of Parca which can scrape profiles from pprof endpoints and then immediately send them off via gRPC not having a local storage.
It's probably similar to the Prometheus agent mode.

metalmatze:
If there is a central place (cluster) to store profiles we still might want to collect pprof profiles and then forward them via gRPC as written in #629. 
To be able to do that we now support a `--mode=scraper` (instead of the current only `--mode=all`). 
Possibly making this a sub command instead of just a flag might be better. 

The UI is broken in --mode=scraper as only the API for the Targets page is available. We should hide the "Profiles" page in that case. Most likely we'll need to communicate the mode from the backend to the frontend for that to happen.

Closes #629

### DIFF
--- a/pkg/parca/parca.go
+++ b/pkg/parca/parca.go
@@ -15,15 +15,18 @@ package parca
 
 import (
 	"context"
+	"crypto/tls"
 	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
+	"strings"
 	"syscall"
 	"time"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
+	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	"github.com/oklog/run"
 	"github.com/prometheus/client_golang/prometheus"
@@ -36,6 +39,8 @@ import (
 	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
 	"go.opentelemetry.io/otel/trace"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
 	"gopkg.in/yaml.v2"
 
 	debuginfopb "github.com/parca-dev/parca/gen/proto/go/parca/debuginfo/v1alpha1"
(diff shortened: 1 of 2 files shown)


metalmatze:
cc @yeya24

brancz (approved):
Love how simple this is thanks to everything being gRPC.

lgtm 🎉

brancz:
Maybe the mode should be `scraper-only`? Just to make it crystal clear what the distinction is.

metalmatze:
Yes, right? :joy: 
During the implementation (on Twitch stream) I was surprised as well that we're already done. :smile:

pkg/parca/parca.go
```diff
@@ -108,6 +120,10 @@ func Run(ctx context.Context, logger log.Logger, reg *prometheus.Registry, flags
 		return err
 	}
 
+	if flags.Mode == "scraper" {
```
yeya24, line 123: Small nit, maybe make this a const?

yeya24 (commented):
Amazing! I am surprised by how easy it is to implement this feature 👍 
One thing to clarify about the deployment model:
I have one central cluster and multiple edge clusters. At edge cluster, I just have to deploy scraper mode parca and sent data to a central parca, right?

pkg/parca/parca.go
```diff
@@ -75,6 +81,12 @@ type Flags struct {
 
 	UpstreamDebuginfodServer     string        `default:"https://debuginfod.systemtap.org" help:"Upstream private/public server for debuginfod files. Defaults to https://debuginfod.systemtap.org."`
 	DebugInfodHTTPRequestTimeout time.Duration `default:"5m" help:"Timeout duration for HTTP request to upstream debuginfod server. Defaults to 5m"`
+
+	StoreAddress       string `kong:"help='gRPC address to send profiles and symbols to.'"`
+	BearerToken        string `kong:"help='Bearer token to authenticate with store.'"`
+	BearerTokenFile    string `kong:"help='File to read bearer token from to authenticate with store.'"`
+	Insecure           bool   `kong:"help='Send gRPC requests via plaintext instead of TLS.'"`
+	InsecureSkipVerify bool   `kong:"help='Skip TLS certificate verification.'"`
```
yeya24, line 89: Be able to specify external labels would be better. That way I can filter series by external labels.

brancz: 💯

pkg/parca/parca.go
metalmatze: Yes, I was thinking the same while implementing. Sadly we still have an occurrence in the flag kong tag that won't be possible to be a constant. I'll do that anyway. :+1:

pkg/parca/parca.go
metalmatze: Yes, totally! 
That will be a follow-up PR.

kakkoyun (approved):
LGTM :rocket:

PR #706, opened 2022-03-03
pkg/scrape: Add externalLabels for scrape-only mode

metalmatze:
The `--mode=scraper-only` now adds support for external labels with a flag like `--external-label="foo=bar;node=metalmatze-desktop"`

I setup two Parca instances and the scraper-only one send to the other Parca instance where I could see the given external labels show up.

Screenshot from 2022-03-03 18-00-38

cc @yeya24 for trying this out on your machines once we merge to `main` have a container to try.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -127,5 +127,13 @@ dev/down:
 tmp/help.txt: go/bin
 	bin/parca --help > $@
 
-README.md: tmp/help.txt
-	embedmd -w README.md
+embedmd:
+ifeq (, $(shell which embedmd))
+	go install github.com/campoy/embedmd@latest
+EMBEDMD=$(GOBIN)/embedmd
+else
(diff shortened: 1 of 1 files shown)


pkg/scrape/scrape.go
```diff
@@ -464,6 +474,12 @@ mainLoop:
 					Value: l.Value,
 				})
 			}
+			for _, l := range sl.externalLabels {
```
brancz, line 477: We need to maintain a sorted order, I think it would be better to add these labels before we call `sort.Sort` above.

metalmatze:
I have no clue why the docs CI fails now... I added the `embedmd` to be installed on the fly too. Weird.

brancz:
it looks like it's building the parca binary to template the help text, but we don't build the frontend, so it errors on embedding the frontend

Issue #707, opened 2022-03-03
Missing step(s) in CONTRIBUTING.md


importhuman:
The [getting started](https://github.com/parca-dev/parca/blob/main/CONTRIBUTING.md#getting-started) section for 

```
To profile all containers using Kubernetes, the parca-server can be run alongside parca-ui using Tilt.
```

doesn't mention the need to run `make dev/setup`. A response similar to the attached image is obtained if this step is not carried out.

![Screenshot from 2022-03-03 23-14-44](https://user-images.githubusercontent.com/69148722/156621753-86d94cdc-23a8-4d7d-b09a-d6f4c4485e8f.png)

My internet isn't working too well today so I haven't been able to complete the setup yet, will try tomorrow and update this issue if anything else is needed. 

Will also submit the required PR.

kakkoyun:
Thanks for pointing out. We should have done a better job for the docs. In case you haven't seen, we have a couple of setup scripts https://github.com/parca-dev/parca/blob/main/env.sh, https://github.com/parca-dev/parca/blob/main/env-jsonnet.sh

Looking forward to your PR. Let us know if you need anthing.

importhuman:
Yes, the setup scripts are run by `make dev/setup` so there's no need to mention it separately in the doc.
Rest of the setup looks good, no issues encountered.

One question I have is, the commit log looks like PRs are made from forks of the repo, but the contributing document doesn't ask to do so. Which path should I follow? Will add the fork step to the doc if that's required.

brancz:
I think there's a bit of a mix happening. When maintainers are collaborating on a feature branch it might be pushed to the main repo, but generally speaking, development happens in forks.

PR #709, opened 2022-03-04
pkg: Rename FlatProfile to Profile

### LINKED ISSUE #608: Rename FlatProfile to Profile
Since we have made the transition from the profile trees to the flat profiles and in the meantime deleted the `Profile` we're just left with the intermediate `FlatProfile`. 
We should finalize the refactoring by renaming `FlatProfile` to `Profile` since it might not be as obvious anymore what makes it _flat_.

https://github.com/parca-dev/parca/blob/1b2cb213b9f227283cccb7c679fc35acc466476b/pkg/profile/profile.go#L58-L66

From there most things should fall in place and can be discussed on a opened PR.

importhuman:
Closes #608

Please let me know if there's any contributing guideline I missed. I had some confusion regarding repo forks that I mentioned in [an issue](https://github.com/parca-dev/parca/issues/707#issuecomment-1058937048), but I've gone ahead with forking this repo, adding the remote, pushing to it, and opening the PR here.

In this commit, I've only refactored the `FlatProfile` struct to `Profile` and wherever it's used. There are still some places which include the name `FlatProfile`, shall I refactor them as well? This would also involve removing the [InstantFlatProfile interface](https://github.com/parca-dev/parca/blob/8ad0e103287156104c4bd0d977efc64fdbbf8056/pkg/profile/profile.go#L59), as there is already an `InstantProfile` interface with the same methods.

### DIFF
--- a/pkg/profile/profile_flat.go
+++ b/pkg/profile/profile_flat.go
@@ -24,23 +24,23 @@ import (
 	"github.com/parca-dev/parca/pkg/metastore"
 )
 
-// FlatProfilesFromPprof extracts a Profile from each sample index included in the pprof profile.
-func FlatProfilesFromPprof(ctx context.Context, l log.Logger, s metastore.ProfileMetaStore, p *profile.Profile) ([]*Profile, error) {
-	fps := make([]*Profile, 0, len(p.SampleType))
+// ProfilesFromPprof extracts a Profile from each sample index included in the pprof profile.
+func ProfilesFromPprof(ctx context.Context, l log.Logger, s metastore.ProfileMetaStore, p *profile.Profile) ([]*Profile, error) {
+	ps := make([]*Profile, 0, len(p.SampleType))
 
 	for i := range p.SampleType {
-		fp, err := FlatProfileFromPprof(ctx, l, s, p, i)
+		p, err := ProfileFromPprof(ctx, l, s, p, i)
 		if err != nil {
 			return nil, err
 		}
-		if fp != nil {
-			fps = append(fps, fp)
+		if p != nil {
+			ps = append(ps, p)
 		}
 	}
-	return fps, nil
+	return ps, nil
 }
 
-func FlatProfileFromPprof(ctx context.Context, logger log.Logger, metaStore metastore.ProfileMetaStore, p *profile.Profile, sampleIndex int) (*Profile, error) {
+func ProfileFromPprof(ctx context.Context, logger log.Logger, metaStore metastore.ProfileMetaStore, p *profile.Profile, sampleIndex int) (*Profile, error) {
 	pfn := &profileFlatNormalizer{
 		logger:    logger,
 		metaStore: metaStore,
--- a/pkg/profile/profile_flat.go
+++ b/pkg/profile/profile_flat.go
@@ -29,7 +29,7 @@ func ProfilesFromPprof(ctx context.Context, l log.Logger, s metastore.ProfileMet
 	ps := make([]*Profile, 0, len(p.SampleType))
 
 	for i := range p.SampleType {
-		p, err := ProfileFromPprof(ctx, l, s, p, i)
+		p, err := FromPprof(ctx, l, s, p, i)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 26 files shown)


importhuman:
The tests are failing because of ["no matching files found"](https://github.com/parca-dev/parca/runs/5445208794?check_suite_focus=true#step:4:305) and ["gofumpt not found"](https://github.com/parca-dev/parca/runs/5445209004?check_suite_focus=true#step:5:7). Is there anything needed on my end to fix these?

metalmatze:
Yes, those were two broken CI steps... They should have been fixed in the meantime. To resolve this for your PR could you please rebase against the current main branch once more? 
This looks really good! 🚀

metalmatze:
Seems like something with your rebase went wrong... 
Did you rebase like this?
```
git checkout main
git pull origin main
git checkout rename-flatprofile
git rebease main
```

importhuman:
Pulled from upstream (this repo, not my fork) instead of origin. Then I tried to push to origin, got an error that origin had changes I didn't, so pulled those changes, then finally pushed.

What should I do instead, and how do I resolve it in this case?

metalmatze:
Ok so in your case it sounds like this:
```
# Let's make sure we're talking about the same origin and upstream
git remote set-url upstream git@github.com:parca-dev/parca.git
git remote set-url origin git@github.com:importhuman/parca.git

git checkout main
git pull upstream main
git checkout rename-flatprofile
git rebase main
git push origin rename-flatprofile --force
```
I'm not 100% if that resolves the changes you did just before now, but let's try if you can get the amount of commits in this PR from 66 down to the 1-2 you actually did. Let me know how it goes.

importhuman:
This fixed it, though I can't pin down what exactly went wrong the first time. Perhaps it had to do with the fork having changes I didn't, me pulling them _after_ I rebased, then pushing, which led to all the commits reflecting in this PR?

Thanks for your help! Would you like me to refactor other instances of `FlatProfile` in the codebase as well? Currently only the instances that used the `FlatProfile` struct have been renamed.

metalmatze:
> This fixed it, though I can't pin down what exactly went wrong the first time. Perhaps it had to do with the fork having changes I didn't, me pulling them _after_ I rebased, then pushing, which led to all the commits reflecting in this PR?

Yes that sounds like something that could cause this indeed!

> Thanks for your help! Would you like me to refactor other instances of `FlatProfile` in the codebase as well? Currently only the instances that used the `FlatProfile` struct have been renamed.

Yes, that would be great. We can either merge this one as it is and continue in another PR or add the other changes to this one. Whatever you prefer! :blush:

importhuman:
Have refactored other instances of `FlatProfile`. 

Haven't done it for [GenerateFlatPprof](https://github.com/parca-dev/parca/blob/29ae58832107f178239f66a080b744733311a150/pkg/query/pprof.go#L65) (and related instances), as I couldn't tell whether the "Flat" refers to `FlatProfile` or something else. My previous commit didn't modify the return type for this function, only a [passed value in a test](https://github.com/parca-dev/parca/pull/709/commits/575a88a5f220dc3bf7b06914d91ff77bc28180c0#diff-2fdeeaebb4d32eed7df45b2f29d6d13560e09e93efef22d0fc947ef7f4ea10f5), so I wasn't sure. Please let me know if this is to be modified.

Also, while running tests locally, `Test_QueryRange_Limited` [failed at L191](https://github.com/parca-dev/parca/blob/29ae58832107f178239f66a080b744733311a150/pkg/query/query_test.go#L191) (with expected value of 0 and actual of 10). I suspect this was because of the behaviour of `rand.Intn` at [L182](https://github.com/parca-dev/parca/blob/29ae58832107f178239f66a080b744733311a150/pkg/query/query_test.go#L182), since it picks from a 0 inclusive range. I ran the tests 2 more times, and they passed both times (although I couldn't figure out how to clear the cache, tried 
```
go test -v -clearcache `go list ./...`
```
and got the error `no Go files in <my-path>/parca`).

pkg/profile/profile_flat.go
```diff
@@ -24,23 +24,23 @@ import (
...
+		if p != nil {
+			ps = append(ps, p)
 		}
 	}
-	return fps, nil
+	return ps, nil
 }
 
-func FlatProfileFromPprof(ctx context.Context, logger log.Logger, metaStore metastore.ProfileMetaStore, p *profile.Profile, sampleIndex int) (*FlatProfile, error) {
+func ProfileFromPprof(ctx context.Context, logger log.Logger, metaStore metastore.ProfileMetaStore, p *profile.Profile, sampleIndex int) (*Profile, error) {
```
metalmatze, line 43: Let's rename this as golangci-lint points out the package name is already profile. :relaxed: 
```suggestion
func FromPprof(ctx context.Context, logger log.Logger, metaStore metastore.ProfileMetaStore, p *profile.Profile, sampleIndex int) (*Profile, error) {
```

metalmatze (commented):
If you could fix that last naming issue I also commented on then I'd say let's merge it now before we continue any longer.
The QueryRange is indeed flaky and if you have an idea how to make it better feel free to send another PR :wink: 
The same for the FlatGenerateFlamegraph. That one should be renamed as well, but let's do that in another PR. 

Amazing work! :100: :tada:

metalmatze:
Amazing! :tada: :rocket: 
Keep the PRs coming! :relaxed:

PR #718, opened 2022-03-08
feat: empty state container for targets page

### LINKED ISSUE #669: Targets page displays an empty page
The targets page displays an empty view on demo.parca.dev. This is because we're not scraping any targets but are actually receiving profiles being sent from parca-agent.

It'll be nice to show an empty state placeholder explaining this.

monicawoj:
resolves #669 

Creates a reusable EmptyState container component in the shared folder, which can be used throughout the application as a placeholder when data is not available

### DIFF
--- a/ui/packages/app/web/tailwind.config.js
+++ b/ui/packages/app/web/tailwind.config.js
@@ -10,6 +10,9 @@ module.exports = {
       maxWidth: {
         '1/2': '50%',
       },
+      minWidth: theme => ({
+        ...theme('spacing'),
+      }),
     },
   },
   variants: {},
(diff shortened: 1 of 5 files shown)


ui/packages/app/web/tailwind.config.js
```diff
@@ -10,6 +10,9 @@ module.exports = {
       maxWidth: {
         '1/2': '50%',
       },
+      minWidth: theme => ({
+        ...theme('spacing'),
```
manojVivek, line 14: Just curious, what does this do?

manojVivek (approved):
Lgtm!

ui/packages/app/web/tailwind.config.js
monicawoj: @manojVivek this extends the tailwind css potential values available for setting 'minimum width' by making them the same as the options available for setting 'width'. Min-width had very few available options. For example, 'min-w-64' was not available, but 'w-64' (which equates to 16rem) was. So basically now, all the options available for width are also available for min width.

metalmatze:
Amazing! 😊 🎉

ui/packages/app/web/tailwind.config.js
manojVivek: Got it, thanks for the explanation, Monica. 
I wonder why wouldn't they do it by default. 🤔

Issue #473, opened 2021-11-29
Label value autocompletion in UI


brancz:
The backend is already capable of returning label values for a given label name, but the UI is not currently using it to suggest it in autocompletion. This would be great in order to explore the data through suggestions.

yomete:
@brancz afaik, these are the possible labels we can have autocompletion for, lmk if i missed any or remove from the list as this is what I'm going to implement.

- `instance`
- `__address__`
- `__name__`
- `__profile_path__`
- `__scheme__`
- `job`

metalmatze:
There are two API calls that are made. 
First, we get all the names (which is basically the list that you posted) by calling `/api/parca.query.v1alpha1.QueryService/Labels` and then we can request all values for the given names by calling `/api/parca.query.v1alpha1.QueryService/Values`. The second call has a payload that by default requests all `__name__` values, but can be changed to whatever label name we need to auto complete.

yomete:
I see. thanks for explaining! I thought we had a fixed list but knowing it's all from an API makes it better

PR #743, opened 2022-03-10
area/ui: Add label-value autocompletion

### LINKED ISSUE #348: Autocompletion in parca UI
Will be nice to have autocompletion in parca ui, at least for namespaces in pods

yomete:
Resolves #348 #473 

The PR adds the label value autocompletion to the Parca UI. Parca users will now get a complete suggestion when trying to filter profiles. It also include little UI updates like label-value being displayed as tags after a complete suggestion.

image


ui/packages/app/web/src/components/ui/Card.tsx
```diff
@@ -21,10 +21,12 @@ const Card = ({children, className = ''}: CardProps) => {
   );
 };
 
-const Header = ({children}: {children: React.ReactNode}) => {
+const Header = ({children, className}: {children: React.ReactNode; className?: string}) => {
```
manojVivek, line 24: Should we also assign a default parameter value for `className` to be `''` as it is an optional, and we have a `className.length` below?

ui/packages/shared/components/src/MatchersInput/index.tsx
```diff
@@ -166,31 +263,85 @@ const MatchersInput = ({
...
+    let suggestion = getSuggestion(suggestionIndex);
+
+    if (suggestion.type === 'labelValue') {
+      suggestion.value = addQuoteMarks(suggestion.value);
+    }
+
     const newValue = complete(suggestion);
     resetHighlight();
+
+    if (suggestion.type === 'labelName') {
```
manojVivek, line 275: Shall we make these literals as constants? That will help us define even stricter type for this field.

parca.yaml
```diff
@@ -13,4 +13,4 @@ scrape_configs:
     scrape_interval: "3s"
     scrape_timeout: "2s"
     static_configs:
-      - targets: [ '127.0.0.1:7070' ]
+      - targets: [ "127.0.0.1:7070" ]
```
metalmatze, line 16: This change seems unrelated.
I'd leave it as is, probably different editors will do different things and there will always be a back and forward. :grimacing:

metalmatze (commented):
This will be so good! Looking forward :star_struck:

brancz:
This is really awesome, nice work!

PR #762, opened 2022-03-15
Properly handle base images for multi-arch releases

### LINKED ISSUE #757: Images built with x86_64 base images for arm64 binaries
The basic functionality is actually there and works with current static binaries. However, this could a problem for the external dependencies that we pack with container images. 

CleanShot 2022-03-14 at 14 08 56

kakkoyun:
- Fix multi-arch base image problem
- ~This still keeps~ The builds scripts that uses podman/buildah for reproducible builds
- ~Goreleaser uses docker buildx to build images~ It's still not possible to use docker for reproducible builds. And Goreleaser only supports podman with pro version, for full build pipeline transparency sake we don't want to use pro version for Parca
- Pin and update base images
- Reorganize and reformat github/workflows

Fixes #757

### DIFF
--- a/.github/workflows/codeql-analysis-go.yml
+++ b/.github/workflows/codeql-analysis-go.yml
@@ -14,9 +14,17 @@ name: "CodeQL"
 on:
   push:
     branches: [ main ]
+    paths:
+      - '**.go'
+      - 'go.mod'
+      - 'go.sum'
   pull_request:
     # The branches below must be a subset of the branches above
     branches: [ main ]
+    paths:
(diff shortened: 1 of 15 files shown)


kakkoyun:
I'm still checking the reproducibility of the images using docker buildx, so please don't merge :D

brancz:
There seem to be quite a lot of changes and I can't make out whether all of them actually have to do with multi-arch things.

kakkoyun:
:+1: Happy to split the changes into several PRs

PR #767, opened 2022-03-15
debuginfo: Prevent redundant debug info uploads

### LINKED ISSUE #716: debuginfo: Prevent redundant uploads with a proper error
I think primarily I'd err on the side of not uploading if we don't absolutely have to because we would be uploading the same debuginfos thousands of times most likely.

__Originally posted by @brancz in https://github.com/parca-dev/parca/issues/705#issuecomment-1060318529__

kakkoyun:

Fixes #716

### DIFF
--- a/pkg/debuginfo/store.go
+++ b/pkg/debuginfo/store.go
@@ -138,19 +138,13 @@ func newCache(cacheCfg []byte) (*FilesystemCacheConfig, error) {
 }
 
 func (s *Store) Exists(ctx context.Context, req *debuginfopb.ExistsRequest) (*debuginfopb.ExistsResponse, error) {
-	err := validateID(req.BuildId)
-	if err != nil {
+	if err := validateID(req.BuildId); err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
-	found := false
-	err = s.bucket.Iter(ctx, req.BuildId, func(_ string) error {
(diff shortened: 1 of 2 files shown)


pkg/debuginfo/store.go
```diff
@@ -198,6 +200,20 @@ func validateID(id string) error {
 	return nil
 }
 
+func (s *Store) find(ctx context.Context, key string) (bool, error) {
+	found := false
+	err := s.bucket.Iter(ctx, key, func(_ string) error {
```
brancz, line 205: I'm happy for this to be a follow-up, but I expected that we also check the upstream debug info endpoint here.

brancz (approved):
Nice! (not sure why but I had assumed we already did this so great find!)

pkg/debuginfo/store.go
kakkoyun: To be crystal clear here, by the “the upstream debug info endpoint” you meant to look up the debuginfod servers that have been specified? 

Or do you want to utilize existing expose `Exists` API? 

For the first one, we check it at access time:

https://github.com/parca-dev/parca/blob/36b893ae5eb7adfe0ee28c34020b4d64fe123ced/pkg/debuginfo/store.go#L252-L260

brancz: I did indeed mean the first one, thanks for clarifying! Ignore my comment then!

PR #804, opened 2022-03-24
go.mod: Remove direct Thanos dependency

brancz:


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/parca-dev/parca
 go 1.16
 
 require (
+	github.com/OneOfOne/xxhash v1.2.6 // indirect
```
metalmatze, line 6: I think this one split into here as a wrong import somewhere?

metalmatze: Or actually did it split in the `github.com/thanos-io/objstore`?

brancz: huh, I honestly have no idea where it came from

PR #806, opened 2022-03-24
*: Fix address normalization

kakkoyun:
With yet to be release version of Parca Agent, we normalized addresses on the agent side. So no need to consider the mapping to normalize in the server-side for the profiles generated by the agent.

```
// NOTICE: We only store a single version of a mapping.
// Which means the m.Start actually correct for a single process.
// For a multi-process shared library, this will always be wrong.
// And storing the mapping for each process will be very expensive.
```


pkg/profile/profile_flat.go
```diff
@@ -250,7 +258,11 @@ func (pn *profileFlatNormalizer) mapMapping(ctx context.Context, src *profile.Ma
 		return mapInfo{}, err
 	}
 	if m != nil {
-		mi := mapInfo{m, int64(m.Start) - int64(src.Start)}
+		// NOTICE: We only store a single version of a mapping.
```
kakkoyun, line 261: @brancz I think this is the most important bit.

pkg/metastore/metastore_test.go
```diff
@@ -31,7 +31,7 @@ func TestMappingKeyBytes(t *testing.T) {
 		Offset: 2,
 	}
 
-	require.Equal(t, MakeMappingKey(m), []byte{
+	require.Equal(t, []byte{
```
kakkoyun, line 34: This has changed because expected should be first and actual the second.

kakkoyun:
I think actions are completely out-of-order right now.

brancz (approved):
I can't wait to see this work. 💯 🎉

PR #816, opened 2022-03-28
ui/grpc: Added provision to pass grpc metadata in @parca/components

manojVivek:


metalmatze:
Nice idea! Does that mean that the metadata will simply be be added if included?

manojVivek:
Yes that is right, Matthias. 

If there is a need to include some metadata to the Grpc-web requests then we can wrap the component or the app with the 'GrpcMetadataProvider' and pass the appropriate metadata object as the value.

ui/packages/shared/components/src/GrpcMetadataContext/index.tsx
```diff
@@ -0,0 +1,29 @@
+import {grpc} from '@improbable-eng/grpc-web';
+import React from 'react';
+
+const EMPTY_METADATA = new grpc.Metadata();
+const GrpcMetadataContext = React.createContext<grpc.Metadata>(EMPTY_METADATA);
```
monicawoj, line 5: maybe instead of using "React.createContext" and "React.useContext" below, we can just import the named "createContext" and "useContext"?

`import { createContext, useContext, ReactNode, ... } from 'react';

monicawoj (approved):
This looks good to me in general! But just so I'm on the same page - what exactly is the metadata we will be using? Why does the metadata need to be passed to @parca?

ui/packages/shared/components/src/GrpcMetadataContext/index.tsx
manojVivek: Fixed it.

manojVivek:
> what exactly is the metadata we will be using? Why does the metadata need to be passed to Parca?

These metadata can be mainly used to extend some functionality on the server side. 
Parca doesn't need it yet, but there is a need on the cloud app, so implementing it.

PR #826, opened 2022-03-28
pkg/parcaparquet: Store stacktrace UUIDs as []byte string

metalmatze:
We therefore need to concatenate all UUIDs as 16-bytes and cast those []byte as string to be stored in our columnar store. For reading back we read the string as []byte and then split it every 16 byte and use as individual UUIDs.

### DIFF
--- a/pkg/parcacol/parcacol.go
+++ b/pkg/parcacol/parcacol.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
-	"github.com/google/uuid"
 	"github.com/polarsignals/arcticdb/dynparquet"
 	"github.com/prometheus/prometheus/model/labels"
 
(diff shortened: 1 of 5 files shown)


pkg/parcacol/parcacol.go
```diff
@@ -97,10 +96,10 @@ func flatProfileToBuffer(logger log.Logger, ls labels.Labels, schema *dynparquet
 	return buf, nil
 }
 
-func extractLocationIDs(locs []*metastore.Location) []uuid.UUID {
-	res := make([]uuid.UUID, 0, len(locs))
+func extractLocationIDs(locs []*metastore.Location) []byte {
+	b := make([]byte, 0, len(locs)*16) // UUID are 16 bytes thus multiply by 16
 	for i := len(locs) - 1; i >= 0; i-- {
-		res = append(res, locs[i].ID)
+		b = append(b, locs[i].ID[:]...)
```
brancz, line 102: Does the go compiler optimize this to `copy(..., ...)` under the hood? If not we should probably do that here.

metalmatze: Oh, yes. Good point. Will update! :+1:

brancz:
Nice!!

PR #833, opened 2022-03-29
ui: Sort Targets by URL, make job name bigger

metalmatze:
The most important change in this PR is the fact that the targets for a job are now sorted by their URL. 
Sometimes, for debugging, I frequently reload the targets page, and not having the tables sorted makes it hard to follow what's going on.

Additionally, I've made the headline of each job bigger.

## Old
old
## New
new

cc @monicawoj


monicawoj (approved):
looks good to me!

PR #835, opened 2022-03-29
pkg/parcacol: Add 2 columns for pprofLabel and pprofNumLabel

metalmatze:
I've add two new columns for pprof labels.
The `TestConsistency` test works until the point where each row needs to be written and afaik panics trying to read values from a non existing dictionary: 
https://github.com/segmentio/parquet-go/blob/672347989f74307da3611cc2c2716c279704e946/dictionary.go#L103

### DIFF
--- a/pkg/parca/parca_test.go
+++ b/pkg/parca/parca_test.go
@@ -25,20 +25,23 @@ import (
 	"testing"
 	"time"
 
+	"github.com/apache/arrow/go/v7/arrow"
 	"github.com/apache/arrow/go/v7/arrow/memory"
 	"github.com/cenkalti/backoff/v4"
 	"github.com/fatih/semgroup"
 	"github.com/go-kit/log"
 	"github.com/google/pprof/profile"
+	"github.com/polarsignals/arcticdb"
+	columnstore "github.com/polarsignals/arcticdb"
+	"github.com/polarsignals/arcticdb/query"
+	"github.com/polarsignals/arcticdb/query/logicalplan"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/prometheus/model/labels"
-	"github.com/prometheus/prometheus/model/timestamp"
 	"github.com/prometheus/prometheus/promql/parser"
 	"github.com/stretchr/testify/require"
 	"go.opentelemetry.io/otel/trace"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
-	"google.golang.org/protobuf/types/known/timestamppb"
 
 	pb "github.com/parca-dev/parca/gen/proto/go/parca/profilestore/v1alpha1"
 	querypb "github.com/parca-dev/parca/gen/proto/go/parca/query/v1alpha1"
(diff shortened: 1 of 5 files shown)


pkg/parca/parca_test.go
```diff
@@ -295,46 +295,43 @@ func TestConsistency(t *testing.T) {
...
-	api := queryservice.NewColumnQueryAPI(
-		logger,
-		tracer,
-		m,
-		query.NewEngine(
-			memory.DefaultAllocator,
-			colDB.TableProvider(),
-		),
-		"stacktraces",
+	q := query.NewEngine(
```
brancz, line 308: Why replace the call that we expect a user to make with this lower-level call?

pkg/parcacol/parcacol.go
```diff
@@ -59,17 +60,41 @@ func FlatProfileToBuffer(logger log.Logger, ls labels.Labels, schema *dynparquet
...
+		}
+
+		pprofNumLabels := labels.Labels{}
+		for name, values := range s.NumLabel {
+			if len(values) > 1 {
+				panic("don't expect more than 1 value per pprof num label")
+			}
+			pprofNumLabels = append(pprofNumLabels, labels.Label{
+				Name:  name,
+				Value: strconv.FormatInt(values[0], 10),
```
brancz, line 81: We can have dynamic columns of type int, why not use that?

pkg/parcacol/schema.go
```diff
@@ -6,68 +6,83 @@ import (
...
+				Name:          columnPeriodUnit,
+				StorageLayout: parquet.Encoded(parquet.String(), &parquet.RLEDictionary),
+				Dynamic:       false,
+			}, {
+				Name:          columnPprofLabels,
+				StorageLayout: parquet.Encoded(parquet.Optional(parquet.String()), &parquet.RLEDictionary),
+				Dynamic:       true,
+			}, {
+				Name:          columnPprofNumLabels,
+				StorageLayout: parquet.Encoded(parquet.Optional(parquet.String()), &parquet.RLEDictionary),
```
brancz, line 55: Here we would use an `parquet.Int(64)` type

pkg/parcacol/schema.go
```diff
@@ -6,68 +6,83 @@ import (
...
+				StorageLayout: parquet.Int(64),
+				Dynamic:       false,
+			},
+		},
 		[]dynparquet.SortingColumn{
 			dynparquet.Ascending(columnSampleType),
 			dynparquet.Ascending(columnSampleUnit),
 			dynparquet.Ascending(columnPeriodType),
 			dynparquet.Ascending(columnPeriodUnit),
+			dynparquet.NullsFirst(dynparquet.Ascending(columnPprofLabels)),
```
brancz, line 84: I expected these to be after the stacktrace in their sorting. Potentially even after the timestamp (which is what was originally proposed in the RFC).

pkg/parca/parca_test.go
metalmatze: Good question. This might have been a rebase mistake!

pkg/parcacol/parcacol.go
metalmatze: Yes, now that I understand how the dynamic columns work, I agree. Was just easier, haha. Will update this!

brancz (commented):
Nice!!

brancz (approved):

PR #836, opened 2022-03-30
ui: Add rounded MatcherInput labels

metalmatze:
I'm not sure if these matcher input labels are so squared on purpose, but they somehow don't fit the rest of the UI all being rounded. 

The left one is how it currently looks and the right is my change. 
matchers


yomete (approved):
LGTM

PR #842, opened 2022-03-31
Validate labels for profiles received from agent

### LINKED ISSUE #817: Validate label names
It's currently possible to write invalid label names to the database that are then impossible to query using the query language. That's how https://github.com/parca-dev/parca-agent/pull/316 could sneak in without us noticing.

Sylfrena:
Fixes #817

### DIFF
--- a/pkg/profilestore/profilestore.go
+++ b/pkg/profilestore/profilestore.go
@@ -21,6 +21,7 @@ import (
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/google/pprof/profile"
+	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/labels"
 	"go.opentelemetry.io/otel/trace"
 	"google.golang.org/grpc/codes"
@@ -65,6 +66,10 @@ func (s *ProfileStore) WriteRaw(ctx context.Context, r *profilestorepb.WriteRawR
 	for _, series := range r.Series {
 		ls := make(labels.Labels, 0, len(series.Labels.Labels))
 		for _, l := range series.Labels.Labels {
+			if err := model.LabelName(l.Name).IsValid(); !err {
+				return nil, status.Errorf(codes.InvalidArgument, "invalid label name: %v", l.Name)
+			}
+
 			ls = append(ls, labels.Label{
 				Name:  l.Name,
 				Value: l.Value,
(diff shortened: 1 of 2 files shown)


Sylfrena:
@brancz, Should I have sanitized the labels instead?

pkg/profilestore/profilestore_test.go
```diff
@@ -0,0 +1,78 @@
...
+		Labels: []*profilestorepb.Label{{
+			Name:  "n1:n",
+			Value: "v1",
+		}},
+	}
+
+	samples := []*profilestorepb.RawSample{{RawProfile: buf.Bytes()}}
+
+	req := &profilestorepb.WriteRawRequest{
+		Tenant: "",
```
brancz, line 67: Slight nit: This line is not necessary as the empty string is the default value already. Also, this field is deprecated.

pkg/profilestore/profilestore_test.go
```diff
@@ -0,0 +1,78 @@
...
+		trace.NewNoopTracerProvider().Tracer(""),
+		db,
+		mStr,
+	)
+
+	buf := bytes.NewBuffer(nil)
+	ctx := context.Background()
+
+	// labelset with invalid label name
+	labelSet := profilestorepb.LabelSet{
```
brancz, line 57: A bit of a style choice but we usually inline these when we can, which is possible here.

pkg/profilestore/profilestore_test.go
```diff
@@ -0,0 +1,78 @@
...
+		Labels: []*profilestorepb.Label{{
+			Name:  "n1:n",
+			Value: "v1",
+		}},
+	}
+
+	samples := []*profilestorepb.RawSample{{RawProfile: buf.Bytes()}}
```
brancz, line 64: I think you should be able to skip this entirely, no? Validation of labels happens before samples are even looked at so we can just leave it completely empty.

brancz (commented):
Nice! Yes, this is pretty much exactly what I was thinking. Just a few style choices to be consistent with the rest of the codebase.

pkg/profilestore/profilestore.go
```diff
@@ -65,6 +66,10 @@ func (s *ProfileStore) WriteRaw(ctx context.Context, r *profilestorepb.WriteRawR
 	for _, series := range r.Series {
 		ls := make(labels.Labels, 0, len(series.Labels.Labels))
 		for _, l := range series.Labels.Labels {
+			if err := model.LabelName(l.Name).IsValid(); !err {
```
kakkoyun, line 69:
```suggestion
			if valid := model.LabelName(l.Name).IsValid(); !valid {
```

This is more idiomatic. `err` usually reserved for a type conforms `error` interface.

brancz (commented):
Nice! lgtm 🎉

PR #872, opened 2022-04-04
Add diffing to the Top Table

metalmatze:
First, we now expose the diff cumulative values via the API. For that we had to add a new field to the proto files and generate the Go and JS files from it. 
Second, we updated the top table generation on the backend side.
Third, we added a column in the UI to the top table and it automatically is sortable.

As the change in the UI was so simple I went from prototype straight to finishing it 😄 

## Preview
Screenshot 2022-04-04 at 19 04 46


brancz:
I don't really understand that error from the proto linter, so this lgtm 👍

metalmatze:
I'll take a look at it again. It might just be a version mismatch in the headers.

metalmatze:
Oh, seems like this is actually due to not being on the main branch. Not sure why this fails because of that. I don't think it should. Merging for now.

PR #847, opened 2022-04-03
Fix flaky 'TestAggregateTopByFunction`

importhuman:
This sorts the list by the meta location address, so the flaky `TestAggregateTopByFunction` passes every time. (#770)

I ran the test multiple times without any changes, and encountered 2 types of failures: `UniqueFunction` and `UniqueAddress`.

<details>
<summary>Logs for failing `UniqueFunction` test</summary>
<br>
Running tool: /usr/local/go/bin/go test -timeout 30s -run ^TestAggregateTopByFunction$ github.com/parca-dev/parca/pkg/query

--- FAIL: TestAggregateTopByFunction (0.00s)
    --- FAIL: TestAggregateTopByFunction/UniqueFunction (0.00s)
        /home/ujjwal/oss/parca-dev/parca/pkg/query/top_test.go:300: 
            	Error Trace:	top_test.go:300
            	Error:      	Not equal: 
            	            	expected: &queryv1alpha1.Top{state:impl.MessageState{NoUnkeyedLiterals:pragma.NoUnkeyedLiterals{}, DoNotCompare:pragma.DoNotCompare{}, DoNotCopy:pragma.DoNotCopy{}, atomicMessageInfo:(*impl.MessageInfo)(nil)}, sizeCache:0, unknownFields:[]uint8(nil), List:[]*queryv1alpha1.TopNode{(*queryv1alpha1.TopNode)(0xc00012a280), (*queryv1alpha1.TopNode)(0xc00012a2c0)}, Reported:2, Total:2, Unit:""}
            	            	actual  : &queryv1alpha1.Top{state:impl.MessageState{NoUnkeyedLiterals:pragma.NoUnkeyedLiterals{}, DoNotCompare:pragma.DoNotCompare{}, DoNotCopy:pragma.DoNotCopy{}, atomicMessageInfo:(*impl.MessageInfo)(nil)}, sizeCache:0, unknownFields:[]uint8(nil), List:[]*queryv1alpha1.TopNode{(*queryv1alpha1.TopNode)(0xc00012a240), (*queryv1alpha1.TopNode)(0xc00012a200)}, Reported:2, Total:2, Unit:""}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -50,5 +50,5 @@
            	            	      Id: ([]uint8) (len=16) {
            	            	-      00000000  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 02  |................|
            	            	-     },
[...]

PR #879, opened 2022-04-05
pkg/query: Sort list in aggregateTopByFunction

importhuman:
Follow-up PR after discussion in #847. Sorts by descending flat value by default, but if flat value is same, sorts by ascending meta location address (this second case is needed currently for the tests to pass every time).

### DIFF
--- a/pkg/query/top.go
+++ b/pkg/query/top.go
@@ -177,9 +177,15 @@ func aggregateTopByFunction(top *pb.Top) *pb.Top {
 		}
 	}
 
-	// Sort the list by meta location address, ascending
+	// Sort the list
 	sort.Slice(list, func(i, j int) bool {
-		return list[i].Meta.Location.Address < list[j].Meta.Location.Address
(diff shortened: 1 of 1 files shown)


pkg/query/top.go
```diff
@@ -177,9 +177,15 @@ func aggregateTopByFunction(top *pb.Top) *pb.Top {
 		}
 	}
 
-	// Sort the list by meta location address, ascending
+	// Sort the list
 	sort.Slice(list, func(i, j int) bool {
-		return list[i].Meta.Location.Address < list[j].Meta.Location.Address
+		// If flat value is equal, sort by address, ascending
+		if list[i].Flat == list[j].Flat {
+			return list[i].Meta.Location.Address < list[j].Meta.Location.Address
```
brancz, line 184: Nice catch!

brancz (approved):

brancz:
Thank you so much!

PR #880, opened 2022-04-05
pkg/query: Aggregate by dynamic pprof label columns for consistency

metalmatze:
This finally fixes the `TestConsistency`. As the pprof labels might differ for the same stack trace we need to aggregate by those pprof label columns too, to get a consistent and feature-parity pprof profile back.


brancz (approved):
Amazing!!

PR #883, opened 2022-04-05
proto:  Configure buf to use remote plugins

kakkoyun:
- Format proto files
- Add action step for format checking
- Upgrade buf (needed for format)
- Remove protoc
  - Rename actions
  - Clean up
  - Fix linter issue
- Use remote buf plugins
  - Remove protoc
  - Update local env scripts
  - Update github actions


kakkoyun:
@thorfour I was wondering if we still need `protoc` with `buf`? How about https://github.com/parca-dev/parca/blob/97f6a55b9abf9c463fee246b8917d652b598c436/proto/tools.go?

metalmatze:
> @thorfour I was wondering if we still need `protoc` with `buf`? How about https://github.com/parca-dev/parca/blob/97f6a55b9abf9c463fee246b8917d652b598c436/proto/tools.go?

I don't think we do. The docs seem to indirectly say that too:
https://docs.buf.build/how-to/replace-protoc-with-buf

kakkoyun:
@brancz @thorfour This is finally ready to review and get merged :tada:

brancz (approved):
Awesome!

PR #885, opened 2022-04-06
area/ui: Only show the diff column for the Top table when comparing

### LINKED ISSUE #837: Add Diff to the Top Table
It'd be nice to have a `Diff` column in the top table report when comparing two profiles.

yomete:
Resolves #837 

To be aware of when we are comparing profiles, I decided to make use of the global state management, which in turn led to creating @parca/store npm package so that UI components can easily access the store. 

**When comparing**
image

**Not comparing**
image


brancz:
Tiny addition, could we add a `+` in front of the number in the column if it's an increase (I assume `-` is already there in the case of an improvement since it should be a negative number).

yomete:
That makes sense, I can add that :)

Issue #612, opened 2022-02-09
symbolizer: Add Rust demangling support


kakkoyun:

javierhonduco:
Turns out we already supported the Rust mangling scheme. Added some basic tests in #892 for both C++ and Rust.
Let's keep an eye on the new mangling scheme that'll roll out in the future https://github.com/rust-lang/rust/issues/60705!

PR #892, opened 2022-04-07
demangler: add tests for the Cpp and Rust demanglers

javierhonduco:
Saw that @kakkoyun opened an issue to add Rust demangling support
(https://github.com/parca-dev/parca/issues/612) and was going to add another
library or implement it myself, but noticed that [the library we are currently
using for C++ demangling](https://github.com/ianlancetaylor/demangle) already supports the Rust mangling scheme!

Happy days!

This PR updates the comment in `demangle.go` and adds some unit tests for the
demangler to make sure we don't regress on basic functionality and set ups
the file for further testcases once we expand the languages we support.

Have decided to create independent test cases as I think it's more readable and
easier to understand, but happy to change it if you would rather use parametric
tests.

Please, let me know if you have any feedback!

Note: Something we probably want to bear in mind is that Rust's default mangling
scheme will change in the future: https://github.com/rust-lang/rust/issues/60705

**Test plan**

```
$ go test github.com/parca-dev/parca/pkg/symbol/demangle
ok      github.com/parca-dev/parca/pkg/symbol/demangle  0.004s
```

### DIFF
--- a/pkg/symbol/demangle/demangle.go
+++ b/pkg/symbol/demangle/demangle.go
@@ -47,8 +47,8 @@ func NewDemangler(mode string, force bool) *Demangler {
 	}
 }
 
-// Demangle updates the function names in a profile with demangled C++
-// names, simplified according to demanglerMode. If force is set,
+// Demangle updates the function names in a profile demangling C++ and
+// Rust names, simplified according to demanglerMode. If force is set,
 // overwrite any names that appear already demangled.
 // A modified version of pprof demangler.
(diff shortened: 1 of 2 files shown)


kakkoyun (approved):
LGTM :100:

kakkoyun:
> Note: Something we probably want to bear in mind is that Rust's default mangling
scheme will change in the future: https://github.com/https://github.com/rust-lang/rust/issues/60705

For this we can have another issue or we can update the existing one. Great catch :100:

PR #893, opened 2022-04-07
WIP run multiple podman jobs in parallel

javierhonduco:
**WIP WIP WIP** moving from draft state so all the checks run

The last builds are taking a very long time (>35'). Let's see if this helps.

Running all the `podman build` concurrently might be even better, but unsure if this is valid as there might be some shared state


javierhonduco:
Superseded by https://github.com/parca-dev/parca/pull/893

PR #918, opened 2022-04-11
Run `podman build` with as many workers as CPUs

javierhonduco:
To try to make the build a bit faster. Right now it's taking ~45 minutes.

Ideally we would run builds in parallel, too but it seems like support
for this was recently merged (https://github.com/containers/podman/pull/13404), so let's
wait a bit for it to be stable and then we can add it as well if.

### DIFF
--- a/scripts/make-containers.sh
+++ b/scripts/make-containers.sh
@@ -5,6 +5,7 @@ VERSION="$1"
 COMMIT="$2"
 MANIFEST="$3"
 ARCHS=('amd64' 'arm64' 'armv6' 'armv7' '386')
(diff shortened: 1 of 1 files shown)


javierhonduco:
It doesn't seem to improve timings in CI. Seems like these boxes should have 2 cores (https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) though but let me doublecheck, @kakkoyun!

javierhonduco:
Oops, removed my forked repo and understandably all outstanding parca PRs were closed. Sorry for the noise! Will re-open

PR #942, opened 2022-04-14
go.mod: Update arcticdb and remove replace directives

brancz:
With this change, Parca server actually becomes go-install/get-able 🎉


brancz:
Merging since same as https://github.com/polarsignals/arcticdb/pull/39 this is a no-op it just changes from our fork to mainline which is identical to the fork now.

kakkoyun (approved):
:rocket:

PR #956, opened 2022-04-19
pkg/query: Fix incorrect aggregation of flat values in table

brancz:
Previously we were only setting the flat value when we see the leaf for the first time, but it's possible that the same function occurs in many leafs, thus it's "cumulative flat value" is different. It also meant that whatever the first leaf value was that we saw was the one used. With this patch they are now correctly summed up.

@metalmatze


metalmatze:
How often do we need to fix this bug again and again (with every new storage)? LGTM! :+1:

PR #960, opened 2022-04-20
area/ui: Update fonts in the UI

### LINKED ISSUE #877: Use roboto mono in flamegraph and use softer color
Mock up 

Normal:

Screenshot 2022-04-05 at 12 17 08

Diff:

Screenshot 2022-04-05 at 12 18 19

yomete:
This PR updates the fonts used in the Parca UI and also uses softer colors for the Icicle graph. 

The Icicle graph and Top table now use the `Roboto Mono` font and I also decided to update the main font being used in the UI to be `Poppins` since that's the font used in the upcoming Parca website redesign.

Resolves #877 

Some screenshots:
image

image

image

image

image

and because it looks so gooood in **light mode**:
image

### DIFF
--- a/ui/packages/app/web/src/style/profile.css
+++ b/ui/packages/app/web/src/style/profile.css
@@ -1,3 +1,5 @@
+@import url('https://fonts.googleapis.com/css2?family=Roboto+Mono&display=swap');
+
 .profile-search-bar .queryExpressionInput {
   box-shadow: none;
   border-radius: 0;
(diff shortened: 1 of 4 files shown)


brancz:
Gorgeous 😍

brancz:
Perfectly happy with this being a follow up PR, but can we do the coloring differently for light dark mode? The problem with fading the coloring is that in dark mode the red actually ends up looking darker, which indicates it got “more worse”.

yomete:
I see what you mean @brancz. I can play around with a different (but close) shade of colors for the dark mode in this PR.

monicawoj (approved):
looks good to me!

yomete:
updated the colors now, thoughts @brancz 

image

image

brancz:
Love it!

brancz (approved):

PR #963, opened 2022-04-24
remove deduplicate set when distinct is enabled

yeya24:
Signed-off-by: Ben Ye <ben.ye@bytedance.com>

`seen` set is not needed because there is already a distinct operator in the logical plan.


brancz:
Awesome! Thank you!

PR #974, opened 2022-04-26
ui: Gracefully handle undefined range keys

brancz:
Whenever a new query is started, the range key of the query is
undefined, so we need to be able to handle that state.


yomete (approved):
lgtm!

manojVivek (approved):
LGTM!

PR #975, opened 2022-04-26
elfutils: Introduce fast Go binary predict helper function

### LINKED ISSUE #961: elfutils: Optimize IsSymbolizableGoObjFile
https://pprof.me/c7387bf/

As its current state, in order to be thoroughly determine if a binary is Go or not, we open and load the symbol section and check for specific symbols to determine if the binary is Go binary. Even though, this is a sound approach, it makes us to allocate large chunks of memory during search. We can tread-off soundness against efficiency here. At the end it's trying to catch and edge case, in case someone deliberately removes go sections the build ID section should be there.

https://github.com/parca-dev/parca/blob/cae3b001ca655793ddd23ced87d7e3a32dc6ca93/pkg/symbol/elfutils/elfutils.go#L77-L122

cc @metalmatze @javierhonduco

kakkoyun:

Fixes #961


metalmatze (approved):
Great! 
Seems the function isn't used in parca itself but is going to be used in the agent, right?

kakkoyun:
> Great! Seems the function isn't used in parca itself but is going to be used in the agent, right?

Yes, it will only be used in the agent for now.

PR #977, opened 2022-04-26
Rename additional class prop

### LINKED ISSUE #976: ui: rename "additionalClasses" prop to "className"
This occurs in the Button component, but it may also appear in other components. A prop named "additionalClasses" used for styling does not make much sense when there is no context on what it is "additional" to. This should be renamed to the more standard "className" prop.

monicawoj:
Resolves #976


manojVivek (approved):
LGTM

PR #990, opened 2022-04-28
Remove tsdb & introduce profile types

brancz:
Profile types introduce various things:

* Profile names can now be fully qualified, which means in addition to
  their name in a query, their sample type, sample unit, period type,
  period unit and whether it is a delta profile or not can be signified
  using optional made up the following way:

  `<name>:<sample-type>:<sample-type>:<period-type>:<period-unit>:delta`

* Profile types can be requested using the new ProfileTypes gRPC method,
  which is part of the Query gRPC service.

* Using profile types, the unit of the data displayed can be determined,
  meaning units are now working with the columnstore.

* To make all of this work, Parca needed to switch the typescript
  protobuf library to protobuf-ts, which in hindsight would have been
  better as a separate patch, but difficult to separate now.

With all of the above, the migration to the columnstore is finished.

Depends on https://github.com/polarsignals/arcticdb/pull/49


parca.yaml
```diff
@@ -11,6 +11,11 @@ debug_info:
...
   - job_name: "default"
     scrape_interval: "3s"
-    scrape_timeout: "2s"
     static_configs:
       - targets: [ '127.0.0.1:7070' ]
+    profiling_config:
+      pprof_config:
+        fgprof:
+          enabled: true
+          path: /debug/pprof/fgprof
```
metalmatze, line 20: We shouldn't add this to the default scrape config. Verrry few apps will have this.

metalmatze (commented):
Amazing to see so much old code gone! 😊 🎉

PR #991, opened 2022-04-28
parca.yaml: Add explanation of additional endpoints to example config

brancz:
Follow up to #990 

cc @metalmatze


metalmatze:
Nice, that's a really good idea. Still have it there but not too confusing for folks running with the default config :100:

PR #992, opened 2022-04-28
.gitattributes: Mark some paths as generated

javierhonduco:
So linguist doesn't tally them https://github.com/github/linguist/blob/master/docs/overrides.md

cc/ @brancz

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+gen/ linguist-generated
+ui/packages/shared/client/src/ linguist-generated


.gitattributes
```diff
@@ -0,0 +1,2 @@
+ui/packages/shared/client/src/parca/ linguist-generated
+ui/packages/shared/client/src/google/ linguist-generated
```
kakkoyun, line 2:
```suggestion
ui/packages/shared/client/src linguist-generated
gen linguist-generated
```
this also :point_up:

kakkoyun (commented):
:100: 

You can check https://github.com/parca-dev/parca/blob/main/buf.gen.yaml for generated code paths

PR #993, opened 2022-04-28
area/ui: Add legend to icicle graph when comparing profiles

### LINKED ISSUE #463: Improved diff icicle-graph coloring
1) It tends to be separately interesting if stacks are entirely new in icicle-graphs as opposed to having gotten worse/better currently denoted with green/red.
2) Measurements that are only very slightly different in contrast to the cumulative root diff, should probably still be colored in a variation of blue rather than green/red.
3) A legend for coloring would be great, as it's quite hard to grasp right now what the different colors mean.

yomete:
This PR adds a legend for the icicle graph when comparing two profiles so users understand what the colors mean. Additional textual information is shown when you hover on the legend.

Resolves #463 

image

image

image

image


brancz (approved):
Niiiice! lgtm 👍 🎉

ui/packages/shared/profile/src/components/DiffLegend.tsx
```diff
@@ -0,0 +1,109 @@
...
+}: {
+  onMouseEnter: () => void;
+  onMouseLeave: () => void;
+}) => {
+  const isDarkMode = useAppSelector(selectDarkMode);
+
+  const newSpanColor = isDarkMode ? '#B3BAE1' : '#929FEB';
+  const getIncreasedSpanColor = (transparency: number) => {
+    return isDarkMode
+      ? `rgba(255, 177, 204, ${transparency})`
```
manojVivek, line 20: Is it possible to share these color contants/logic from the `shared/functions` package file above?

manojVivek (commented):
Looks good except the minor feedback.

ui/packages/shared/profile/src/components/DiffLegend.tsx
```diff
@@ -0,0 +1,109 @@
...
+            leaveFrom="opacity-100 translate-y-0"
+            leaveTo="opacity-0 translate-y-1"
+          >
+            <Popover.Panel ref={setPopperElement} style={styles.popper} {...attributes.popper}>
+              <div className="overflow-hidden rounded-lg shadow-lg ring-1 ring-black ring-opacity-5">
+                <div className="p-4 bg-gray-50 dark:bg-gray-800">
+                  <div className="flex items-center justify-center"></div>
+                  <span className="block text-sm text-gray-500 dark:text-gray-50">
+                    This is a differential icicle graph, where the purple means unchanged, and the
+                    darker the red, the worse it got, and the darker the green, the better it got.
```
monicawoj, line 97: When we say "it", will the user know what "it" was that got worse or better?

metalmatze:
Looks amazing! 😎 

As for the wording I'm wondering if "better" and "worse" might be better suited to describe the difference since it's all relative?

ui/packages/shared/profile/src/components/DiffLegend.tsx
yomete: I reworded to this: `This is a differential icicle graph, where a purple-colored node means unchanged, and the darker the red, the worse the node got, and the darker the green, the better the node got.`

yomete:
pls have a look at this again when you can @manojVivek @monicawoj

monicawoj (approved):
lgtm!

PR #994, opened 2022-04-28
ui: Fix matcher for parca agent cpu profile type

brancz:


brancz:
Merging since the second commit is fixing CI which is failing on main.

PR #1011, opened 2022-05-03
pkg/parcacol: Create FlatProfileToRows to get back rows

metalmatze:
Instead of calling `parcacol.FlatProfileToBuffer` and getting back the buffer, I need to be able to `parcacol.FlatProfileToRows` and then only later create the buffer.


brancz:
lgtm

brancz (approved):

PR #1013, opened 2022-05-03
debuginfod: Fix tests depends on external service

kakkoyun:

The library used: ~https://github.com/sebdah/goldie~ https://github.com/dnaeon/go-vcr (goldie needs https://pkg.go.dev/net/http/httptest#ResponseRecorder which is useful when you test against your own server. Unfortunately, in this case we need to request to an external service. One alternative would be to create a proxy and use `ResponseRecorder `but I think it won't worth the hassle, so I used the go-vcr package directly.)

About golden test files: https://www.youtube.com/watch?v=yszygk1cpEc

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,13 @@ check-license:
 go/test:
 	 go test -v `go list ./...`
 
+VCR_FILES ?= $(shell find ./pkg/*/testdata -name "fixtures.yaml")
+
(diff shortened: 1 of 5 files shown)


brancz (approved):

brancz (approved):

brancz:
Sorry for double approving, I blame instable internet connection.

Issue #931, opened 2022-04-12
Enable specifying a prefix for pprof


roidelapluie:
While setting up Parca for prometheus' prombench, I had to configure all the profiling configs, because we use a prefix.

Since the prefix is common to all, it would be useful to be able to provide the prefix.

instead of:

```
    profiling_config:                                                      
      pprof_config:
       memory_total:        
               enabled: true                                               
               path: /10543/prometheus-pr/debug/pprof/allocs               
       block_total:         
               enabled: true                                               
               path: /10543/prometheus-pr/debug/pprof/block                
       goroutine_total:              
               enabled: true                                               
               path: /10543/prometheus-pr/debug/pprof/goroutine            
       mutex_total:                                                        
               enabled: true
               path: /10543/prometheus-pr/debug/pprof/mutex                
       process_cpu:                                                        
               enabled: true
               path: /10543/prometheus-pr/debug/pprof/profile
```

We could have:

```
    profiling_config:                                                      
      path_prefix: /10543/prometheus-pr
```

bonus point if it can be set via relabeling :)

importhuman:
This looks interesting, can I pick this up?

kakkoyun:
Of course, @importhuman go ahead, give it a shot!

importhuman:
Hey, had a couple of questions.

- Adding the `path_prefix` field seems simple enough, modifying the [ProfilingConfig](https://github.com/parca-dev/parca/blob/ba113635bf4708d3a730028d4c538a8ee1d03239/pkg/config/config.go#L171) struct. But how/where exactly do we look for the paths to add? As in, once the prefix is specified, where would the config paths be looked for?

- What is relabeling? I understand there is a Prometheus [package with the name](https://github.com/prometheus/prometheus/blob/main/model/relabel/relabel.go). In the context of this issue, would it be for replacing old paths with the new prefix?

roidelapluie:
Here is what I mean with relabeling:

```
static_configs:
- targets: [prometheus-pr, prometheus-release]
relabel_configs:
 - source_labels: [__address__]
   target_label: __pprof_prefix__
   replacement: /10543/$1/
 - target_label: __address__
   replacement: prombench.prometheus.io
```

brancz:
Did you mean the source label of the first relabel config to be `__prefix__`?

roidelapluie:
I originally wanted to scrape both the PR and RELEASE scrapes within one scrape job, which is challenging since they share the same domain.

[...]

PR #1022, opened 2022-05-08
pkg/config: Support prefix for pprof config (#931)

importhuman:
It seems I may have misunderstood the issue. 

Looking at the 2 configs in the issue description, I thought it meant discovering pprof configs (from somewhere) once the user has specified a particular prefix. However, from [this comment](https://github.com/parca-dev/parca/issues/931#issuecomment-1117506666) I realise it means not having to write a particular string repeatedly while specifying multiple configs. If it is the second case that needs to be addressed, this PR should be good.

Need help understanding relabelling though and what it means in the context of this issue.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -209,7 +209,7 @@ func (c *ScrapeConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	// If path prefix is specified, add to PprofConfig path
 	if unmarshalled.ProfilingConfig.PprofPrefix != "" {
 		for pt := range unmarshalled.ProfilingConfig.PprofConfig {
-			unmarshalled.ProfilingConfig.PprofConfig[pt].Path = unmarshalled.ProfilingConfig.PprofPrefix + unmarshalled.ProfilingConfig.PprofConfig[pt].Path
+			unmarshalled.ProfilingConfig.PprofConfig[pt].Path = filepath.Join(unmarshalled.ProfilingConfig.PprofPrefix, unmarshalled.ProfilingConfig.PprofConfig[pt].Path)
 		}
 	}
 
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -170,6 +170,7 @@ type ServiceDiscoveryConfig struct {
 
 type ProfilingConfig struct {
 	PprofConfig PprofConfig `yaml:"pprof_config,omitempty"`
+	PprofPrefix string      `yaml:"path_prefix,omitempty"`
 }
 
 type PprofConfig map[string]*PprofProfilingConfig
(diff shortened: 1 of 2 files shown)


brancz:
This implementation is pretty much exactly what I expected. What were your thoughts @metalmatze @roidelapluie @kakkoyun?

pkg/config/config_test.go
```diff
@@ -112,6 +112,84 @@ scrape_configs:
 	require.Equal(t, expected, c)
 }
 
+func TestLoadPrefixConfig(t *testing.T) {
+	prefixYAML := `
+scrape_configs:
+  - job_name: 'conprof'
```
kakkoyun, line 118:
```suggestion
  - job_name: 'parca'
```

pkg/config/config_test.go
```diff
@@ -112,6 +112,84 @@ scrape_configs:
...
+    scrape_interval: 10s
+    static_configs:
+      - targets: [ 'localhost:10902' ]
+    profiling_config:
+      path_prefix: /test/prefix
+      pprof_config:
+        memory:
+          enabled: true
+          path: /conprof/debug/pprof/allocs
```
kakkoyun, line 127:
```suggestion
          path: /parca/debug/pprof/allocs
```

pkg/config/config_test.go
```diff
@@ -112,6 +112,84 @@ scrape_configs:
...
+          enabled: true
+          path: /debug/fgprof
+  - job_name: 'empty-profiling-config'
+    profiling_config: {}
+`
+
+	expected := &Config{
+		ScrapeConfigs: []*ScrapeConfig{
+			{
+				JobName:        "conprof",
```
kakkoyun, line 138:
```suggestion
				JobName:        "parca",
```

pkg/config/config.go
```diff
@@ -205,6 +206,13 @@ func (c *ScrapeConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		}
 	}
 
+	// If path prefix is specified, add to PprofConfig path
+	if unmarshalled.ProfilingConfig.PprofPrefix != "" {
+		for pt := range unmarshalled.ProfilingConfig.PprofConfig {
+			unmarshalled.ProfilingConfig.PprofConfig[pt].Path = unmarshalled.ProfilingConfig.PprofPrefix + unmarshalled.ProfilingConfig.PprofConfig[pt].Path
```
roidelapluie, line 212: I think we should use `path.Join`

kakkoyun (commented):
LGTM 💯 

Just a couple of nits

importhuman:
I've updated the PR with the suggested changes, and also updated existing test to use `parca` instead of `conprof`.
I'll do some more research on relabelling before starting on it.

PR #1039, opened 2022-05-12
sign releases and main branch parca images with cosign

cpanato:
Sign the images that the Parca project generates with [cosign](https://github.com/sigstore/cosign)

- sign the releases
- sign when we build the images for the main branch and push both to ghcr and quay


kakkoyun (changes requested):
Looks awesome 💯 @cpanato Thank you very much for taking care for this 🌮🌮🌮

ghcr.io registry is the main one that we use. Could we have an action for that and use it in action? Or do we even need it?

.github/workflows/container.yml
```diff
@@ -52,6 +62,8 @@ jobs:
 
       - name: Push container
         if: ${{ github.event_name != 'pull_request' }}
+        env:
+          COSIGN_EXPERIMENTAL: true
         run: |
           make push-container
```
kakkoyun, line 68: We need a signed version of this, no?

.github/workflows/release.yml
```diff
@@ -110,6 +119,13 @@ jobs:
       - name: Login to registry
         run: |
           echo "${{ secrets.PERSONAL_ACCESS_TOKEN }}" | podman login -u parca-dev --password-stdin ghcr.io
+
       - name: Push container
         run: |
           make push-container
+
+      - name: Sign container
```
kakkoyun, line 127: Shouldn't this be before we actually push the image?

kakkoyun:
Shamelessly copying things for parca-agent https://github.com/parca-dev/parca-agent/pull/402

.github/workflows/container.yml
cpanato: errr 🤦 forgot that, added, ptal
thanks so mush for your review

kakkoyun (approved):
LGTM

cpanato:
@kakkoyun I'm checking why the job failed https://github.com/parca-dev/parca/actions/runs/2321776702

looks like when we use the `podman inspect` it is returning a different digest

kakkoyun:
@cpanato Hmmm 🤔 What would be the reason? Sorry, I'm a total noob on this stuff. You need to show me the ropes.

cpanato:
> @cpanato Hmmm 🤔 What would be the reason? Sorry, I'm a total noob on this stuff. You need to show me the ropes.

propose a fix here https://github.com/parca-dev/parca/pull/1041

PR #1057, opened 2022-05-17
Using default date range when partial query params are found in the URL

### LINKED ISSUE #1056: UI fails to load if start and end timestamps are not supplied
If you strip the `from_a` and `to_a` parameters from the query string, the UI fails to load.

I would exepect that, in the absence of these parameters, Parca would pick 'sensible' defaults such as from: now, to: now-3h.

image

This is unfortunate because I wanted to demo how to hop between Prometheus exemplars and Parca in my Kubecon [demo](https://prometheusdayeu22.sched.com/event/zfKp?iframe=no), but that is on me for leaving this so late :(

manojVivek:
Resolves #1056.


yomete (approved):
lgtm 👍🏾

Issue #562, opened 2022-01-18
Multi-arch testing support on CI


kakkoyun:
As suggested we need to figure out testing this with our CI. Dropping some links for future reference:
* https://github.blog/changelog/2019-12-03-github-actions-self-hosted-runners-on-arm-architectures/
* https://github.com/uraimo/run-on-arch-action

_Originally posted by @kakkoyun in https://github.com/parca-dev/parca/pull/559#pullrequestreview-855072364_

mrueg:
There are also some steps listed to verify it in this example: https://github.com/redhat-actions/buildah-build/blob/main/.github/workflows/multiarch.yml

kurious-diru:
Heyy

is this issue in progress? I saw few PRs from @mrueg but not sure

thanks!

mrueg:
@kurious-diru currently not working on anything here. Testing on an actual machine is still on my list.

knrt10:
If no one is, I would like to work on this.

kakkoyun:
@knrt10 Just go for, it would be awesome!

knrt10:
I just want to confirm some things. For which pipelines we need multi arch testing. 

Do we need to change the previous pipeline `.github/workflows/go-build-test.yml` to support multi arch test. Or create a separate ci pipeline for this?

kakkoyun:
It's ok to modify existing pipeline.

knrt10:
On various testing of trying to run on ARM machine(which takes a lot of time), I would like to share some of my findings

- [Github hosted runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners) does not have ARM arch yet and there is no plan for it. 
- https://github.com/uraimo/run-on-arch-action this executes command a non-x86 CPU architecture via QEMU. 
- This spwans a docker container, so you are inside a container of a already present docker container of Github actions.
- After which you have to install **Go, node, yarn** to it. (this takes some time).
- Network speed is slow due to which `yarn install` takes a lot of time and is quite flaky and unreliable. It gives the output most of the time.
```sh
nfo There appears to be trouble with your network connection. Retrying...

error An unexpected error occurred: "https://registry.yarnpkg.com/date-fns/-/date-fns-2.28.0.tgz: ESOCKETTIMEDOUT".
```

It would be easy if you folks have an ARM machine, you could easily host the github actions on it, and run self hosted github action. 

Any other suggestions are appreciated.

knrt10:
[...]

PR #1060, opened 2022-05-17
Multi arch testing support for CI

knrt10:
Added support for arm64 architecture. A new docker container with
`arm64v8/ubuntu` image is spawned using QEMU.
    
`yarn install` is slow on arm64 container in container mode. So
 installing yarn packages in amd64 ubuntu machine and mounting it while
 running docker container.

Fixes: #562 
    
 Signed-off-by: Kautilya Tripathi <tripathi.kautilya@gmail.com>

### DIFF
--- a/.github/workflows/go-build-test.yml
+++ b/.github/workflows/go-build-test.yml
@@ -15,30 +15,150 @@ on:
       - 'go.sum'
 
 jobs:
-  build-and-test:
+  build-and-test-multi-arch:
     runs-on: ubuntu-latest
+    name: Build on ${{ matrix.arch }}
+    # Run steps on a matrix of 2 arch.
+    strategy:
+      matrix:
+        arch:
+          - x64
+          # Names as per arch/ubuntu docker images.
+          - arm64v8
     steps:
-      - uses: actions/checkout@v3
+    - uses: actions/checkout@v3
+    - name: Set up Go
+      uses: actions/setup-go@v3
+      with:
+        go-version: '1.18'
+        check-latest: true
 
-      - name: Set up Go
-        uses: actions/setup-go@v3
-        with:
-          go-version: '1.18'
-          check-latest: true
+    - name: Build UI
+      if: ${{ matrix.arch == 'x64' }}
+      run: |
+        make ui/build
 
-      - name: Build UI
-        run: make ui/build
+    - name: Test
+      if: ${{ matrix.arch == 'x64' }}
+      run: |
+        make go/test
 
-      - name: Test
-        run: make go/test
+    - name: Build
+      if: ${{ matrix.arch == 'x64' }}
+      run: |
+        make build
 
-      - name: Build
-        run: make build
+    - name: Archive generatated artifacts
+      if: ${{ matrix.arch == 'x64' }}
+      uses: actions/upload-artifact@v3
+      with:
+        name: parca-bin
+        if-no-files-found: error
+        path: |
+          bin
 
-      - name: Archive generatated artifacts
-        uses: actions/upload-artifact@v3
-        with:
-          name: parca-bin
-          if-no-files-found: error
-          path: |
-            bin
+    - name: 'Run ${{ matrix.arch }}'
+      if: ${{ matrix.arch != 'x64' }}
+      run: |
+        # Install yarn packages initially on AMD machine so that it can be mounted as volume
+        # on platform specific docker container. This will save time as `yarn install` takes a lot of time
+        # on ARM container running using QEMU.
+        cd ui && yarn install && cd -
+        export RUNNER_ALLOW_RUNASROOT="1"
+
+        # Install QEMU and it's dependencies.
+        sudo apt-get install qemu binfmt-support qemu-user-static
+        docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
+
+        # Run platform specific based ubuntu image. Run it as a daemon in the background.
+        # Sleep the container for 1 day so that it keeps running until
+        # other steps are completed and the steps below can use the same container.
+        ARCH=${{ matrix.arch }}
+        # Ubuntu image name are different from arch image. So using this different variable.
+        IMAGE=${{ matrix.arch }}
+        if [ "$ARCH" == "arm64v8" ]; then
+          ARCH="arm64"
+        fi &&\
+        docker run --name ${{ matrix.arch}}_ubuntu -d --platform linux/$ARCH -v `pwd`:/parca $IMAGE/ubuntu /bin/bash -c \
+        'uname -m &&\
+        sleep 1d'
+
+    - name: 'Install packages on ${{ matrix.arch }}'
+      if: ${{ matrix.arch != 'x64' }}
+      run: |
+        # Install necessary packages on the ${{ matrix.arch}}_ubuntu container which will be used
+        # by below steps.
+        docker exec -t ${{ matrix.arch}}_ubuntu /bin/bash -c \
+        '# Update the container packages.
+        apt-get update -y -q &&\
+        apt-get upgrade -y -q &&\
+        apt-get install -y -q wget make git file build-essential'
+
+    - name: 'Setup Go on ${{ matrix.arch }}'
+      if: ${{ matrix.arch != 'x64' }}
+      run: |
+        export RUNNER_ALLOW_RUNASROOT="1"
+        # Install Golang, which will be used to build the code.
+        docker exec -t ${{ matrix.arch}}_ubuntu /bin/bash -c \
+        'ARCH=${{ matrix.arch }} && \
+        if [ "$ARCH" == "arm64v8" ]; then
+          ARCH="arm64"
+        fi &&\
+        wget https://dl.google.com/go/go1.18.2.linux-$ARCH.tar.gz &&\
+        tar -C /usr/local/ -xzf go1.18.2.linux-$ARCH.tar.gz &&\
+        export PATH=$PATH:/usr/local/go/bin && \
+        go version'
+
+    - name: 'Setup yarn on ${{ matrix.arch }}'
+      if: ${{ matrix.arch != 'x64' }}
+      run: |
+        export RUNNER_ALLOW_RUNASROOT="1"
+        docker exec -t ${{ matrix.arch}}_ubuntu /bin/bash -c \
+        '# Install LTS version of node.
+        ARCH=${{ matrix.arch }} &&\
+        if [ "$ARCH" == "arm64v8" ]; then
+          ARCH="arm64"
+        fi
+        cd /tmp && wget https://nodejs.org/dist/v16.15.0/node-v16.15.0-linux-$ARCH.tar.gz &&\
+        tar -xf node-v16.15.0-linux-$ARCH.tar.gz &&\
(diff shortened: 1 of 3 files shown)


knrt10:
Currently I don't know for sure whether the CI is slow or the arm machine is slow, but the `make ui/build` step takes lot of time. To be specific. `yarn install` is taking a lot of time. 

`make ui/build` passes but `make go/test` fails for some reason. Trying to figure it out and fix it.

.github/workflows/go-build-test.yml
```diff
@@ -15,25 +15,112 @@ on:
       - 'go.sum'
 
 jobs:
-  build-and-test:
+  build-and-test-amd64:
```
kakkoyun, line 18: Is there a reason to separate this out?

knrt10: Initially I though we have to run in a different job but it doesn't make sense. You are right, I think we can use `if: ${{ matrix.arch != 'x64' }}` and update strategy according it it. Will update it.

pkg/config/secret.go
```diff
@@ -19,6 +19,7 @@ package config
 // Secret special type for storing secrets.
 type Secret string
 
+// this is test
```
kakkoyun, line 22: I think we should remove this before merge.

kakkoyun (commented):
Looking better :tada:

pkg/debuginfo/debuginfod_test.go
```diff
@@ -72,47 +73,50 @@ func TestHTTPDebugInfodClient_request(t *testing.T) {
...
-				timeoutDuration: tt.fields.timeoutDuration,
-				client: &http.Client{
-					// Recorder transport passed, so the recorded data from testdata/fixtures will ber passed.
-					// Use make go/test-clean to remove the recorded data.
-					Transport: r,
-				},
-			}
-			ctx, cancel := context.WithTimeout(context.Background(), tt.fields.timeoutDuration)
-			t.Cleanup(cancel)
+		arch := runtime.GOARCH
```
kakkoyun, line 76: Everything should work with all the architecture, I think. Why do we have this?

pkg/config/secret.go
knrt10: yes, this is here just for testing. Once PR is approved, will remove it before merging.

pkg/debuginfo/debuginfod_test.go
knrt10: The want path here https://github.com/parca-dev/parca/pull/1060/commits/22106b0e35cdc9be5b7610b124ceef81bebe86b3#diff-6299d5e6bee95423b1ed778126b7433378b37d6a60daf47d5f332d5c2e33af1bR71 has ELF file in x86-64 format. So had to do it like this.

pkg/query/columnquery_test.go
```diff
@@ -499,74 +500,77 @@ func TestColumnQueryAPIQueryDiff(t *testing.T) {
...
-	res, err := api.ProfileTypes(ctx, &pb.ProfileTypesRequest{})
-	require.NoError(t, err)
 
-	require.True(t, proto.Equal(&pb.ProfileTypesResponse{Types: []*pb.ProfileType{
-		{Name: "memory", SampleType: "alloc_objects", SampleUnit: "count", PeriodType: "space", PeriodUnit: "bytes", Delta: true},
-		{Name: "memory", SampleType: "alloc_space", SampleUnit: "bytes", PeriodType: "space", PeriodUnit: "bytes", Delta: true},
-		{Name: "memory", SampleType: "inuse_objects", SampleUnit: "count", PeriodType: "space", PeriodUnit: "bytes", Delta: true},
-		{Name: "memory", SampleType: "inuse_space", SampleUnit: "bytes", PeriodType: "space", PeriodUnit: "bytes", Delta: true},
-	}}, res))
+		require.True(t, proto.Equal(&pb.ProfileTypesResponse{Types: []*pb.ProfileType{
```
knrt10, line 567: This one is quite falky. `res` returned in arm machine is not as the same order everytime. The ordering on `SampleType: "inuse_objects"` and `SampleType: "inuse_space"` is interchangeable and different every time the test is run. Figuring out a way to fix this.

knrt10:
Thank you for the review @kakkoyun. Addressed the changes, PTAL whenever you get time.

.github/workflows/go-build-test.yml
```diff
@@ -15,30 +15,102 @@ on:
...
+    - name: Archive generatated artifacts
+      if: ${{ matrix.arch == 'x64' }}
+      uses: actions/upload-artifact@v3
+      with:
+        name: parca-bin
+        if-no-files-found: error
+        path: |
+          bin
+
+    - name: 'Build and Test on ${{ matrix.arch }}'
```
kakkoyun, line 59: I think it would be more readable to split this into multiple steps.

.github/workflows/go-build-test.yml
```diff
@@ -15,30 +15,102 @@ on:
...
+        # installs yarn to build UI workspace.
+        docker run --rm -t -v `pwd`:/parca arm64v8/ubuntu /bin/bash -c \
+        "uname -m &&\
+        cd parca &&\
+        # Update the container packages.
+        apt-get update -y -q &&\
+        apt-get upgrade -y -q &&\
+        apt-get install -y -q wget make git file build-essential &&\
+
+        # Install Golang.
```
kakkoyun, line 84: So the setup-go action doesn't handle multi arch?

.github/workflows/go-build-test.yml
```diff
@@ -15,30 +15,102 @@ on:
...
+    - name: Set up Go
+      uses: actions/setup-go@v3
+      with:
+        go-version: '1.18'
+        check-latest: true
+
+    - name: Build UI
+      if: ${{ matrix.arch == 'x64' }}
+      run: |
+        make ui/build
```
kakkoyun, line 38: What do we do in these make actions that are specific to arch?

kakkoyun (commented):
Amazing work :+1: 

I'm trying to understand how we can make it more readable and avoid redundant work. So I've a couple of inlined questions.

.github/workflows/go-build-test.yml
knrt10: We have strategy defined [here](https://github.com/parca-dev/parca/pull/1060/files#diff-02549322dd3f50a460caa5562cd8842ab35f0f93ce890f824242bab7c338300bR22-R26). It takes list of architecture. Currently we are testing for `x86_64` and `aarch86`.

.github/workflows/go-build-test.yml
knrt10: So how it works is we setup Go, but it is done inside host container  i.e github runner container which is  `linux/amd64`. But to test `linux/arm64`. We create a new docker image `arm64v8/ubuntu` and install all the necessary packages. For more information please refer to https://github.com/parca-dev/parca/issues/562#issuecomment-1129723914 and https://github.com/parca-dev/parca/issues/562#issuecomment-1129839656. I hope it makes sense.

.github/workflows/go-build-test.yml
knrt10: That makes sense, different steps can be. 

- installing necessary packages and golang
- installing and setting up node and yarn
- Building ui
- Running test
- Running build

Do you agree?

kakkoyun: Sounds good) :+1:

pkg/query/columnquery_test.go
```diff
@@ -552,12 +552,12 @@ func TestColumnQueryAPITypes(t *testing.T) {
 	res, err := api.ProfileTypes(ctx, &pb.ProfileTypesRequest{})
 	require.NoError(t, err)
 
-	require.True(t, proto.Equal(&pb.ProfileTypesResponse{Types: []*pb.ProfileType{
+	require.Equal(t, proto.Size(&pb.ProfileTypesResponse{Types: []*pb.ProfileType{
```
kakkoyun, line 555: Why did we need to change this? It would be nice to add a comment about this.

.github/workflows/go-build-test.yml
```diff
@@ -15,30 +15,137 @@ on:
...
+    - name: Archive generatated artifacts
+      if: ${{ matrix.arch == 'x64' }}
+      uses: actions/upload-artifact@v3
+      with:
+        name: parca-bin
+        if-no-files-found: error
+        path: |
+          bin
+
+    - name: 'Run ${{ matrix.arch }}'
```
kakkoyun, line 59: Even though this templating indicates this task is parametric and will work for a different architecture, it's actually not. There is a lot of hardcoded arm-specific configuration below. 

It would be awesome to have a task that would work for future architecture as well. If it's too much work, let's indicate that this is just for arm64.

kakkoyun (commented):
Awesome work! :tada: It's already working!

Last nit-picks (bear with me) and we will merge it :rocket:

.github/workflows/go-build-test.yml
knrt10: Added support for different architecture. When adding new architecture just add the name in strategy array and it should work. I added a `s390x` job on the same pipeline to test. I think it should work

knrt10: Ok so new error popped up on s390x arch test but I guess currently `ARCHS=('amd64' 'arm64' 'armv6' 'armv7' '386')` are the supported architecures right? cc @kakkoyun

kakkoyun: Yes, I think so. You can check the list of binaries we release in the latest release https://github.com/parca-dev/parca/releases/tag/v0.11.1

knrt10: As mentioned [here](https://github.com/parca-dev/parca/pull/1060#discussion_r886419828), I have added support for multi arch. But the tests on 32bit architecure is failing. Please check logs [here](https://github.com/parca-dev/parca/runs/6686157753). So currently removed adding another arch test other than arm64.  I guess that should be the scope of another issue/PR. WDYT @kakkoyun?

kakkoyun: We shouldn't actually support 32-bit. We should even stop building binaries for 32-bit arch. Let's exclude those architectures and I'll send a PR to stop releasing those binaries.

knrt10: Cool, then PR is ready for review then. PTAL whenever free

kakkoyun: FYI https://github.com/parca-dev/parca/pull/1115

kakkoyun (approved):

kakkoyun (approved):

kakkoyun:
@knrt10 Thanks for the contribution. Shall I just merge it? Or do you want to clean things up? (you mentioned something like that)

knrt10:
Yes please wait, just have to remove one code from a file

knrt10:
Sure. Rebased and pushed

kakkoyun:
@knrt10 Amazing work. Thanks a lot for taking the time and making this happen. We need exactly the same facilities for https://github.com/parca-dev/parca-agent Looking forward to your contributions over there as well :hugs:

knrt10:
Sure, will do that

Issue #1070, opened 2022-05-23
[Docs] Storage Link broken


avestuk:
The link in the README to "Optimized Storage & Querying" is returning a 404. It's broken because of the trailing `/` although why this link breaks when the others are fine isn't something I could figure out.

avestuk:
Fixed by https://github.com/parca-dev/parca/pull/1071

PR #1071, opened 2022-05-23
Fix storage docs link

avestuk:
Fixes the broken link reported in https://github.com/parca-dev/parca/issues/1070

This is probably not the correct fix, however I'm not familiar with the docs engine you are using so I was not able to provide a fix there instead.


knrt10:
I think this should fix it.

brancz:
Thank you, looks good!

We’ll need you to sign the CLA (which is just a one time DCO so you don’t have to sign off each commit), to accept the contribution.

brancz:
We had another contributor say the same thing the other week, looks like we should have a look into another system for this. With the other contributor it eventually loaded, but it took about 5 minutes to render. My sincere apologies for the inconvenience.

avestuk:
No worries at all, just waiting worked a charm.

brancz:
Thank you so much for bearing with us!

manojVivek:
@all-contributors Please add @avestuk for docs.

PR #1079, opened 2022-05-25
smooth transition on chart resize

### LINKED ISSUE #1076: [ui] Make chart transition smooth on window resize
currently Parca charts adjust to the new window width on resize, but not smoothly

Smoothing the transition can be accomplished by altering the functionality of the 'CalcWidth' component or creating a "useContainerDimensions" hook. Given our more extensive use of hooks throughout the application (vs. HOCs), the hook seems like the preferable approach.

monicawoj:
Resolves #1076


brancz:
I don't really understand how this happened but the `yarn.lock` appears to be out of date.

monicawoj:
> I don't really understand how this happened but the `yarn.lock` appears to be out of date.

Yeah, strange... I noticed, there are a bunch of linter errors in files that were not touched by this PR as well, so I'll take a look at addressing those as well

brancz:
My best guess is that this has something to do with dependabot.

PR #1087, opened 2022-05-30
pkg/parcacol: Create a pprof ingester that converts to row directly

metalmatze:
WriteRaw requests have been the bottleneck in the last weeks as profiles have shown.
Since the migration to columnar storage we don't need the FlatProfiles anymore and now convert pprof profiles into columnar rows directly. 

```
name          old time/op    new time/op    delta
_WriteRaw-24     1.03s ± 2%     0.95s ± 3%   ~     (p=0.100 n=3+3)

name          old alloc/op   new alloc/op   delta
_WriteRaw-24    3.28GB ± 2%    2.97GB ± 4%   ~     (p=0.100 n=3+3)

name          old allocs/op  new allocs/op  delta
_WriteRaw-24     3.56M ± 0%     3.47M ± 0%   ~     (p=0.100 n=3+3)
```

Two things are missing for this to be merged:
- [x] Rebase for newer parquet-go changes
- [x] Add the pprof labels back to the StacktraceKey

### DIFF
--- a/pkg/parca/parca.go
+++ b/pkg/parca/parca.go
@@ -64,8 +64,9 @@ import (
 )
 
 const (
-	symbolizationInterval = 10 * time.Second
-	flagModeScraperOnly   = "scraper-only"
+	symbolizationInterval   = 10 * time.Second
+	flagModeScraperOnly     = "scraper-only"
+	metaStoreBadgerInMemory = "badgerinmemory"
 )
 
 type Flags struct {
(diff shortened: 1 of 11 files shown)


brancz:
Any reason not to delete now unused things from `pkg/profile`?

brancz:
This looks really good and happy to merge and do cleanups as follow ups as well!

metalmatze:
Yes, I wanted to keep this PR simple and in a way that it's easily reviewable. Will clean up in a follow-up PR which will probably move a few files around too.

brancz:
Sounds good to me!

PR #1109, opened 2022-05-31
Update Makefile go/test to run benchmarks once

metalmatze:
This will run every benchmark with just a single iteration to make sure the benchmarks aren't broken.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,8 +62,9 @@ check-license:
 
 .PHONY: go/test
 go/test:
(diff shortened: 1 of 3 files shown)


metalmatze:
Sweet! :relaxed: 
One Benchmark was still broken and after updating it seems to work running the benchmarks with `benchtime=1x` :tada:

brancz (approved):
lgtm, nice!

PR #1116, opened 2022-06-01
area/ui: Add a search input for easy filtering of nodes

### LINKED ISSUE #482: Add search for spans in the UI
We'd love to be able to search for specific functions by name and then have them highlighted if they match.
Right now, I often use the browser's search but if the function's name isn't fully readable I won't even see which spans are matching.

yomete:
This PR adds a search input field for easily identifying nodes in the icicle graph and top table report.

I also refactored the `FlamegraphTooltip` component to be its own standalone component so as to reduce the LOC in the `ui/packages/shared/profile/src/IcicleGraph.tsx` file.

image
image

Resolves #482


brancz:
At first sight, the colors made me think of a diff. Maybe we can instead fade out the non-matches?

yomete:
image
image

brancz:
Love it!

kakkoyun:
We need to address this CI failure. Something is not right with the main. Any ideas?

knrt10:
As mentioned [here](https://github.com/multiarch/qemu-user-static/issues/59#issuecomment-885634148) adding `qemu binfmt-support` in `Install qemu dependency` step might solve the issue. I have to try it locally tough to confirm

ui/packages/shared/functions/src/index.ts
```diff
@@ -255,3 +257,11 @@ export const diffColor = (diff: number, cumulative: number, isDarkMode: boolean)
 
   return color;
 };
+
+export const isSearchMatch = (currentSearchString: string | undefined, name: string) => {
+  return (
+    currentSearchString !== undefined &&
+    currentSearchString !== '' &&
```
monicawoj, line 264: can we simplify these two statements with Boolean(currentSearchString)

ui/packages/shared/profile/src/IcicleGraph.tsx
```diff
@@ -102,6 +94,12 @@ function IcicleRect({
         width={width - 1}
         height={height - 1}
         style={{
+          opacity:
+            currentSearchString !== undefined &&
+            currentSearchString !== '' &&
```
monicawoj, line 99: same thing here https://github.com/parca-dev/parca/pull/1116/files#r887756385

metalmatze:
Love the update with the faded colors. :100: 
Since you only posted dark mode screenshots, I want to make sure you potentially don't forget to update the light theme too. 
Looking forward to this! :rocket:

yomete:
@metalmatze yep, the light theme was also updated.

image

image

knrt10:
@kakkoyun regarding https://github.com/parca-dev/parca/pull/1116#issuecomment-1144542095 adding `qemu binfmt-support` only removes the errror `Unknown host QEMU_IFLA type: 54`. To fix `info There appears to be trouble with your network connection. Retrying...` , which is a known error for installing yarn packages in container, one has to increase the network-timeout because default timeout is 30sec I think.

So in [dockerfile](https://github.com/parca-dev/parca/blob/main/Dockerfile#L12) adding `--network-timeout 100000` fixes the issue. I have tried it locally it does not fail at step 2 while fetching packages. 

For more information: please [check](https://github.com/yarnpkg/yarn/issues/4890#issuecomment-358179301)

Creating a PR for the same

PR #1118, opened 2022-06-02
actions: Fix build and push job for arm64

knrt10:
To run arm64 arch, QEMU is used. `qemu binfmt-support` packages are
required qemu dependency.

While building container images for arm64 the network is slow due to
container in container mode and yarn install fails if the packages are
big as default retry timeourt is 30sec. Increasing it fixes the issue.

Signed-off-by: Kautilya Tripathi <tripathi.kautilya@gmail.com>

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -41,8 +41,8 @@ jobs:
       # To build multi-arch images, see also: https://github.com/redhat-actions/buildah-build#multi-arch-builds
       - name: Install qemu dependency
         run: |
-          sudo apt-get update
-          sudo apt-get install -y qemu-user-static
+          sudo apt-get update -y
+          sudo apt-get install -y binfmt-support qemu-user-static
 
       - name: Build container
         run: make container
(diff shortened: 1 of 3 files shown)


kakkoyun (approved):
LGTM :100:

knrt10:
While browsing through previous passed action runs I found that even though `arm64` build for images was failing and `exiting with code 1`, the job still passed. I guess it was because in [make-containers.sh](https://github.com/parca-dev/parca/blob/main/scripts/make-containers.sh#L2) there is no `-e` flag, which prevents it from exiting even though the it has has a non-zero exit status. I have updated this in a different commit.

cc @kakkoyun 

For reference please [check](https://github.com/parca-dev/parca/runs/6673787925?check_suite_focus=true) and open `Build container` step and search for `Exit code: 1`

kakkoyun:
> While browsing through previous passed action runs I found that even though `arm64` build for images was failing and `exiting with code 1`, the job still passed. I guess it was because in [make-containers.sh](https://github.com/parca-dev/parca/blob/main/scripts/make-containers.sh#L2) there is no `-e` flag, which prevents it from exiting even though the it has has a non-zero exit status. I have updated this in a different commit.
> 
> cc @kakkoyun
> 
> For reference please [check](https://github.com/parca-dev/parca/runs/6673787925?check_suite_focus=true) and open `Build container` step and search for `Exit code: 1`

This is a known issue, we would like to fix. It happens in parca-agent as well. Something we've overlooked. Feel free to fix it.
Otherwise, it's on my to-do list.

Issue #1062, opened 2022-05-19
SIGKILL after a few seconds when CPU doesn't support AVX instructions


jorgelbg:
I've been running a small parca instance (it is awesome 🎉). Today I updated to the latest version (0.11.1):

```
$ ./parca --version
parca, version 0.11.1 (commit: fcbc4371da7ec149d58b7cd1e39c782a1bccd595)
```

After the update parca starts successfully but it dies after a few seconds with:

```
SIGILL: illegal instruction
PC=0x9bb34f m=11 sigcode=2
instruction bytes: 0xc4 0xe2 0x7d 0x58 0x4c 0x24 0x20 0x48 0xc1 0xe9 0x20 0x48 0xf 0xaf 0xcb 0x48

goroutine 141 [running]:
github.com/segmentio/parquet-go/bloom.filterInsert({0xc007711e40, 0x1, 0x1}, 0x418fb529e8aa5d)
	github.com/segmentio/parquet-go@v0.0.0-20220421002521-93f8e5ed3407/bloom/filter_amd64.s:110 +0xf fp=0xc009845400 sp=0xc0098453f8 pc=0x9bb34f
github.com/segmentio/parquet-go/bloom.SplitBlockFilter.Insert(...)
	github.com/segmentio/parquet-go@v0.0.0-20220421002521-93f8e5ed3407/bloom/filter.go:62
github.com/segmentio/parquet-go/bloom.(*SplitBlockFilter).Insert(0x203000?, 0xc007ff8600?)
	<autogenerated>:1 +0x48 fp=0xc009845430 sp=0xc009845400 pc=0x9bb628
github.com/segmentio/parquet-go.(*bloomFilterEncoder).insert(0xc00108a480, {0xc007ff8600?, 0x7f164bbe1c80?, 0x42?})
	github.com/segmentio/parquet-go@v0.0.0-20220421002521-93f8e5ed3407/bloom.go:200 +0x47 fp=0xc009845460 sp=0xc009845430 pc=0xab5967
github.com/segmentio/parquet-go.(*bloomFilterEncoder).EncodeByteArray.func1({0xc007ff8600?, 0x480?, 0x18?})
	github.com/segmentio/parquet-go@v0.0.0-20220421002521-93f8e5ed3407/bloom.go:183 +0x2e fp=0xc009845490 sp=0xc009845460 pc=0xab572e
[...]

PR #1127, opened 2022-06-06
Gate avx2 instructions support for parquet-go

javierhonduco:
Some CPUs / hypervisors don't have avx2 support. This PR gates using
this feature in the parquet-go library during compilation depending
on whether the CPU has this feature

Thanks, @achille-roussel for the hint! (https://github.com/segmentio/parquet-go/issues/189#issuecomment-1147093453) 😄 

This is a remediation for: https://github.com/parca-dev/parca/issues/1062

## Test plan 

- With no avx2 support
```shell
$ make 
[...]
go build -tags purego -o bin/ ./cmd/parca
```

Ran the binary for several minutes with parca-agent sending profiles without any issues

- With avx2 support
```shell
$ make 
[...]
go build  -o bin/ ./cmd/parca
```

@jorgelbg could you please confirm if this fixes the issue for you as well? Thanks!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,10 @@ else
 endif
 VERSION ?= $(if $(RELEASE_TAG),$(RELEASE_TAG),$(shell $(CMD_GIT) describe --tags 2>/dev/null || echo '$(BRANCH)$(COMMIT)'))
 OUT_DOCKER ?= ghcr.io/parca-dev/parca
+HAVE_AVX2 := $(shell grep avx2 /proc/cpuinfo)
+ifneq ($(.SHELLSTATUS),0)
+	GO_BUILD_TAGS := -tags purego
+endif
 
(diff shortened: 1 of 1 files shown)


kakkoyun (approved):
LGTM :rocket:

brancz:
I don't think we should do this. This means that all the binaries we publish, which is what most people use, will not take advantage of the vectorized instructions even if they have them. Most users do not build from source. I think we should fix this in parquet-go.

achille-roussel:
I agree with @brancz, I'm sorry if my suggestion led to confusion here. When I brought it up I imagined it might be useful to unblock the particular build that was being blocked on the crash.

brancz:
No worries, I imagined it was something like that.

javierhonduco:
Makes sense, closing this PR

PR #1132, opened 2022-06-07
Callgraph visualization

monicawoj:
This PR introduces a new visualization to the profiler - the callgraph. The callgraph is interactive, with a tooltip displaying information of the node appearing on node hover. 

This PR also introduces a slight expansion of the `FlamegraphTooltip` component which will allow the tooltip to be displayed in a 'fixed' position (such as always above or below a given chart) if we decide that it would make sense to have this is a single location that doesn't cover the entire graph <-- which could be useful for screenshots for example


monicawoj:
Note - the most difficult piece when trying to use the JSON output from Graphviz was to parse the "pos" (position) data of the edges. This is not in normal SVG path format, so some manipulation was necessary to properly render the start, end, and control points defining the curve.

yomete (approved):
Really awesome work with the Callgraph implementation 👏🏾  Learnt a whole lot about D3 as I was reviewing :)

syepes:
+1 Great feature

monicawoj:
Still need to address self-loops (method that calls itself / edge that has the same start and end node)

manojVivek:
Should we try creating a feature flag for `callgraph`?  So that we can merge the changes and do improvements on top of it, instead of waiting with this huge PR? 

Maybe we can keep the option `disabled` by default and a simple query param in the URL can enable it?

monicawoj:
> Should we try creating a feature flag for `callgraph`? So that we can merge the changes and do stability improvements incrementally, instead of waiting with this huge PR?
> 
> Maybe we can keep the option `disabled` by default and a simple query param in the URL can enable it?

Yeah I'm a fan of that idea!

yomete (approved):

ui/packages/shared/functions/src/useUIFeatureFlag/index.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+  if (disableFlag) {
+    delete flags[disableFlag];
+  }
+  window.localStorage.setItem(UI_FLAGS, JSON.stringify(flags));
+};
+
+initializeFlagsFromURL();
+
+const useUIFeatureFlag = (featureFlag: string, defaultValue: boolean = false) => {
```
monicawoj, line 41: could we rename 'featureFlag' to something that indicates that this is the feature name? For example, featureKey or featureName? When I see "featureFlag" I think of the entire name-value pair (the key along with the boolean flag value). Maybe it's just me though 🤔  let me know what you think!

monicawoj (commented):
@manojVivek thanks for adding the feature flag commit - very cool solution! Just one thought, could we perhaps add a comment that communicates what you shared in the Discord channel regarding the use of 'enable-ui-flag' and 'disable-ui-flag'? I think it would be helpful for anyone else making use of this in the future. 

Something that communicates the below:
"For easy modification of the flags, I have added two query params too, enable-ui-flag and disable-ui-flag.

Visiting this URL would enable the flag for that user browser: http://localhost:3000/?enable-ui-flag=callgraph
And this would disable the same: http://localhost:3000/?disable-ui-flag=callgraph"

Additionally, may be good to know that these 'enable' and 'disable' params only work for setting one flag value at a time (rather than for example enabling "callgraph" and another feature at the same time).

manojVivek:
> @manojVivek thanks for adding the feature flag commit - very cool solution! Just one thought, could we perhaps add a comment that communicates what you shared in the Discord channel regarding the use of 'enable-ui-flag' and 'disable-ui-flag'? I think it would be helpful for anyone else making use of this in the future.
> 

Done, just posted a message with a link to the doc and the code.

Issue #1097, opened 2022-05-30
symbolizer: Reject invalid dwarf files


javierhonduco:
Related issue on the parca-agent side: https://github.com/parca-dev/parca-agent/issues/420

kakkoyun:
ref: 77b9181cac2efd241edba41890699b99231eeccf

PR #1136, opened 2022-06-08
Add Debuginfo metadata

javierhonduco:
At the moment, debug files are uploaded to our gRPC server by [streaming](https://github.com/parca-dev/parca/blob/2af5503bba08b709eaf07a3b83b18c9bc29cd262/pkg/debuginfo/reader.go#L34) its chunks. [Once we receive all the chunks](https://github.com/parca-dev/parca/blob/e0c7d0cf16275d30591a93a373c292683efab360/pkg/debuginfo/client.go#L54), we write the whole file in blob storage.

Right now there's no metadata associated with said debug files. Adding metadata to the uploads would allow us, among other things, to:
- debug issues more easily, as we can add extra data to uploads
- add new features in the future (such as different debug formats), and many more

Another welcomed effect is that, while this doesn't eliminate Upload() race conditions, it heavily reduces their likelihood. 

Imagine two concurrent uploads for the same debug info. While this might seem unlikely, when profiling several hosts, or profiling all the processes in a machine (this is how this issue surfaced) this is something that happens often.

```
(time) -----------------------------------------> (requests)

|    Upload()
|    check if file exists in blob storage = false  Upload()
|    read all chunks [slow critical section]       check if file exists in blob storage = false            
|    write the file in blob storage                read all chunks [slow critical section]
|                                                        write the file in blob storage (again!)
|
v

```

The reason for the diminished race window is that the critical section becomes smaller. Before this PR, the critical section is relatively slow as it has to read all the chunks and read a file that could be hundreds of MBs to blob storage.

With this PR, the critical section is smaller, as we just need to read a small file and parse the JSON, which is significantly faster.

In order to make sure that previously running parca-agent instances that might have already uploaded debug info files, we don't error if we don't find the metadata file, and just log this event. In another PR, if we think it's necessary we can 'backfill' the metadata for previously uploaded debug files.

## Test plan

Wrote some basic tests:
```shell
[javierhonduco@localhost-live parca]$ go test -v github.com/parca-dev/parca/pkg/debuginfo
=== RUN   TestHTTPDebugInfodClient_request
=== RUN   TestHTTPDebugInfodClient_request/success
=== CONT  TestHTTPDebugInfodClient_request
    debuginfod_test.go:40: <nil>
--- PASS: TestHTTPDebugInfodClient_request (0.35s)
    --- PASS: TestHTTPDebugInfodClient_request/success (0.01s)
=== RUN   TestMetadata
--- PASS: TestMetadata (0.00s)
=== RUN   TestStore
--- PASS: TestStore (0.00s)
PASS
ok  	github.com/parca-dev/parca/pkg/debuginfo	0.371s
```

And ran `parca` and `parca-agent` side to side with no issues for several hours, without issues

With the following structure created:
```shell
[javierhonduco@localhost-live parca]$ tree tmp/
tmp/
├── 618d3581ff908f8b3bd57c8ba53324903a6c24e4
│   ├── debuginfo
│   └── metadata
├── b23b4013416032e081da00934207e154233bcda7
│   ├── debuginfo
│   └── metadata
├── c8e7be08e00f53b66489178db08f7c49e0b123c0
│   ├── debuginfo
│   └── metadata
└── cf014e05d88e8b0168baf3b6d51ae070ff9fdd7a
    ├── debuginfo
    └── metadata
[...]
```

And the right contents in the file:
```shell
[javierhonduco@localhost-live parca]$ cat tmp/618d3581ff908f8b3bd57c8ba53324903a6c24e4/metadata
{
        "state": "METADATA_STATE_UPLOADED",
        "started_upload_at": 1654686039,
        "finished_upload_at": 1654686039
}
```

_Note_: This is **not** a fix for https://github.com/parca-dev/parca/issues/1097, but it's something that was developed while debugging it

### DIFF
--- a/pkg/debuginfo/metadata.go
+++ b/pkg/debuginfo/metadata.go
@@ -27,15 +27,15 @@ import (
 )
 
 var (
-	ErrDebugInfoMetadataShouldExist            = errors.New("debug info metadata should exist")
-	ErrDebugInfoMetadataExpectedStateUploading = errors.New("debug info metadata state should be uploading")
+	ErrMetadataShouldExist            = errors.New("debug info metadata should exist")
+	ErrMetadataExpectedStateUploading = errors.New("debug info metadata state should be uploading")
+	ErrMetadataUnknownState           = errors.New("debug info metadata state is unknown")
 )
 
 type metadataState int64
 
 const (
-	// There was an unexpected error. The error will be filled in the return value.
-	metadataStateError metadataState = iota
+	metadataStateUnknown metadataState = iota
 	// There's no debug info metadata. This could mean that an older version
 	// uploaded the debug info files, but there's no record of the metadata, yet.
 	metadataStateEmpty
@@ -46,14 +46,14 @@ const (
 )
 
 var mdStateStr = map[metadataState]string{
-	metadataStateError:     "METADATA_STATE_ERROR",
+	metadataStateUnknown:   "METADATA_STATE_UNKNOWN",
 	metadataStateEmpty:     "METADATA_STATE_EMPTY",
 	metadataStateUploading: "METADATA_STATE_UPLOADING",
 	metadataStateUploaded:  "METADATA_STATE_UPLOADED",
 }
 
 var strMdState = map[string]metadataState{
-	"METADATA_STATE_ERROR":     metadataStateError,
+	"METADATA_STATE_UNKNOWN":   metadataStateUnknown,
 	"METADATA_STATE_EMPTY":     metadataStateEmpty,
 	"METADATA_STATE_UPLOADING": metadataStateUploading,
 	"METADATA_STATE_UPLOADED":  metadataStateUploaded,
@@ -101,7 +101,7 @@ type metadata struct {
 }
 
 func (m *metadataManager) update(ctx context.Context, buildID string, state metadataState) error {
-	level.Debug(m.logger).Log("msg", "Attempting state update to", "state", state)
+	level.Debug(m.logger).Log("msg", "attempting state update to", "state", state)
 
 	switch state {
 	case metadataStateUploading:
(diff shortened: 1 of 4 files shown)


javierhonduco:
Lint fix

pkg/debuginfo/metadata.go
```diff
@@ -0,0 +1,165 @@
...
+	// There's no debug info metadata. This could mean that an older Parca version
+	// uploaded the debug info files, but there's not record of the metadata, yet.
+	metadataStateEmpty
+	// The debug info file is being uploaded.
+	metadataStateUploading
+	// The debug info file is fully uploaded.
+	metadataStateUploaded
+)
+
+func setMetadataLogger(l log.Logger) {
```
kakkoyun, line 50: Let's create a `MetaDataManager` to manage the state rather than using package-level variables. Which can store the `logger` and the `bucket`.
We can initialize an instance per debuginfo.Store.

kakkoyun (changes requested):
Great initiative :+1:  This will pave the ground for a lot of features. 

I have commented inline for possible changes.

kakkoyun (commented):
I have modified the PR a little bit.

- Refactored and added MetadataManager
- Made the MetadataState human-readable when marshalled

pkg/debuginfo/metadata.go
```diff
@@ -0,0 +1,195 @@
...
+	// uploaded the debug info files, but there's no record of the metadata, yet.
+	metadataStateEmpty
+	// The debug info file is being uploaded.
+	metadataStateUploading
+	// The debug info file is fully uploaded.
+	metadataStateUploaded
+)
+
+var mdStateStr = map[metadataState]string{
+	metadataStateError:     "METADATA_STATE_ERROR",
```
metalmatze, line 49: Do we even ever write an errored debuginfo and metadata to object storage?

metalmatze (commented):
Some question in regards to how this even works when writing to object storage.

pkg/debuginfo/metadata.go
```diff
@@ -0,0 +1,195 @@
...
+var (
+	ErrDebugInfoMetadataShouldExist            = errors.New("debug info metadata should exist")
+	ErrDebugInfoMetadataExpectedStateUploading = errors.New("debug info metadata state should be uploading")
+)
+
+type metadataState int64
+
+const (
+	// There was an unexpected error. The error will be filled in the return value.
+	metadataStateError metadataState = iota
```
brancz, line 38: The `0` value shouldn't be overloaded with one that has semantic meaning. I would suggest we use an `Unknown` state here, which we never explicitly set.

brancz (commented):
I think this is good to go once we remove/rename the error state.

pkg/debuginfo/metadata.go
```diff
@@ -0,0 +1,195 @@
...
+
+func (m metadataState) String() string {
+	val, ok := mdStateStr[m]
+	if !ok {
+		return "<not found>"
+	}
+	return val
+}
+
+func (m metadataState) MarshalJSON() ([]byte, error) {
```
javierhonduco, line 70: This is very nice, thanks for doing it!!

Issue #1121, opened 2022-06-03
CPU profile: timestamp column not found


marselester:
Parca v0.11.1 returns `timestamp column not found` error when I try to query "Process CPU Samples" of a newly added Go app `B`. Memory profiles work ok though.

```
level=error name=parca ts=2022-06-03T15:02:34.475796457Z caller=logger.go:35 protocol=grpc grpc.component=server grpc.service=parca.query.v1alpha1.QueryService grpc.method=QueryRange grpc.method_type=unary grpc.start_time=2022-06-03T15:02:34Z grpc.code=Unknown grpc.error="timestamp column not found" grpc.time_ms=2.319 finishedcall=(MISSING)
```

The CPU profiles of an older Go app `A` are still shown properly.

dogukanzorlu:
actually it's not a real error. When there is only one instance or job, if your query is equal to "not equal", you get a timestamp error because there is no label to show the result. Maybe They can show a better error message or add validation. :)

marselester:
Interesting :) I get this error message when I filter by instance or by job of the app `B` (newly added). But there are no errors when filtering on the app `A` (old).

dogukanzorlu:
as I said app B most likely has a single instance or single job. but still, for more accurate bug detection, it is useful to share your queries and target settings in a way that is safe for you.

marselester:
Oh I see what you mean. BTW, both apps have single instance/job per app.

Queries with an error (APP_B):

```
https://parca.default/?currentProfileView=icicle&expression_a=process_cpu%3Asamples%3Acount%3Acpu%3Ananoseconds%3Adelta%7Binstance%3D%22APP_B.default%3A8081%22%7D&from_a=1654300487348&to_a=1654304087348&merge_a=false&time_selection_a=relative:hour|1
https://parca.default/?expression_a=process_cpu%3Asamples%3Acount%3Acpu%3Ananoseconds%3Adelta%7Bjob%3D%22APP_B%22%7D&from_a=1654300603268&to_a=1654304203268&merge_a=false&time_selection_a=relative:hour|1&currentProfileView=icicle
```

Queries without an error (APP_A):

```
https://parca.default/?currentProfileView=icicle&expression_a=process_cpu%3Asamples%3Acount%3Acpu%3Ananoseconds%3Adelta%7Binstance%3D%22APP_A.default%3A8080%22%7D&from_a=1654300668860&to_a=1654304268860&merge_a=false&time_selection_a=relative:hour|1
[...]

PR #1140, opened 2022-06-09
Better error handling when db result is empty

marselester:
In https://github.com/parca-dev/parca/issues/1121 @brancz suggested to change `timestamp column not found` error to be handled differently.
I am not sure how exactly it should be handled, so at least here is a test case to start a conversation :)

### DIFF
--- a/pkg/query/columnquery.go
+++ b/pkg/query/columnquery.go
@@ -283,6 +283,12 @@ func (q *ColumnQueryAPI) QueryRange(ctx context.Context, req *pb.QueryRangeReque
 	if err != nil {
 		return nil, err
 	}
+	if ar.NumRows() == 0 {
+		return nil, status.Error(
+			codes.NotFound,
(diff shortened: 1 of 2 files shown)


brancz:
I'm so sorry for missing this. I would suggest that what we can do is return immediately a (new) `NoData` error just before we try to find the timestamp column: 

https://github.com/parca-dev/parca/blob/726f66e51935ac5fc32aba807e5d8c5f8a88133e/pkg/query/columnquery.go#L287

As if `ar.NumRows() == 0` then there is nothing to show anyways (and the timestamp column won't be there). Then we can return a `NotFound` status code in gRPC with a message such as "No data found for the query, try a different query or time range or no data has been written to be queried yet."

marselester:
No worries, I should have posted the link in the issue. Please have a look at the code changes.

brancz:
This looks great! lgtm 🎉 👍

PR #1141, opened 2022-06-09
feat: auto-reload configuration

### LINKED ISSUE #179: Auto reload configuration
This may not be safe in every environment, but when delpoyed in Kubernetes for example it can be nice to reload configuration automatically as it is provisioned atomically anyways by secrets/configmaps.

maxbrunet:
Inspired from https://github.com/prometheus/prometheus/blob/5d1756c8228f457fb08b63e537ceadba2d0b96ec/cmd/prometheus/main.go#L1174-L1216
It uses: https://github.com/fsnotify/fsnotify

It currently does not reload Debuginfod-related components, if it is desired, let me know, it will be a separate PR (maybe issue).

Tested on Linux. Testers for MacOS or Windows welcome.

Closes #179


pkg/scrape/manager.go
```diff
@@ -113,16 +116,6 @@ func NewManager(
 	}
 	m.scrapeConfigs = c
 
-	// Cleanup and reload pool if config has changed.
-	for name, sp := range m.scrapePools {
```
maxbrunet, line 117: I have removed this block as I think it does nothing, let me know if I am wrong and I'll add it back.

It seems that at this point `m.scrapePools` is still an empty slice. And `m.ApplyConfig()` is called after `NewManager()` which itself calls `sp.reload()`

pkg/config/config.go
```diff
@@ -242,7 +242,7 @@ func (c *ScrapeConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	// Validate the scrape and timeout internal configuration. When /debug/pprof/profile scraping
 	// is enabled we need to make sure there is enough time to complete the scrape.
 	if c.ScrapeTimeout > c.ScrapeInterval {
-		return fmt.Errorf("scrape timeout must be larger or equal to inverval for: %v", c.JobName)
```
maxbrunet, line 245: The error message was not in line with the condition. I assume the condition is correct and we do not want to avoid concurrent scrapes

pkg/scrape/scrape.go
```diff
@@ -180,7 +180,7 @@ func (sp *scrapePool) reload(cfg *config.ScrapeConfig) {
 	for fp, oldLoop := range sp.loops {
 		var (
 			t       = sp.activeTargets[fp]
-			s       = &targetScraper{Target: t, client: sp.client, timeout: timeout}
+			s       = &targetScraper{Target: t, logger: sp.logger, client: sp.client, timeout: timeout}
```
maxbrunet, line 183: There was a bug here, `logger` was left `nil`, causing Parca to panic as soon as the new scraper tries to log

pkg/config/config.go
```diff
@@ -242,7 +242,7 @@ func (c *ScrapeConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+		return fmt.Errorf("scrape timeout must be smaller or equal to inverval for: %v", c.JobName)
```
maxbrunet, line 245: The error message was not in line with the condition. I assume the condition is correct and we want to avoid concurrent scrapes

pkg/config/reloader.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	l := []interface{}{"msg", "completed loading of configuration file", "filename", r.filename, "totalDuration", time.Since(start)}
+	level.Info(r.logger).Log(append(l, timings...)...)
+	return nil
+}
+
+// Run starts watching the config file and wait for reload triggers.
+func (r *ConfigReloader) Run(ctx context.Context) error {
+	go r.watchFile()
```
metalmatze, line 152: It probably makes sense to use the `run.Group` we use here too. 
Not sure we will notice if this goroutine stops right now.

metalmatze (commented):
Really nice work! :+1: 
One suggestion to improve the goroutines within the config reloader. Let me know what you think.

pkg/config/reloader.go
maxbrunet: It seems it would only be useful if the goroutine can return an error, but here it only exits if one of the watcher's channels is closed, it happens when `watcher.Close()` is called, which is the normal behavior when Parca exits. I have added debug logs to troubleshoot potential issues around the channel closing if ever it arises, please let me know if it is enough

fyi (and not that it justifies the handling of `watchFile()`) there is this goroutine which is in the same case:
https://github.com/parca-dev/parca/blob/1c12bfb5ca4009c45f61c0aa711c3b5b8a4a6e7c/pkg/scrape/manager.go#L157-L160

If I understand correctly, it triggers config reload for the service discovery, reloads which used to panic as I pointed above, probably nobody had a case where discovery would change a target's config

metalmatze: Ok. Let's try with the logs for now. :+1:

maxbrunet:
@metalmatze @brancz anything blocking this PR from being merge?

pkg/config/reloader.go
```diff
@@ -0,0 +1,166 @@
...
+			Name: "parca_config_last_reload_successful",
+			Help: "Whether the last configuration reload attempt was successful.",
+		}),
+		configSuccessTime: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "parca_config_last_reload_success_timestamp_seconds",
+			Help: "Timestamp of the last successful configuration reload.",
+		}),
+	}
+
+	reg.MustRegister(
```
kakkoyun, line 81: Better to use `Register` and propagate the error, rather than panicing.

pkg/config/reloader.go
kakkoyun: I think at the very least we should pass the context to the `watchFile` method and check for the cancellation.

pkg/config/reloader.go
```diff
@@ -0,0 +1,166 @@
...
+			failed = true
+		}
+		timings = append(timings, rl.Name, time.Since(rstart))
+	}
+	if failed {
+		return fmt.Errorf("one or more errors occurred while applying the new configuration (--config-path=%q)", r.filename)
+	}
+
+	l := []interface{}{"msg", "completed loading of configuration file", "filename", r.filename, "totalDuration", time.Since(start)}
+	level.Info(r.logger).Log(append(l, timings...)...)
```
kakkoyun, line 148:
```suggestion
	level.Debug(r.logger).Log(append(l, timings...)...)
```
nit: Debug would be a better fit, I think.

kakkoyun (commented):
I think this is super close. I have pointed out a couple of things. I'm happy to merge it afterwards.

pkg/config/reloader.go
maxbrunet: It happens in cascade from the parent goroutine, do you still think it necessary to pass it directly in addition to the channels closure?

```go
		case <-ctx.Done():
			r.watcher.Close()
			return nil
		}
```

https://github.com/parca-dev/parca/pull/1141/files#diff-c1a28a96ac2f667229d0cfd4d99f4b4106c61ce7626adeedbd4a4c03f54e09e3R161-R164

pkg/config/reloader.go
maxbrunet: IMHO Info is the right level, it is the only log informing the user their config has been reloaded (by default). When re-configuring Prometheus/Alertmanager, I often look for this message

pkg/config/reloader.go
maxbrunet: `reg.MustRegister()` is used consistently in Parca, I have no problem changing for `Register()` here, just curious, what makes this one different?

kakkoyun: IMHO, we should always prefer the methods that return errors over panicking. In this case, the constructor method already returns an error and we handle it on the caller site. If we have usage that fits this type of error handling we should convert them as well.

pkg/config/reloader.go
kakkoyun: I don't have a strong opinion on this. I think If everything works well, logging is really necessary. We already use logs extensively (thanks to me :grin:), it just creates noise.

pkg/config/reloader.go
kakkoyun: Aha, I have missed this part! Then it should be all good.
Intuitively I expect to see the context passed down and handled in the goroutine itself.

pkg/config/reloader.go
maxbrunet: Makes sense, thank you for the explanation

pkg/config/reloader.go
maxbrunet: My opinion is strong. As a user, I need this feedback when updating the config file to know Parca reacted to it, it can also be used to identify issues that happen after. Enabling debug level would drown it in an insane amount of logs. From my experience with Prometheus/Alertmanager, it will only log every few days in most cases

kakkoyun: Alright. Sounds good to me.

kakkoyun (approved):

PR #1150, opened 2022-06-13
Ignore io.EOF errors when checking dwarf from objstore

metalmatze:
Instead, we will ask the client to upload their debuginfo once more.


kakkoyun (commented):
LGTM.

I'm ok with this approach. However, we might need to do the checks for `Exists` as well.
At least the agent first checks if it exists using the Exist API and then attempts to upload. So we should return false from that API.

PR #1151, opened 2022-06-13
ui: Add last 15min to DateTimeRangePicker

metalmatze:
Also make 15min the default.

I've never used the 3 days time range, however I often find myself typing 5 or 10 min into the time range picker. 15min is probably totally fine, and thus went with that. 

It's more helpful for the deployments of Parca we currently have, until we have actual persistence and long term storage, then we should discuss the time ranges again.


brancz (approved):
Agreed! 💯

manojVivek (approved):
💯

PR #1153, opened 2022-06-13
Build binaries outside the container

### LINKED ISSUE #1152: chore: Reduce the time spend in PR checks
Especially we spend too much time emulating ARM instructions. Since in parca we use statically linked binaries, we don't need to rely on building everything in their native environment or in containers. This should reduce the number of actions we run under QEMU, so the actions get faster.

Let's make sure:
- [ ] The Go binary is statically linked (no need for libc at all) (this is probably the case already just double-check)
- [ ] Cross build UI and the binaries outside of containers and just copy the binaries in (this would be super easy to achieve with GoReleaser, however, to be able to achieve reproducible builds we need to use podman and GoReleaser free version doesn't support it)
- [ ] If it's not getting faster enough we can run the container build action only when the PR is merged or Dockerfiles have changed.

kakkoyun:

Fixes #1152 

### Tasks

- [x] Production Dockerfiles
- [x] CI actions
- [x] Make sure CI action caches exist 
- [x] Re-generate proto (update deps)
- [x] Remove Windows as a target


metalmatze (approved):
So many changes. Phew.
Looks really good and seems to at least half the time our CI needs now.

PR #1157, opened 2022-06-14
debuginfo: Improve debuginfo quality checking using metadata

kakkoyun:
- Add source file hash to the metadata file
- Reduce metadata states
- Check if metadata is stale
- Check and handle the file corruption
  - Using a limited buffer 

Fixes https://github.com/parca-dev/parca/issues/1097

Blocked by https://github.com/parca-dev/parca/pull/1265


pkg/debuginfo/store.go
```diff
@@ -263,16 +295,42 @@ func (s *Store) Upload(stream debuginfopb.DebugInfoService_UploadServer) error {
...
+func check(objFile string) (bool, error) {
+	hasDWARF, err := elfutils.HasDWARF(objFile)
+	if err != nil {
+		if errors.Is(err, io.EOF) {
+			// Failed to read the file, better for client to upload a new version.
+			return true, nil
+		}
+		var fe *elf.FormatError
+		if errors.As(err, &fe) {
+			if strings.Contains(fe.Error(), "bad magic number '[0 0 0 0]'") {
```
brancz, line 324: elf expects to be a very specific magic number, no? I would not just match on `'[0 0 0 0]'`, but any `bad magic number` error.

brancz:
if we've tested this, I'm happy to merge this

kakkoyun:
It's not fully tested. I have parked it for now, and I'll return to it soonish.

kakkoyun:
In subsequent PRs. We should add integration tests for `debuginfo.Store.Exists` and `debuginfo.Store.Upload` using the `debuginfo.Client`.

kakkoyun:
@javierhonduco I'll wait for your input on this before I proceed with this.

pkg/debuginfo/metadata.go
```diff
@@ -27,36 +28,37 @@ import (
...
+	ErrMetadataNotFound = errors.New("debug info metadata not found")
 )
 
 type metadataState int64
 
 const (
 	metadataStateUnknown metadataState = iota
-	// There's no debug info metadata. This could mean that an older version
-	// uploaded the debug info files, but there's no record of the metadata, yet.
-	metadataStateEmpty
```
javierhonduco, line 41: The reason why this state exists is to be able to backfill the metadata files in the future, but happy to delete it if it's not needed!

pkg/debuginfo/store.go
```diff
@@ -152,28 +160,33 @@ func (s *Store) Exists(ctx context.Context, req *debuginfopb.ExistsRequest) (*de
...
+			}
 			return nil, status.Error(codes.Internal, err.Error())
 		}
 
-		h, err := hash.File(dbgFile)
-		if err != nil {
-			return nil, status.Error(codes.Internal, err.Error())
+		if metadataFile.Hash == req.Hash {
+			exists = true
+			if metadataFile.State == metadataStateUploading {
```
javierhonduco, line 175: The hash is set once the file is persisted in `uploaded()` I think, so we don't have to check the state, as we know it will be `uploaded`, no?

Side note: perhaps checking for the state first can be clearer, like:

```
switch state:
  case metadataStateUploaded:
     // we know we should have a hash by now
  case metadataStateUploading:
     // we should check if the file is stale which means that it never completed uploading and the client should re-upload
...
```

What do you think?

pkg/debuginfo/store.go
```diff
@@ -192,87 +205,118 @@ func (s *Store) Upload(stream debuginfopb.DebugInfoService_UploadServer) error {
...
 
 	found, err := s.find(ctx, buildID)
 	if err != nil {
 		return err
 	}
 
+	var objFileHash string
 	if found {
+		if req.GetInfo().Hash != "" {
+			if metadataFile != nil && metadataFile.Hash != "" {
```
javierhonduco, line 240: nit: we don't need the `metadataFile.Hash != ""` check as we will check if it's equal to the sent hash below, and it can't be an empty string

pkg/debuginfo/store.go
```diff
@@ -192,87 +205,118 @@ func (s *Store) Upload(stream debuginfopb.DebugInfoService_UploadServer) error {
...
+		err := fmt.Errorf("failed to update metadata before uploading: %w", err)
+		return status.Error(codes.Internal, err.Error())
 	}
 
-	// At this point we know that we still have a better version of the debug information file,
+	//nolint:forcetypeassert
+	b := s.pool.Get().([]byte)
+	buf := flexbuf.With(b)
+	defer func() {
+		// TODO(kakkoyun): Check if this creates any problems with reused buffers.
```
javierhonduco, line 281: Which kind of problems are you anticipating?

pkg/debuginfo/store.go
```diff
@@ -192,87 +205,118 @@ func (s *Store) Upload(stream debuginfopb.DebugInfoService_UploadServer) error {
...
+		buf.Release()
+		b = b[:0]
+	}()
+
+	// At this point we know that we received a better version of the debug information file,
 	// so let the client upload it.
 	r := &UploadReader{stream: stream}
-	if err := s.bucket.Upload(ctx, objectPath(buildID), r); err != nil {
+	if err := s.bucket.Upload(ctx, objectPath(buildID), io.TeeReader(r, buf)); err != nil {
```
javierhonduco, line 289: This is beautiful, so unixy! :)

pkg/debuginfo/store.go
```diff
@@ -192,87 +205,118 @@ func (s *Store) Upload(stream debuginfopb.DebugInfoService_UploadServer) error {
...
 		msg := "failed to upload"
 		level.Error(s.logger).Log("msg", msg, "err", err)
 		return status.Errorf(codes.Unknown, msg)
 	}
 
-	if err := s.metadataManager.update(ctx, buildID, metadataStateUploaded); err != nil {
-		level.Error(s.logger).Log("msg", "failed to update metadata", "err", err)
+	if err := elfutils.ValidateReader(buf); err != nil {
```
javierhonduco, line 295: I don't think there's a better way to do this, but I am wondering what can we do to reduce memory usage.

My concern is that we will store the debug info files we are uploading in memory which can put a lot of memory pressure in the box. By using the pool, we will make sure that we don't allocate too much, which is good! But the memory will still be used, and when several multi-gigabyte debug files are uploaded at the same time, we will have to store them in memory.

Please correct me if I am wrong and happy to hear what you think!

pkg/debuginfo/store.go
```diff
@@ -313,19 +358,43 @@ func (s *Store) Symbolize(ctx context.Context, m *pb.Mapping, locations ...*meta
...
+		level.Warn(logger).Log("msg", "failed to validate debug information", "err", err)
+		// Mark the file as corrupted, and let the client try to upload it again.
+		err := s.metadataManager.corrupted(ctx, buildID)
+		if err != nil {
+			level.Warn(logger).Log(
+				"msg", "failed to mar debug information",
+				"err", fmt.Errorf("failed to update metadata for corrupted: %w", err),
+			)
+		}
+		if !downloadedFromDebugInfod {
```
javierhonduco, line 375: This looks like a great candidate for another metadata field! We can do this in another PR, but I think it would help to debug gnarly situations, as well as potentially making the code easier to follow

javierhonduco (commented):
Great stuff!! Loving this, and learned a bunch of Go-isms on the way! 

Left a couple of nits and some questions, and will test this on my machine right now!

proto/parca/debuginfo/v1alpha1/debuginfo.proto
```diff
@@ -43,7 +43,7 @@ message UploadInfo {
   // build_id is a unique identifier for the debug data
   string build_id = 1;
 
-  // hash is the hash of the debug information file
+  // hash is the hash of the source file that debug information extracted from
```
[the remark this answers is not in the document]
kakkoyun: But it's not the hash of the debug information. I can work on the wording. However, it was the mistake previously we made we should store the hash of the source we extracted debug information from. I'm open to suggestions for the wording.

pkg/symbolizer/symbolizer.go
```diff
@@ -56,7 +56,8 @@ func (s *Symbolizer) Run(ctx context.Context, interval time.Duration) error {
 
 		err = s.symbolize(ctx, locations)
 		if err != nil {
-			level.Error(s.logger).Log("msg", "symbolization attempt finished with errors", "err", err)
+			level.Warn(s.logger).Log("msg", "symbolization attempt finished with errors")
+			level.Debug(s.logger).Log("err", err)
```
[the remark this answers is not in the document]
kakkoyun: I actually even get rid of these logs, maybe. It really creates clutter and most of the time, it's hard to determine the cause from this log line. I'll think of a better way for this.

proto/parca/debuginfo/v1alpha1/debuginfo.proto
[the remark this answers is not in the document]
javierhonduco: Ohhh! I see, I misunderstood what it was, I thought it was `hash(extracted_debug_file)` instead of `hash(original_executable)`. Everything makes sense now 😄

pkg/symbol/elfutils/elfutils.go
```diff
@@ -153,3 +154,33 @@ func HasSymbols(filePath string) (bool, error) {
 	}
 	return false, nil
 }
+
+// ValidateFile returns true if the given object file is valid.
+func ValidateFile(path string) error {
```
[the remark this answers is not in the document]
kakkoyun: I have forgotten to update the comment :see_no_evil: I had a lot of back and forth on this. None of the errors is an unrecoverable error in this context. And the existence of the error indicates if it's valid or not. That's why I went for a simpler approach. Otherwise, error handling gets super messy.

pkg/debuginfo/metadata.go
```diff
@@ -96,98 +98,112 @@ func newMetadataManager(logger log.Logger, bucket objstore.Bucket) *metadataMana
 
 type metadata struct {
 	State            metadataState `json:"state"`
-	StartedUploadAt  int64         `json:"started_upload_at"`
-	FinishedUploadAt int64         `json:"finished_upload_at"`
+	Hash             string        `json:"hash"`
+	UploadStartedAt  int64         `json:"upload_started_at"`
```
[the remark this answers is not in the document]
kakkoyun: And I haven't noticed it during reviews :D

pkg/debuginfo/metadata.go
```diff
@@ -96,98 +98,112 @@ func newMetadataManager(logger log.Logger, bucket objstore.Bucket) *metadataMana
...
+	UploadFinishedAt int64         `json:"upload_finished_at"`
 }
 
-func (m *metadataManager) update(ctx context.Context, buildID string, state metadataState) error {
```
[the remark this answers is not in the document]
kakkoyun: I'm happy to improve the design. Leaking the internal state type to the caller just feels wrong. We should just carry over the state machine logic into these methods to have the best of both worlds. But I'll leave it to consecutive PRs.

pkg/symbol/elfutils/elfutils.go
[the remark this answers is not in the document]
javierhonduco: Sure thing! Either way is fine :D

pkg/debuginfo/metadata.go
kakkoyun: The unknown state or lack of the metadata file implies this state. Since we don't have a lot of historical data, this shouldn't be a huge problem, don't you think?

pkg/debuginfo/store.go
kakkoyun: Good point. I think I've overlooked this.

pkg/debuginfo/metadata.go
```diff
@@ -96,98 +98,112 @@ func newMetadataManager(logger log.Logger, bucket objstore.Bucket) *metadataMana
...
-		_, err := m.bucket.Get(ctx, metadataObjectPath(buildID))
-		// The metadata file should not exist yet. Not erroring here because there's
-		// room for a race condition.
-		if err == nil {
-			level.Info(m.logger).Log("msg", "there should not be a metadata file")
-			return nil
-		}
+func (m *metadataManager) corrupted(ctx context.Context, buildID string) error {
+	if err := m.write(ctx, buildID, &metadata{
+		State: metadataStateCorrupted,
```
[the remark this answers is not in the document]
javierhonduco: That's fair, it's definitely not a big deal. I was thinking of being able to inspect the files for patterns and correlations, such as trying to see if more corrupted files happen after a certain time (imagine, a new release, or if the server is overloaded, etc).

But I guess that as you said, the window for inspecting such files is small as they'll most likely be overwritten later on. 

We can always use some Prometheus counters to understand the failures and find correlations!

pkg/debuginfo/store.go
kakkoyun: I haven't properly checked if we reuse the byte slices underneath for these buffers. Because the wrapper buffers don't check boundaries while reading, they read until the end of the []byte underneath. This could be a problem if we got a reused []byte that is larger than the current data. I might be overthinking and missing things in here. That's why I want to double-check before merging.

pkg/debuginfo/metadata.go
javierhonduco: Thought it would be nice to have an explicit state for this situation, but it's not an issue to leave it like this :)

pkg/debuginfo/store.go
javierhonduco: Make sense, thanks for explaining!! I think that for this part, it would be good to have someone with actual Go expertise to take a look 😄

pkg/debuginfo/store.go
kakkoyun: You're definitely right. That's why this is the last piece that I have added. I really don't want to do this, but it's the only way to validate completely what we have received. If we're willing to loosen up the guarantees, we can get by just reading the first few bytes. And if something happens during symbolization, we can mark this file as corrupted (which we already do). We just need to make sure the error that we receive is not recoverable. (And this is what I meant by the last to-do item).

pkg/debuginfo/store.go
kakkoyun: This is a to-do item that I have on the agent side. I think we should add a source type enum file to metadata and the API. So that this could help us debug issues. And we can improve the state machine we have in hear.

kakkoyun: What do you think, shall we just add it as part of this PR?

javierhonduco: Either works, but this PR is already quite comprehensive, should we do it in another PR?

kakkoyun: Created an issue for it https://github.com/parca-dev/parca/issues/1258

pkg/debuginfo/store.go
```diff
@@ -192,87 +202,118 @@ func (s *Store) Upload(stream debuginfopb.DebugInfoService_UploadServer) error {
...
-		level.Error(s.logger).Log("msg", "failed to update metadata", "err", err)
+	if err := s.metadataManager.markAsUploading(ctx, buildID); err != nil {
+		err := fmt.Errorf("failed to update metadata before uploading: %w", err)
+		return status.Error(codes.Internal, err.Error())
 	}
 
-	// At this point we know that we still have a better version of the debug information file,
+	//nolint:forcetypeassert
+	b := s.pool.Get().(*bytes.Buffer)
+	// NOTICE: The ELF header is 52 or 64 bytes long for 32-bit and 64-bit binaries respectively.
```
javierhonduco, line 276: Oh, this is great to know and take into account!! 😄

pkg/debuginfo/store.go
```diff
@@ -192,87 +202,118 @@ func (s *Store) Upload(stream debuginfopb.DebugInfoService_UploadServer) error {
...
-	if err := s.bucket.Upload(ctx, objectPath(buildID), r); err != nil {
+	if err := s.bucket.Upload(ctx, objectPath(buildID), io.TeeReader(r, buf)); err != nil {
 		msg := "failed to upload"
 		level.Error(s.logger).Log("msg", msg, "err", err)
 		return status.Errorf(codes.Unknown, msg)
 	}
 
-	if err := s.metadataManager.update(ctx, buildID, metadataStateUploaded); err != nil {
-		level.Error(s.logger).Log("msg", "failed to update metadata", "err", err)
+	if err := elfutils.ValidateHeader(b); err != nil {
```
javierhonduco, line 292: Nice!!

pkg/symbol/elfutils/elfutils.go
```diff
@@ -153,3 +158,109 @@ func HasSymbols(filePath string) (bool, error) {
...
+
+	// TODO(kakkoyun): How can we improve this without allocating too much memory.
+	if len(elfFile.Sections) == 0 {
+		return errors.New("ELF does not have any sections")
+	}
+	return nil
+}
+
+// ValidateHeader returns an error if the given object file header is not valid.
+func ValidateHeader(r io.Reader) error {
```
javierhonduco, line 178: This is fantastic!! Love how you are validating a bunch of things I did not even think about!! 💯

javierhonduco (approved):
Looks great!! 

Given that there are many changes here, could you run it for a little while on some test binary and see if everything works as expected? 

🚢

kakkoyun:
This is still being tested and blocked by https://github.com/parca-dev/parca/pull/1265

javierhonduco (approved):
Amazing work!

pkg/debuginfo/metadata.go
```diff
@@ -96,98 +98,120 @@ func newMetadataManager(logger log.Logger, bucket objstore.Bucket) *metadataMana
 
 type metadata struct {
 	State            metadataState `json:"state"`
-	StartedUploadAt  int64         `json:"started_upload_at"`
-	FinishedUploadAt int64         `json:"finished_upload_at"`
+	BuildID          string        `json:"build_id"`
```
brancz, line 101: Why is this needed? The BuildID is already in the path of the object, no?

pkg/debuginfo/store.go
```diff
@@ -214,72 +253,89 @@ func (s *Store) Upload(stream debuginfopb.DebugInfoService_UploadServer) error {
...
+	if err := s.metadataManager.markAsUploading(ctx, buildID); err != nil {
+		err = fmt.Errorf("failed to update metadata before uploading: %w", err)
+		return status.Error(codes.Internal, err.Error())
 	}
 
-	// At this point we know that we still have a better version of the debug information file,
+	//nolint:forcetypeassert
+	b := s.pool.Get().(*bytes.Buffer)
+	// NOTICE: The ELF header is 52 or 64 bytes long for 32-bit and 64-bit binaries respectively.
+	buf := limitio.NewWriter(b, 64, true)
```
brancz, line 296: This took me a couple of reads to understand, and I have a few questions about this.

1) Is the pooling really necessary? It seems that we'd only ever allocate max 64 bytes per upload.
2) I think for future me I would need a comment that this limitio Writer does not error when you attempt to write more than 64 bytes.
3) If I'm reading the logic right, then we're not actually validating the elf header until after we uploaded to the bucket. Isn't the whole point to prevent that?

pkg/debuginfo/metadata.go
javierhonduco: I think it would be convenient to have the build_id here, in general. Especially since it's typed and not a loose string that might change in the future and might make use add awkward build_id parsing out of the path

brancz: I don't feel strongly. I'm ok with keeping it.

pkg/debuginfo/store.go
kakkoyun: 1. Yes, pooling is not strictly necessary here. Since this is a frequent recurring task, I've put the pool there. If you think it's overkill, I'm happy t remove it.
2. I should have commented and added an explanation. I will do that.
3. Yes, this happens after the fact. We're optimistically uploading directly to the bucket and if something goes wrong we mark it as corrupted, so it could be overwritten, in subsequent calls. We only want to make sure we don't read a corrupted file while symbolizing. I wanted to prevent any form of buffering this data on the server-side thus the optimistic writes directly to the object-store. I'm open to other suggestions.

brancz: Oooooooo I see! Could you add what you wrote in 3 as a comment as well? :D 

Now it's all clear 💯 

(and yeah I think the pool is unnecessary)

brancz (approved):
Very nice! 💯

PR #1158, opened 2022-06-14
area/ui: Use a default 15mins value for DateTimePicker

yomete:


metalmatze:
Thank you. :100: 
Seems like I wasn't able to find that yesterday. :grimacing:

metalmatze (approved):

PR #1183, opened 2022-06-22
*: Fix TODOs from metastore refactor

brancz:


metalmatze (approved):

pkg/metastore/test.go
```diff
@@ -0,0 +1,23 @@
...
+
+	pb "github.com/parca-dev/parca/gen/proto/go/parca/metastore/v1alpha1"
+)
+
+func NewTestMetastore(
+	t require.TestingT,
+	logger log.Logger,
+	reg prometheus.Registerer,
+	tracer trace.Tracer,
+) pb.MetastoreServiceServer {
```
maxbrunet, line 17: I think it is good practice to call `t.Helper()` in these kind of functions, it outputs the line of the test if it fails instead of the helper

https://pkg.go.dev/testing#T.Helper

brancz:
I think moving a lot of this to black-box tests sounds like a good idea, but I think maybe a bit too drastic of a change within the scope of this PR.

PR #1188, opened 2022-06-24
chore: prepare dependencies for Renovate

maxbrunet:
Similar to parca-dev/parca-agent#485

And:
* Add `.go-version` file (double usage: GH Actions and version managers)
* Add `.node-version` file (double usage: GH Actions and `nodenv`, consistent name with `.go-version`)
* Add `.nvmrc` because `nvm` is more popular than `nodenv` and it does not cost more
* Remove Alpine version from `golang` and `node` container images. Only the language version is important and the digest is pinned so we should notice if an update causes an issue
* Downgrade `node` container image to match version used in CI


maxbrunet:
This does not seem related to my changes, only because the workflow is triggered

```diff
buf generate
diff --git a/proto/buf.lock b/proto/buf.lock
index ba73ab0..dc85128 100644
--- a/proto/buf.lock
+++ b/proto/buf.lock
@@ -4,7 +4,7 @@ deps:
   - remote: buf.build
     owner: googleapis
     repository: googleapis
-    commit: 1068d85d09894f7587a3236f489c0d80
+    commit: 8f04c867ddf14e13a231d5119b8ebb8b
   - remote: buf.build
     owner: grpc-ecosystem
     repository: grpc-gateway
Error: Process completed with exit code 1.
```

kakkoyun (approved):
LGTM :100:

kakkoyun:
> This does not seem related to my changes, only because the workflow is triggered
> 
> ```diff
> buf generate
> diff --git a/proto/buf.lock b/proto/buf.lock
> index ba73ab0..dc85128 100644
> --- a/proto/buf.lock
> +++ b/proto/buf.lock
> @@ -4,7 +4,7 @@ deps:
>    - remote: buf.build
>      owner: googleapis
>      repository: googleapis
> -    commit: 1068d85d09894f7587a3236f489c0d80
> +    commit: 8f04c867ddf14e13a231d5119b8ebb8b
>    - remote: buf.build
>      owner: grpc-ecosystem
>      repository: grpc-gateway
> Error: Process completed with exit code 1.
> ```

We need to update the buf dependencies. `make proto/all`. It could be handled in this PR or in another one.

kakkoyun:
We need to check if the error is an actual error or a flake. I have run it a couple of times. It seems like something broken.

maxbrunet:
> We need to check if the error is an actual error or a flake. I have run it a couple of times. It seems like something broken.

Yup, I had missed a hard-to-read quoting/line-escape, I have fixed it and rearranged  all the dockerized steps to make them more readable

kakkoyun (approved):
LGTM :heart: Awesome work @maxbrunet as always! We really appreciate your contributions.

PR #1190, opened 2022-06-24
area/ui: Reset SearchNodes component after dashboard actions

### LINKED ISSUE #1181: Search state isn't reset when changing profile (type, timestamp)
When searching for a stack trace in one profile and then changing the profile's time or even the type entirely the search state stays the same. While I can see that making sense for the exact same profile series (so labels not changing between queries) it doesn't make sense if looking at entirely new profiles.

yomete:
Fixes #1181


metalmatze (approved):
🚀

PR #1227, opened 2022-06-28
ui: Checking out the main branch before publish

manojVivek:
Attempting a fix to Lerna publish issue on the GitHub action here: https://github.com/parca-dev/parca/runs/7087395220?check_suite_focus=true


manojVivek:
Merging to test the fix, please leave if you have any comments.

PR #1268, opened 2022-06-29
chore: renovate buf plugins

maxbrunet:
Because we certainly can. It will be a bit clunky though, as the Git tags do not match one-to-one with Buf tags and Renovate will not be able to run `buf generate`. It should at least avoid having to keep an eye out for releases. `buf` is popular enough, it may be feature-request worthy

Note: the `cerbos` org has been renamed to `planetscale`, this will require another PR to update the plugin

### DIFF
--- a/buf.gen.yaml
+++ b/buf.gen.yaml
@@ -7,28 +7,32 @@ managed:
       - buf.build/googleapis/googleapis
 
 plugins:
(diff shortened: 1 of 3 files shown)


kakkoyun (approved):
LGTM. Thanks.
Let's see if this would break anything with  upgrades.

PR #1269, opened 2022-06-30
ci: filter paths with a GH action

maxbrunet:
An attempt at solving issues with required checks

> * GitHub's `paths` does not work with required checks that you really want to pass successfully.

https://github.com/marketplace/actions/skip-duplicate-actions#skip-if-paths-not-changed

Related to parca-dev/parca-agent#509


kakkoyun:
@maxbrunet does this mean we can have more required checks? Say we want to mark a UI as required and it would only run for UI changes? And for non-UI changes, the overall result of the checks will be green?

maxbrunet:
> @maxbrunet does this mean we can have more required checks? Say we want to mark a UI as required and it would only run for UI changes? And for non-UI changes, the overall result of the checks will be green?

Yes!

kakkoyun (approved):

PR #1278, opened 2022-06-30
Update FrostDB for Parca

metalmatze:
Changes: https://github.com/polarsignals/frostdb/compare/571b4dc57775...42eb27a93c4b

PR #1282, opened 2022-06-30
Revert "Update FrostDB for Parca"

brancz:
Reverts parca-dev/parca#1278

Unfortunately, this appears to have introduced some unexpected errors that we weren't getting before.


kakkoyun (approved):

metalmatze:
Alright. Let's revert and take the time to fix it before putting it in a release.

Issue #1258, opened 2022-06-29
debuginfo: Add source type enum to the API and Metadata


kakkoyun:
This looks like a great candidate for another metadata field! Something like `source: upload | debuginfod | othersoruce`  We can do this in another PR, but I think it would help to debug gnarly situations, as well as potentially making the code easier to follow

_Originally posted by @javierhonduco in https://github.com/parca-dev/parca/pull/1157#discussion_r902794396_

PR #1294, opened 2022-07-04
*: Separate debuginfo and symbolizer components further

brancz:
This PR:

* Introduces a new gRPC method to download debuginfos
* Separates the symbolization process from the debuginfo gRPC service
* Abstracts away the debuginfo metadata store


pkg/parca/parca.go
```diff
@@ -256,7 +276,15 @@ func Run(ctx context.Context, logger log.Logger, reg *prometheus.Registry, flags
...
 	{
-		s := symbolizer.New(logger, metastore, dbgInfo)
+		s := symbolizer.New(
+			logger,
+			metastore,
+			dbgInfoMetadata,
+			dbgInfo,
+			debugInfodClient,
+			sym,
+			"", // TODO(brancz): must be fixed before merging!!!
```
brancz, line 286: this still needs fixing

pkg/debuginfo/store.go
```diff
@@ -306,25 +286,72 @@ func (s *Store) upload(ctx context.Context, buildID, hash string, r io.Reader) e
...
 
 	return nil
 }
 
-func isStale(metadataFile *metadata) bool {
+func isStale(metadataFile *Metadata) bool {
 	return time.Now().Add(-15 * time.Minute).After(time.Unix(metadataFile.UploadStartedAt, 0))
 }
 
+func (s *Store) Download(req *debuginfopb.DownloadRequest, stream debuginfopb.DebugInfoService_DownloadServer) error {
```
javierhonduco, line 308: Should we fetch the metadata here and log an error if the state is not "uploaded" as this might indicate an issue?

javierhonduco (approved):
LGTM, just left a minor comment! 

I am sure things are working fine, but would you mind running and end-to-end test with https://github.com/parca-dev/parca-demo/blob/main/cpp/Makefile#L20 to double-check?

pkg/debuginfo/store.go
brancz: Very good point, we should even be more granular in the gRPC return codes! In regards to metadata should we then make it:

* Corrupted: `FAILED_PRECONDITION`
* Unknown, Uploaded, we just attempt to download
* Uploading: `UNAVAILABLE`

What do you think?

javierhonduco: This would be perfect! Just for context for others checking out this PR, we will download on 'unknown' as older Parca installations did not generate metadata

pkg/debuginfo/client.go
```diff
@@ -73,7 +73,7 @@ func (c *Client) Upload(ctx context.Context, buildID, hash string, r io.Reader)
 	}
 
 	reader := bufio.NewReader(r)
-	buffer := make([]byte, 1024)
+	buffer := make([]byte, 1024*1024*5)
```
kakkoyun, line 76: How about introducing a human-readable constant rather than a magic number? And also it would be amazing if add a comment about the implications of changing the buffer size?

pkg/debuginfo/store.go
```diff
@@ -306,25 +286,72 @@ func (s *Store) upload(ctx context.Context, buildID, hash string, r io.Reader) e
...
+	if err != nil {
+		return status.Error(codes.Internal, err.Error())
+	}
+
+	f, err := os.Open(objFile)
+	if err != nil {
+		return status.Error(codes.Internal, err.Error())
+	}
+
+	buf := make([]byte, 1024*1024*5)
```
kakkoyun, line 333: ditto

kakkoyun (commented):
Amazing! Thanks a lot for doin this :pray:

proto/parca/debuginfo/v1alpha1/debuginfo.proto
```diff
@@ -58,3 +61,28 @@ message UploadResponse {
...
+    // chunk_data is the raw bytes of the debug info
+    bytes chunk_data = 2;
+  }
+}
+
+// DownloadInfo metadata for the debug data that is being downloaded.
+message DownloadInfo {
+  // from_public_debuginfod is true if the debug info is from public
+  // debuginfod server.
+  bool from_public_debuginfod = 1;
```
kakkoyun, line 87: How about introducing an enum for `source`? It would be more future-proof.
We already have an issue on it https://github.com/parca-dev/parca/issues/1258, if we change this now we don't need to change it in the future.

brancz: Sounds good!

kakkoyun (approved):

PR #1306, opened 2022-07-05
ci: disable skip_after_successful_duplicate

maxbrunet:
Follow up to #1303 

I did not intend to use the skip duplicate feature, I think we are only interested in path filtering for now.

https://github.com/fkirc/skip-duplicate-actions#skip_after_successful_duplicate


kakkoyun (approved):
LGTM

PR #1320, opened 2022-07-07
ci: use shell option to run commands in Docker

maxbrunet:
A minor change which greatly improves readability of this workflow and some cleanup

It also uses the [`buildpack-deps`](https://hub.docker.com/_/buildpack-deps) image which already includes all the build tools and spares us from installing them ourselves


kakkoyun (approved):
LGTM

PR #1336, opened 2022-07-11
ui: Sort targets on Targets page

metalmatze:
While the URLs within one target are already sorted and stable upon refreshing the page, the targets themselves where often in random order. 
With this change, the targets are always sorted by their name. 

The page is now consistently showing `parca` before `prometheus`:
parca-targets


yomete (approved):
Thanks for the enhancement to the Targets table. I added a small commit to extract the sorting logic into a function. Other than that I think it's g2g!

kakkoyun (approved):

Issue #1334, opened 2022-07-11
Add loading state to ProfileSelector


metalmatze:
The ProfileSelector doesn't indicate when the `ProfileTypes` query is still loading, causing users to click too early and potentially think that Parca is broken. 

![profiletypes](https://user-images.githubusercontent.com/872251/178231096-3dce5ccc-5e79-4b8e-a36c-c23cec0a6b17.png)
The [ProfileSelector](https://github.com/parca-dev/parca/blob/main/ui/packages/shared/components/src/ProfileSelector/index.tsx) should handle the loading state while the request is still in-flight.

One way to reproduce this is to add a 
``` 
time.Sleep(5 * time.Second)
```
to the backend:
https://github.com/parca-dev/parca/blob/11249314d0c5f1a66f36d1a5526a980e0fa6e73f/pkg/query/columnquery.go#L459-L460

PR #1337, opened 2022-07-11
ui: Added loading state to the ProfileType selector

### LINKED ISSUE #1154: ui: Indicate the loading state of profile types data
Currently, when a user clicks on the profile selector while the data is being fetched, the dropdown doesn't react to the click. This gives a perception that something is broken, instead we could indicate the loading state with some kind of loader.

Screenshot 2022-06-14 at 8 47 25 AM

manojVivek:
Fixes #1154 & #1334


yomete (approved):
LGTM!

PR #1348, opened 2022-07-12
debuginfo/client: Fix grpc: received message larger than max error

kakkoyun:

I haven't properly tested it yet. However, I believe these changes would fix the ` grpc: received message larger than max` errors.

### DIFF
--- a/pkg/debuginfo/client.go
+++ b/pkg/debuginfo/client.go
@@ -57,7 +57,7 @@ func (c *Client) Exists(ctx context.Context, buildID, hash string) (bool, error)
 }
 
 func (c *Client) Upload(ctx context.Context, buildID, hash string, r io.Reader) (uint64, error) {
-	stream, err := c.c.Upload(ctx)
+	stream, err := c.c.Upload(ctx, grpc.MaxCallSendMsgSize(ChunkSize))
 	if err != nil {
 		return 0, fmt.Errorf("initiate upload: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


brancz:
I think these should be passed in the dialer and server, no? https://hashnode.com/post/quick-tips-increase-grpc-message-size-in-go-ckayjo5za01m3l6s1yw5pkb56

I would say since we deal with fairly large messages because of the uploads, let's set this to 32MB.

kakkoyun:
> I think these should be passed in the dialer and server, no? https://hashnode.com/post/quick-tips-increase-grpc-message-size-in-go-ckayjo5za01m3l6s1yw5pkb56
> 
> I would say since we deal with fairly large messages because of the uploads, let's set this to 32MB in all directions.

Let me check and update the PR accordingly (or client-side)

pkg/debuginfo/client.go
```diff
@@ -29,10 +29,14 @@ import (
...
-// uploaded and downloaded. AWS S3 has a minimum of 5MB for multi part uploads
-// and a maximum of 15MB, and a default of 8MB.
-var ChunkSize = 1024 * 1024 * 8
+const (
+	// ChunkSize 8MB is the size of the chunks in which debuginfo files are
+	// uploaded and downloaded. AWS S3 has a minimum of 5MB for multi-part uploads
+	// and a maximum of 15MB, and a default of 8MB.
+	ChunkSize = 1024 * 1024 * 8
+	// MaxMsgSize is the maximum message size the server can receive or send. By default, it is 4MB.
+	MaxMsgSize = 1024 * 1024 * 32
```
kakkoyun, line 38: Exported, so that we can use it on the client-side as well.

pkg/debuginfo/client.go
```diff
@@ -57,7 +61,7 @@ func (c *Client) Exists(ctx context.Context, buildID, hash string) (bool, error)
 }
 
 func (c *Client) Upload(ctx context.Context, buildID, hash string, r io.Reader) (uint64, error) {
-	stream, err := c.c.Upload(ctx)
+	stream, err := c.c.Upload(ctx, grpc.MaxCallSendMsgSize(MaxMsgSize))
```
kakkoyun, line 64: Probably overkill

PR #1378, opened 2022-07-18
Create an initial ADOPTERS.md

metalmatze:
Let me know what you think of this first draft.
This is intentionally very simple.

Once we have a handful of users on this list, we can still think about adding those companies with logos to the parca.dev website eventually. :slightly_smiling_face:

### DIFF
--- a/ADOPTERS.md
+++ b/ADOPTERS.md
@@ -1,0 +1,33 @@
+<!--
+
+Inspired by https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/main/ADOPTERS.md
+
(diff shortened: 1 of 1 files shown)


manojVivek:
Looks great! 🎉 

Should we also add a call to action/reference to this somewhere on the Readme, so that people are aware of this?

kakkoyun (approved):

brancz (approved):

metalmatze:
> Looks great! tada
> 
> Should we also add a call to action/reference to this somewhere on the Readme, so that people are aware of this?

It worked for the Prometheus Operator without specifically calling it out. The idea to link to it from the README is great though, we should do that!

PR #1395, opened 2022-07-20
chore: Upgrade golangci-lint to 1.47

kakkoyun:
Enable disabled linters

Fix issues


metalmatze (approved):

pkg/metastore/kv.go
```diff
@@ -85,10 +85,18 @@ func MakeLocationID(l *pb.Location) string {
 		// 0. This works out well as the key is byte aligned to the nearest 8
 		// bytes that way.
 
-		//nolint:errcheck // ignore error as writing to the hash will cannot error
+		//nolint:errcheck,staticcheck
+		// ignore error as writing to the hash will cannot error
```
metalmatze, line 89:
```suggestion
		// ignore the error as writing to the hash will not error
```

pkg/metastore/kv.go
```diff
@@ -85,10 +85,18 @@ func MakeLocationID(l *pb.Location) string {
...
+		// https://staticcheck.io/docs/checks#SA1003
+		// The encoding/binary package can only serialize types with known sizes.
+		// This precludes the use of the int and uint types, as their sizes differ on different architectures.
+		// TODO: Fix this.
 		binary.Write(hash, binary.BigEndian, 1)
 	} else {
-		//nolint:errcheck // ignore error as writing to the hash will cannot error
+		//nolint:errcheck,staticcheck
+		// ignore error as writing to the hash will cannot error
```
kakkoyun, line 97:
```suggestion
		// ignore the error as writing to the hash will not error
```

pkg/metastore/kv.go
```diff
@@ -85,10 +85,18 @@ func MakeLocationID(l *pb.Location) string {
 		// 0. This works out well as the key is byte aligned to the nearest 8
 		// bytes that way.
 
-		//nolint:errcheck // ignore error as writing to the hash will cannot error
+		//nolint:errcheck,staticcheck
+		// ignore the error as writing to the hash will not error
+		// https://staticcheck.io/docs/checks#SA1003
+		// The encoding/binary package can only serialize types with known sizes.
+		// This precludes the use of the int and uint types, as their sizes differ on different architectures.
+		// TODO: Fix this.
```
brancz, line 93: I don't understand what needs to be fixed here.`hash` is fixed size, so the staticcheck link doesn't apply here.

kakkoyun: If I don't understand this incorrectly, `int` is arch dependent so var sized and staticcheck wants us to specify it.

brancz: Ahhhhhh I see now. Then we need to change this to `uint64(0)` and `uint64(1)`.

kakkoyun:
:see_no_evil: I've forgotten that I've marked this as auto-merge. I'll send a fix for the issue as a follow up

PR #1396, opened 2022-07-20
pkg/profilestore: Return grpc InvalidArgument if profile is not unzippable

metalmatze:
If a profile buffer cannot be unzipped we should probably return `InvalidArgument` to the client and not say it's an internal error.


kakkoyun (approved):
LGTM. 
Why this is happening? Is it a network-related issue, or is it preventable?

brancz (approved):

PR #1411, opened 2022-07-25
ui: Resetting the icicle graph selection when a new profile is selected

### LINKED ISSUE #1406: UI: Selecting a different profile doesn't work as expected
To reproduce:
1. Select a point that opens in icicle view.
2. Now click down to a span in a graph.
3. Now click on a different point for a different process.
4. That results in seeing root 100% for example or other unexpected results.

I can provide a video of the screen if the above doesn't make sense.

manojVivek:
Fixes #1406


brancz (approved):

yomete (approved):
👍🏾

PR #1414, opened 2022-07-25
pkg/parcacol: Improve encodings and compression

brancz:
With these changes we have 13.2x better space efficiency, meaning 13.2x
less in-memory and on-disk usage by columnar data managed by FrostDB.

The majority of the improvements came from choosing better encodings for each column but compression ultimately pushed it over the double-digit.

Running the numbers on a couple of blocks and I'm seeing ~2.91bytes per sample including label metadata and I don't even think we have reached what we can with multiple of these columns.

### DIFF
--- a/pkg/parcacol/ingest.go
+++ b/pkg/parcacol/ingest.go
@@ -74,7 +74,7 @@ func separateNameFromLabels(ls labels.Labels) (string, labels.Labels, error) {
 func (ing Ingester) Ingest(ctx context.Context, ls labels.Labels, p *pprofproto.Profile, normalized bool) error {
 	name, ls, err := separateNameFromLabels(ls)
 	if err != nil {
-		return err
+		return fmt.Errorf("prepare labels: %w", err)
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/parcacol/ingest.go
```diff
@@ -106,21 +106,9 @@ func (ing Ingester) IngestProfile(ctx context.Context, ls labels.Labels, p *prof
...
 	}
 
-	buffer.Sort()
-
-	// This is necessary because sorting a buffer makes concurrent reading not
-	// safe as the internal pages are cyclically sorted at read time. Cloning
-	// executes the cyclic sort once and makes the resulting buffer safe for
-	// concurrent reading as it no longer has to perform the cyclic sorting at
-	// read time. This should probably be improved in the parquet library.
-	buffer, err = buffer.Clone()
```
brancz, line 116: We no longer need this since we serialize the parquet buffer, so effectively it's already being cloned. Also, the sorting was moved to the buffer creation.

thorfour (approved):

PR #1418, opened 2022-07-26
area/ui: Use batch to prevent multiple rerenders

yomete:
This fixes the blank screen that occurs when you do a compare in the Polar Signals cloud product. 

The culprit here was having multiple dispatch events next to each other as that causes React to rerender multiple times. Using the [batch](https://redux.js.org/faq/performance/#how-can-i-reduce-the-number-of-store-update-events) fn fixes that.

I tested this by creating a [prerelease](https://www.npmjs.com/package/@parca/components/v/0.14.3-alpha.2) and using it in my local dev environment of the cloud app, and a working preview can be seen [here](https://cloud-app-git-temp-fix-compare-mode.vercel-preview.polarsignals.com/)

### DIFF
--- a/ui/packages/shared/components/src/ProfileExplorer/index.tsx
+++ b/ui/packages/shared/components/src/ProfileExplorer/index.tsx
@@ -11,8 +11,9 @@ import {
   setSearchNodeString,
   store,
 } from '@parca/store';
-import {Provider} from 'react-redux';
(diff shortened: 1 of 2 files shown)


monicawoj (approved):
rerendering bugs are always the trickiest! great job 💪

manojVivek (approved):
👏🏼 👏🏼

Issue #1429, opened 2022-07-27
Automatically publish `@parca` ui packages on merge to main without manual intervention


manojVivek:

manojVivek:
Verified, it is working fine with a few merges on `main`.

PR #1430, opened 2022-07-27
ci/ui: Using a separate token for UI package publish

manojVivek:
Possible solution to #1429.

The publish action fails while trying to push the tags and new publish commit to the repo as the main branch has push conditions, see: https://github.com/parca-dev/parca/runs/7536913694?check_suite_focus=true#step:8:26

As a first step trying, a PAT from admin user, so the push conditions are bypassed.


yomete (approved):
LGTM!

PR #1431, opened 2022-07-27
fix: stop error on profile with no location id

### LINKED ISSUE #1264: scrape: Sample has no location ids
Based off current `main` (`5277034f80d8f7310e4a0ec740f63b66459a2399`) and with the default config, I am seeing these errors that didn't appeared before:

```
level=error name=parca ts=2022-06-29T14:53:49.215847316Z caller=scrape.go:500 scrape_pool=default target=http://127.0.0.1:7070/debug/pprof/goroutine msg="WriteRaw failed for scraped profile" err="rpc error: code = Internal desc = failed to ingest profile: sample 44 has no location ids"
level=error name=parca ts=2022-06-29T14:53:55.215188678Z caller=scrape.go:500 scrape_pool=default target=http://127.0.0.1:7070/debug/pprof/goroutine msg="WriteRaw failed for scraped profile" err="rpc error: code = Internal desc = failed to ingest profile: sample 44 has no location ids"
level=error name=parca ts=2022-06-29T14:53:58.215063347Z caller=scrape.go:500 scrape_pool=default target=http://127.0.0.1:7070/debug/pprof/goroutine msg="WriteRaw failed for scraped profile" err="rpc error: code = Internal desc = failed to ingest profile: sample 44 has no location ids"
```

jnsgruk:
At present, when ingesting a profile that doesn't set a location id an error is logged. This stops that error from being logged.

This fix is simple, and the tests pass. I've just run the resultant binary in the situation where Parca was previously erroring every few seconds and things seem to be stable - perhaps there is some nuance I've missed elsewhere in the code, though?

Fixes #1264

### DIFF
--- a/pkg/parcacol/ingest.go
+++ b/pkg/parcacol/ingest.go
@@ -189,15 +189,14 @@ func validatePprofProfile(p *pprofproto.Profile) error {
 		if len(s.Value) != sampleLen {
 			return fmt.Errorf("mismatch: sample has %d values vs. %d types", len(s.Value), len(p.SampleType))
 		}
-		if len(s.LocationId) == 0 {
-			return fmt.Errorf("sample %d has no location ids", i)
-		}
-		for j, l := range s.LocationId {
-			if l == 0 {
-				return fmt.Errorf("location ids of stacktraces must be non-zero")
-			}
-			if l > locationsNum {
-				return fmt.Errorf("sample %d location number %d (%d) is out of range", i, j, l)
+		if len(s.LocationId) > 0 {
+			for j, l := range s.LocationId {
+				if l == 0 {
+					return fmt.Errorf("location ids of stacktraces must be non-zero")
+				}
+				if l > locationsNum {
+					return fmt.Errorf("sample %d location number %d (%d) is out of range", i, j, l)
+				}
 			}
 		}
 		for j, label := range s.Label {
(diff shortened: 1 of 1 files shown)


pkg/parcacol/ingest.go
```diff
@@ -189,15 +189,14 @@ func validatePprofProfile(p *pprofproto.Profile) error {
...
-		if len(s.LocationId) == 0 {
-			return fmt.Errorf("sample %d has no location ids", i)
-		}
-		for j, l := range s.LocationId {
-			if l == 0 {
-				return fmt.Errorf("location ids of stacktraces must be non-zero")
-			}
-			if l > locationsNum {
-				return fmt.Errorf("sample %d location number %d (%d) is out of range", i, j, l)
+		if len(s.LocationId) > 0 {
```
brancz, line 192: This should be a no-op now, We can just remove the if as the for loop will only have an effect if there are location IDs.

brancz (commented):
I think we might need to double-check that the stack trace keys we generate still work with this (or that there is a special case for them where there are no locations).

brancz:
All good, it appears we already have a special case for the thing I was talking about: https://github.com/parca-dev/parca/blob/185dce38cc80ebb713863f803c91713592ed0f42/pkg/metastore/kv.go#L252

So after removing that `if` statement, this is ready to merge! Thank you so much for the contribution!!

pkg/parcacol/ingest.go
jnsgruk: Doh :man_facepalming:

javierhonduco:
Thanks for working on this!! :)

manojVivek:
@all-contributors please add @jnsgruk for code.

PR #1455, opened 2022-07-29
badger: Retry large txn as multiple txns

### LINKED ISSUE #1397: Transaction too big errors
Hi, I started running v0.12.0 and instantly noticed the following errors which get repeated indefinitely for the `/allocs` endpoint for a thaos-query scrape target.

Note there are no similar issues for the other scrape targets. 

```bash
level=error name=parca ts=2022-07-20T16:35:00.852555555Z caller=scrape.go:500 scrape_pool=rhobs target=http://10.128.2.183:9090/debug/pprof/allocs msg="WriteRaw failed for scraped profile" err="rpc error: code = Internal desc = failed to ingest profile: normalize stacktraces: Txn is too big to fit into one request"

level=error name=parca ts=2022-07-20T16:35:36.216191691Z caller=scrape.go:500 scrape_pool=rhobs target=http://10.131.4.81:9090/debug/pprof/allocs msg="WriteRaw failed for scraped profile" err="rpc error: code = Internal desc = failed to ingest profile: normalize stacktraces: Txn is too big to fit into one request
```

Source https://github.com/dgraph-io/badger/blob/69926151f6532f2fe97a9b11ee9281519c8ec5e6/errors.go#L39

PhilipGough:
This change takes care of ErrTxnTooBig from badger.

ErrTxnTooBig will be reported in case the number of
pending writes/deletes in the transaction exceeds a certain limit.

Fixes #1397 based on the info and discussion in 
1. https://github.com/dgraph-io/badger/blob/7879d23752c1e34564bc5db4e0a1d598b4473df0/docs/content/get-started/index.md#L141
2. https://github.com/dgraph-io/badger/issues/1242

I've built and deployed to our testing cluster and behaviour looks ok.
Based on the logging, it looks like (in our environment at least),
writes are maxing out with a count of 3933.

I can also verify on the logs that there isn't more than a single retry.
Again this is based on our env so may that needs to be tweaked or configurable
but I wasn't convinced we should retry forever

### DIFF
--- a/pkg/metastore/badger.go
+++ b/pkg/metastore/badger.go
@@ -30,6 +30,7 @@ import (
 // store.
 type BadgerMetastore struct {
 	tracer trace.Tracer
+	logger log.Logger
 
 	db *badger.DB
 
@@ -69,6 +70,7 @@ func NewBadgerMetastore(
 	return &BadgerMetastore{
 		db:     db,
 		tracer: tracer,
+		logger: logger,
 	}
 }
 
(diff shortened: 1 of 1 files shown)


pkg/metastore/badger.go
```diff
@@ -481,6 +485,39 @@ func (m *BadgerMetastore) GetOrCreateStacktraces(ctx context.Context, r *pb.GetO
 		stacktraceKeys = append(stacktraceKeys, MakeStacktraceKey(stacktrace))
 	}
 
+	const maxRetries = 2
```
PhilipGough, line 488: not sure about this setting

pkg/metastore/badger.go
```diff
@@ -481,6 +485,39 @@ func (m *BadgerMetastore) GetOrCreateStacktraces(ctx context.Context, r *pb.GetO
...
+		if len(result.retryWith) == 0 {
+			break
+		}
+		stacktraceKeys = result.retryWith
+		m.logger.Debugf("retrying GetOrCreateStacktraces with %d stacktraces", len(stacktraceKeys))
+	}
+
+	if len(result.retryWith) != 0 {
+		m.logger.Warningf("failed to GetOrCreateStacktraces %d stacktraces", len(stacktraceKeys))
+		return res, fmt.Errorf("partial commit of stacktraces: %w", badger.ErrTxnTooBig)
```
PhilipGough, line 509: I'm going to wrap it in the end because we can infer the error and I didn't want to lose the info in the log that originally highlighted it

pkg/metastore/badger.go
brancz: I think this is ok, if people frequently run into problems we can make this a flag ultimately, but the less configuration is available the better.

pkg/metastore/badger.go
```diff
@@ -31,6 +31,7 @@ import (
 // store.
 type BadgerMetastore struct {
 	tracer trace.Tracer
+	logger *BadgerLogger
```
brancz, line 34: The `BadgerLogger` is just used to comply to Badger's logging interface, we use go-kit logger everywhere.

brancz (approved):
The strategy looks sound to me, once we fix the logger I'm happy to merge this! Thank you so much for this very important contribution!

brancz:
Thank you so much for the amazing contribution! Looks good to me!

manojVivek:
@all-contributors Please add @PhilipGough for code.

PR #1457, opened 2022-07-29
Add toleration for deployment pod spec

kakkoyun:


brancz:
I don’t think this universally makes sense. Most likely people don’t want to run their parca server on any node regardless of the taint. I think we should leave the tolerations empty for Parca. With parca agent on the other hand we want to try to have it run on all nodes no matter the circumstances, so we tolerate all taints.

Issue #1448, opened 2022-07-29
Metastore key/value distribution


brancz:
I just did a small analysis of a Parca server's metastore that was running on a relatively small Kubernetes cluster.

| Type | Value |
| --- | --- | 
| Stacktraces | 3 209 744 |
| Locations | 2 268 828 |
| LocationLines | 1 972 027 |
| Unsymbolized | 296 791 |
| Functions | 41 655 |
| Mappings | 144 |
| All | 7 789 189 |

At ~8GB database size this means each object is on average ~1.027kb.

I keep going back and forth in my head whether there is anything to change here, but I think the numbers and their distribution were interesting so I wanted to share.

kakkoyun:
> After a quick analysis, it appears that 1 373 068 of those locations don't even have mappings. This makes me think that Parca Server should probably replace these locations with a single generic "unknown" location.

👍 👍 👍  Totally. Without a mapping, we can't do anything for those.

brancz:
Based on this data, I'm also going to have a look at denormalizing `LocationLines`, and instead of having a separate object for it, I'm going to include them in `Location` objects directly.

PR #1471, opened 2022-08-01
*: Denormalize location lines

brancz:
This patch moves location lines directly into the locations objects,
removing the need to store equal amounts of keys in location lines as in
locations and consequently removing the need to query double the amount
of keys on read requests.

The proto-pr job will fail as this changes the API, which is ok since the API is still in alpha.

This will shave off another ~25% of keys as demonstrated in https://github.com/parca-dev/parca/issues/1448 (meaning less keys to manage for badger, meaning lower memory/disk usage by badger).


kakkoyun (approved):
LGTM :heart:

PR #1480, opened 2022-08-03
*: Allow paginating unsymbolized locations

brancz:
cc @kakkoyun


brancz:
I just wrote an additional test for this and it turns out it's not 100% working correctly yet.

kakkoyun (approved):
LGTM :+1:

Issue #1470, opened 2022-08-01
Generate callgraph report - node / edge list 


monicawoj:
This issue is related to the UI issue of creating a callgraph visualization: #304 

In order to render a callgraph visualization in the UI, we will need the backend to format data as a list of nodes and edges, as such:

```
Node {
  data: {
    id,
    ...any additional info related to the method called
  }
}

Edge {
  data: {
    id,
    source, // id of a node
    target, // id of a node
    ...any additional info related to the edges between nodes
  }
}

{ data: {nodes: Node[], edges: Edge[]} }
```
This data will help inform us as to which layout algorithm will be optimal for rendering the callgraph, so it would ideally be complete before the frontend is ready to merge. Once the layout algorithm informed by real data has been chosen, both PRs (backend and frontend) can be merged.

manojVivek:
- [ ] To reduce the number of nodes, filter out nodes and edges that have weight `0.01`% of the total weight.

PR #1487, opened 2022-08-04
api: Basic callgraph implementation

manojVivek:
Implements #1470


ui/packages/shared/client/src/parca/query/v1alpha1/query.ts
```diff
@@ -602,6 +608,75 @@ export interface FlamegraphNodeMeta {
      */
     line?: Line;
 }
+/**
+ * CallgraphNode represents a node in the graph
+ *
+ * @generated from protobuf message parca.query.v1alpha1.CallgraphNode
+ */
+export interface CallgraphNode {
```
monicawoj, line 616: Do we plan to have some other information for each node in addition to the name and id? At a minimum, it'd be helpful to have the information required by the flamegraph tooltip so we could potentially reuse the contents of that tooltip component. For example, a "cumulative"/total value, the sample unit, and meta info like the function name and location address?

proto/parca/query/v1alpha1/query.proto
```diff
@@ -331,6 +334,39 @@ message FlamegraphNodeMeta {
...
+  string id = 1;
+
+  // source represents the id of the source node
+  string source = 2;
+
+  // target represents the id of the target node
+  string target = 3;
+
+  // visits is the number of visits to the edge
+  int64 visits = 4;
```
brancz, line 358: it's not about the number of visits, what we want to understand here is the amount of cumulative that this edge leads to

proto/parca/query/v1alpha1/query.proto
```diff
@@ -331,6 +334,39 @@ message FlamegraphNodeMeta {
   parca.metastore.v1alpha1.Line line = 4;
 }
 
+// CallgraphNode represents a node in the graph
+message CallgraphNode {
+  // id is the unique id of the node
+  string id = 1;
+
+  // name is the name of the node
+  string name = 2;
```
brancz, line 343: We need to keep track of the cumulative and flat values in each node.

Also, I would not just have a flat `name`. We should have all the metadata associated with this node, just like the flamegraph nodes do (exactly what Monica said).

brancz (commented):
tl;dr 100% agree with Monica

ui/packages/shared/client/src/parca/query/v1alpha1/query.ts
manojVivek: Done, added this.

pkg/query/callgraph_test.go
```diff
@@ -145,3 +147,68 @@ func TestGenerateCallgraph(t *testing.T) {
...
+			{Id: "5", Source: "B", Target: "F", Cumulative: 1},
+			{Id: "6", Source: "F", Target: "H", Cumulative: 1},
+			{Id: "7", Source: "F", Target: "G", Cumulative: 1},
+			{Id: "8", Source: "H", Target: "I", Cumulative: 1},
+			{Id: "9", Source: "E", Target: "I", Cumulative: 1},
+			{Id: "10", Source: "I", Target: "J", Cumulative: 1},
+		},
+		Cumulative: 1000,
+	}
+	defer func() { //catch or finally
```
manojVivek: Please ignore this, I have removed this in a follow-up commit.

pkg/query/callgraph.go
```diff
@@ -165,3 +170,114 @@ func lineToGraphNode(
...
+	for _, node := range graph.Nodes {
+		if nodesToRemove[node.Id] {
+			// patch the edges from its parent to child nodes
+			parentNodeId, cummValue, incomingEdgesToRemove := findAValidParent(node, incomingEdges, outgoingEdges, nodesToRemove)
+			for _, edge := range outgoingEdges[node.Id] {
+				if nodesToRemove[edge.Target] {
+					// Skipping the edge creation as this will be patched by the downstream node that is being removed.
+					continue
+				}
+				newEdge := &querypb.CallgraphEdge{Id: uuid.New().String(), Source: parentNodeId, Target: edge.Target, Cumulative: cummValue}
```
brancz, line 246: In pprof the edge seems to know when nodes were skipped and draws a dotted line (I think this is useful). To do that I think we need some form of indicator on the edge that this happened, no?

brancz (commented):
Strategy looks good!

pkg/query/callgraph.go
manojVivek: Yes, that's a good point. Will add a field `isCollapsed` and set it to `true` on these edges?

manojVivek: Done, implemented it.

PR #1495, opened 2022-08-04
ci: set up pre-commit

maxbrunet:
I went to setup a workflow for shellcheck/hadolint and decided to use [`pre-commit`](https://pre-commit.com/) to simplify the task of file matching, and end up enabling a bunch of them:

* File fixers (end-of-file-fixer, trailing-whitespace)
* insert license
* buf
* golangci-lint (takes care of gofumpt too)
* hadolint
* shellcheck
* shfmt
* jsonnetfmt
* prettier

`husky` is removed, since both pre-commit frameworks cannot be used simultaneously. I have also tried to set up `eslint`, but it is broken, and once fixed, it raises a lot of errors/warnings. I'll create a separate PR to fix it at least.

I have updated `CONTRIBUTING.md` to mention `pre-commit` (already in main branch :sweat_smile:). And of course, all files have been fixed, so the pipeline passes.

Related to parca-dev/parca-agent#466


.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,50 @@
...
+          key: pre-commit-${{ env.PY_HASH }}-${{ hashFiles('.pre-commit-config.yaml') }}
+
+      - name: Run pre-commit
+        run: |
+          declare -a EXTRA_ARGS=()
+          if [[ "${GITHUB_EVENT_NAME}" == 'pull_request' ]]; then
+            EXTRA_ARGS=(--from-ref "origin/${GITHUB_BASE_REF}" --to-ref "${GITHUB_SHA}")
+          fi
+
+          pre-commit run --show-diff-on-failure --color=always "${EXTRA_ARGS[@]}"
```
kakkoyun, line 50: TIL about pre-commit!

kakkoyun (approved):
LGTM. More checks!

I'll leave others a chance to have a look at it.

kakkoyun:
@maxbrunet, I think we should have separated the golangci-lint raised issues separately. Or similar formatting changes in the same fashion. We need those changes now :sweat_smile: 

`pre-commit` changes should wait for other maintainers' consensus.

kakkoyun:
@maxbrunet Let me know if you want to split it. Otherwise, I'll send something to fix the CI, and you can rebase this one.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,65 @@
...
+          - --case-indent
+          - --binary-next-line
+  - repo: https://github.com/google/go-jsonnet
+    rev: v0.18.0
+    hooks:
+      - id: jsonnet-format
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v2.7.1
+    hooks:
+      - id: prettier
```
manojVivek, line 64: Instead of moving prettier as a hook on the root, is it possible to run a pre-commit command(`yarn pre-commit`) for the `ui` folder and let it take care of its pre-commit hooks? 

This way, we can keep the responsibilities separated without moving the UI config files into the root dir, etc. 

Just checking if this is possibility otherwise, I'm ok with the current changes.

manojVivek, line 64: Instead of moving prettier as a hook on the root, is it possible to run a pre-commit command(`yarn pre-commit`) for the `ui` folder and let it take care of its pre-commit hooks? 

This way, we can keep the responsibilities separated without moving the UI config files into the root dir, etc. 

Just checking if this is possibility otherwise, I'm ok with the current changes.

manojVivek (commented):
Just a question, looks good otherwise.

metalmatze:
This looks like a really nice improvement! :tada: 
How long does it take to run all these tasks in total? Often times I commit like 5 things at a time and then if waiting 30s each time this would be too long. Not sure if that's really the case (haven't tried this locally)
Maybe we can separate some of these tasks to only run as a `pre-push` hook too? Maybe that gives more flexibility?

.pre-commit-config.yaml
maxbrunet: Technically everything is possible, but it would be unclean and complex. Prettier is a good tool and running across the whole repo gives some consistency. I am actually tempted to enable it for JSON/YAML to make things extra pretty later on :sparkles:

maxbrunet:
> This looks like a really nice improvement! :tada:
> How long does it take to run all these tasks in total? Often times I commit like 5 things at a time and then if waiting 30s each time this would be too long. Not sure if that's really the case (haven't tried this locally)
> Maybe we can separate some of these tasks to only run as a `pre-push` hook too? Maybe that gives more flexibility?

The initial setup on your 1st commit can take some time as it has too pull all the hooks and build their individual environments (Python virtualenvs, node modules, Go binaries, Docker images), but once this is done, only what is needed by your changes will run, so it is fairly fast

.pre-commit-config.yaml
manojVivek: Cool, thanks!

manojVivek (approved):

metalmatze:
Sounds good. Thanks @maxbrunet I'll try it for now.

maxbrunet:
Do not hesitate to let me know if anybody faces issues with the hooks :+1:

@kakkoyun could you mark the check as required, please? :pray:

PR #1510, opened 2022-08-05
ci: fix snapcraft upload

jnsgruk:
The gift that keeps on giving, I'm not sure why these positional arguments aren't working, I'll take that up separately with the Starcraft team over here.


jnsgruk:
Just gonna check this is really gonna work before marking as ready.

kakkoyun (approved):
LGTM 

:crossed_fingers: I hope GH Action will think the same :)

PR #1514, opened 2022-08-06
ci: cancel previous in-progress workflows

maxbrunet:
This should unjam GitHub Actions when a lot of PRs/branches keep being updated

https://docs.github.com/en/actions/using-jobs/using-concurrency

It also adds a timeout to all jobs to ensure they are cancelled if stuck.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -10,6 +10,10 @@ on:
     - main
     - release-*
 
+concurrency:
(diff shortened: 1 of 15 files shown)


kakkoyun (approved):
LGTM.

I didn't know this did exist. Amazing. Even cancelling the same flow will be an improvement.

Let's start with these timeout values. I believe we can more strict.

manojVivek:
@all-contributors Please add @maxbrunet for infra.

PR #1532, opened 2022-08-10
chore: add license header to TS files

maxbrunet:
Because it is that easy. If it is not desired, please simply close the PR :grin:


kakkoyun (approved):
LGTM

PR #1535, opened 2022-08-10
ci: test minimum Go version

maxbrunet:
Something I mentioned in https://github.com/parca-dev/.github/pull/4#issuecomment-1204483407. This would ensure it still works as advertised and if not, we would know it needs a bump:

https://github.com/parca-dev/parca/blob/172f9173751f6769fe8ac8855bf3dc4fb6ff0947/go.mod#L3

Note: If merged, the required checks will need to be updated, because their names is changed

(I start to wonder how many checks is too many checks :sweat_smile:)


kakkoyun (commented):
If we want to go down this path, I think we need to test against all the major versions between the latest one we use in the CI pipeline and the minimum required version.

On top of that, I'm concerned about how much time this takes. Testing on ARM already takes too much time. Maybe we should only do this when we merge something on main, but I'm not sure. 

We need to find the right balance of testing for all possible things and the amount of time it takes. We should keep the PR checks as fast as possible.

maxbrunet:
Sweet, it's dynamic :slightly_smiling_face:

maxbrunet:
Closing, this was an experiment, I do not think we will use it for now

PR #1540, opened 2022-08-11
Shutdown persistence

thorfour:
This allows you to shutdown Parca, and on restarts query the historical data. Can see below where I shutdown Parca and started it much later
image

 

- updates version of frostdb to support persistence
- enables a graceful shutdown of the database on exit to write blocks to bucket storage
- refactors the persistence related flags to be less confusing:

       1. To enable persistence you only need to provide the `--enable-persistence` flag
       2. `--storage-persist` and `--storage-in-memory` has been removed.
       3. To enable the WAL it is now `--storage-enable-wal`
       4.  The `--metastore` flag now only accepts `badger` and the above persistence flags will determine if it's in-memory or not

### DIFF
--- a/pkg/parca/parca.go
+++ b/pkg/parca/parca.go
@@ -433,6 +433,11 @@ func Run(ctx context.Context, logger log.Logger, reg *prometheus.Registry, flags
 			if err != nil && !errors.Is(err, context.Canceled) {
 				level.Error(logger).Log("msg", "error shutting down server", "err", err)
 			}
+
+			// Close the columnstore after the parcaserver has shutdown to ensure no more writes occur against it.
+			if err := col.Close(); err != nil {
+				level.Error(logger).Log("msg", "error closing columnstore", "err", err)
+			}
 		},
 	)
 	if err := gr.Run(); err != nil {
--- a/pkg/parca/parca.go
+++ b/pkg/parca/parca.go
@@ -70,9 +70,9 @@ import (
 )
 
 const (
-	symbolizationInterval   = 10 * time.Second
-	flagModeScraperOnly     = "scraper-only"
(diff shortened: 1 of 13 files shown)


brancz (approved):
Love it!

pkg/parca/parca.go
```diff
@@ -433,6 +432,11 @@ func Run(ctx context.Context, logger log.Logger, reg *prometheus.Registry, flags
 			if err != nil && !errors.Is(err, context.Canceled) {
 				level.Error(logger).Log("msg", "error shutting down server", "err", err)
 			}
+
+			// Close the columnstore after the parcaserver has shutdown to ensure no more writes occur against it.
+			if err := col.Close(); err != nil {
```
metalmatze, line 437: So this is the crux of this PR, is it? Was looking for a bit to find it. The close function itself is defined in FrostDB? 
Very nice to tie it into the run.Group shutdown too. :+1:

metalmatze (approved):

Sylfrena (approved):

maxbrunet (commented):
The flags needs to be updated for the snap package:

https://github.com/parca-dev/parca/blob/b36f7d17f4de85ead64d46808e1be0bc48b4e698/snap/hooks/configure#L32-L35

https://github.com/parca-dev/parca/blob/b36f7d17f4de85ead64d46808e1be0bc48b4e698/snap/parca-wrapper#L24

https://github.com/parca-dev/parca/blob/b36f7d17f4de85ead64d46808e1be0bc48b4e698/snap/parca-wrapper#L34-L47

https://github.com/parca-dev/parca/blob/b36f7d17f4de85ead64d46808e1be0bc48b4e698/snap/README.md?plain=1#L42

PR #1575, opened 2022-08-18
ui: Fix flickering on Parca UI

yomete:
This fixes the flickering that happens on the Parca UI and also contains another fix for a bug that happens when you try to use the query browser.


javierhonduco (approved):
🥳

PR #1600, opened 2022-08-24
ui: Fixed the unnecessary re-render of MetrixGraph when a timestamp is selected

manojVivek:
This re-render was happening as we were creating the `queryClient` inside the render block of `Profiles` component. As a result, it was creating a new client for every component render, triggering the effect to fetch the metrics graph data every time. 
Fixed it by creating the `queryClient` globally once. 

Before:
https://user-images.githubusercontent.com/1283424/186392190-46a28fef-65a0-4e83-9364-58375d959d95.mov

After:
https://user-images.githubusercontent.com/1283424/186392225-93f60b6e-9096-4dc3-ae57-ef6c6461c91d.mov


yomete (approved):
LGTM! 🚢 🇮🇹

PR #1609, opened 2022-08-25
pkg/debuginfo: Add more tracing for debuginfo requests

brancz:


yomete:
@brancz not sure why the precommit step is failing but i think it can be merged after a review.

monicawoj (approved):
Thanks for making these changes, Yomi. Looks good to me, though we'll want to return to the "eslint-disable"d files at a later time.

Issue #1572, opened 2022-08-17
Don't allow store-address flag when not in scraper only mode


thorfour:
If you set the store-address flag you're likely expecting it to start shipping data to the remote storage. However if you don't set the mode to be scraper only it doesn't behave in that expected way.

We should have the Parca exit with an error suggesting that if you set `--store-address` that you should also set `--mode=scraper-only`

boranx:
hey folks, I've recently started hacking on `parca` and like it so far! thanks for bringing these together
Although I'm not very confident in its code, I wanted to propose a small pr to fix this issue

PR #1611, opened 2022-08-25
bug: exit when store-address is set and the mode is not scraper-only

boranx:
Fix: https://github.com/parca-dev/parca/issues/1572

### DIFF
--- a/pkg/parca/parca.go
+++ b/pkg/parca/parca.go
@@ -143,6 +143,10 @@ func Run(ctx context.Context, logger log.Logger, reg *prometheus.Registry, flags
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -15,3 +15,6 @@ TODO.md
 pkg/columnstore/testdata/merge-bench
 
 data/
+
+# IDE
+.vscode/
```
metalmatze, line 20: Please add this to a global gitignore on your system rather than to this repository. 
https://stackoverflow.com/questions/7335420/global-git-ignore

boranx: ack, this is excluded now

thorfour (approved):
Yay! Thank you for contributing!

manojVivek:
@all-contributors Please add @boranx for code.

Issue #1579, opened 2022-08-18
ui: compile TypeScript in strict mode


maxbrunet:
https://www.typescriptlang.org/tsconfig#strict

PR #1613, opened 2022-08-25
Fix eslint errors

### LINKED ISSUE #1610: ui: Fix eslint issues in ignored files
- [ ] ui/packages/shared/components/src/Callgraph/mockData/index.ts
- [ ] ui/packages/shared/components/src/Callgraph/index.tsx
- [ ] ui/packages/shared/components/src/Callgraph/utils.ts
- [ ] ui/packages/shared/components/src/GraphTooltip/index.tsx
- [ ] ui/packages/shared/functions/src/useUIFeatureFlag/index.ts
- [ ] ui/packages/shared/profile/src/ProfileView.tsx

monicawoj:
Resolves #1610 (UI linter errors)


yomete (approved):
LGTM! :shipit:

maxbrunet (commented):
Pre-commit can be run locally with:

```
pre-commit run --all-files
```

See `--help` to run a subset of hooks/files and iterate faster

Installation instructions are here if needed: [pre-commit.com/#installation](https://pre-commit.com/#installation) (it's fairly popular, so it should be available with most system package managers)

ui/packages/shared/functions/src/useUIFeatureFlag/index.ts
```diff
@@ -15,36 +15,39 @@ import useLocalStorageState from 'use-local-storage-state';
...
   window.localStorage.setItem(UI_FLAGS, JSON.stringify(flags));
 };
 
 initializeFlagsFromURL();
 
-const useUIFeatureFlag = (featureFlag: string, defaultValue: boolean = false) => {
+const useUIFeatureFlag = (
+  featureFlag: string,
+  defaultValue: boolean = false
+): Array<boolean | ((key: string) => boolean)> => {
```
maxbrunet, line 44: This return type does not look right to me, maybe:

```suggestion
): [boolean, ((flag: boolean) => void)] => {
```

ui/packages/shared/profile/src/ProfileView.tsx
```diff
@@ -219,7 +219,7 @@ export const ProfileView = ({
                   </Button>
                 </div>
 
-                {callgraphEnabled ? (
+                {(callgraphEnabled as boolean) ? (
```
maxbrunet, line 222: `as boolean` should not be needed if the return type of the function is correct
```suggestion
                {callgraphEnabled ? (
```

ui/.eslintrc.yml
```diff
@@ -39,3 +39,4 @@ rules:
     - singleline:
         delimiter: 'semi'
         requireLast: false
+  '@typescript-eslint/no-dynamic-delete': 0
```
maxbrunet, line 42: Not fixing these? Maybe only disable where it makes sense instead of globally?

https://github.com/typescript-eslint/typescript-eslint/blob/HEAD/packages/eslint-plugin/docs/rules/no-dynamic-delete.md

ui/packages/shared/components/src/Callgraph/index.tsx
```diff
@@ -87,7 +87,7 @@ const Callgraph = ({graph, sampleUnit, width}: Props): JSX.Element => {
               const targetNode = nodes.find(n => n.id === edge.target);
               return (
                 <Edge
-                  key={`edge-${edge.source}-${edge.target}`}
+                  key={`edge-${edge.source as string}-${edge.target as string}`}
```
maxbrunet, line 90: `edge` should be typed earlier, then casting will not be needed. It will make more sense and will have to be done anyway once `noImplicitAny` is enabled (#1579)

ui/packages/shared/components/src/Callgraph/index.tsx
```diff
@@ -99,7 +99,7 @@ const Callgraph = ({graph, sampleUnit, width}: Props): JSX.Element => {
             })}
             {nodes.map(node => (
               <Node
-                key={`node-${node.data.id}`}
+                key={`node-${node.data.id as string}`}
```
maxbrunet, line 102: ditto, `nodes` should be typed earlier (in the `map()` call)

ui/packages/shared/components/src/Callgraph/mockData/index.ts
```diff
@@ -435,7 +435,7 @@ export const jsonGraphWithGraphvizPositions = {
   ],
 };
 
-export const node_with_meta_data = (id, name) => ({
+export const nodeWithMetaData = (id, name): {id: string; cumulative: number; meta: any} => ({
```
maxbrunet, line 438: `id` and `name` need to be typed to avoid casting

maxbrunet (changes requested):
As the linter and compiler get tighten, these issues should get easier to identify at their root

ui/packages/shared/functions/src/useUIFeatureFlag/index.ts
monicawoj: oh, good catch!

monicawoj: oh, good catch!

ui/packages/shared/components/src/Callgraph/index.tsx
```diff
@@ -34,13 +33,13 @@ const Callgraph = ({graph, sampleUnit, width}: Props): JSX.Element => {
...
     }
-  }, [width]);
+  }, [graph, width]);
 
   // 3. Render the graph with calculated layout in Canvas container
-  if (!width || !graphData) return <></>;
+  if (width == null || graphData == null) return <></>;
 
   const height = width;
   const {objects, edges: gvizEdges, bb: boundingBox} = JSON.parse(graphData);
```
maxbrunet, line 45: `edge` can be typed earlier. I think this is the earliest it can be (with these other variables). Then TypeScript should be able to figure out the types of the derived variables from `edges` being `IEdge[]`

maxbrunet (commented):
Looking good, thank you! I think it may be possible to type the `edges` and `nodes` variables even earlier, but it'll be easier to look into it in the strict mode PR.

Only `pre-commit` is left, it should be as quick and easy as `pre-commit run --all-files`, then committing the changes (but please do not hesitate to let me know if there are any blockers)

monicawoj:
> pre-commit run --all-files

@maxbrunet when I attempt to run pre-commit against all the files, I get this error: 
```An unexpected error has occurred: CalledProcessError: command: ('/opt/homebrew/bin/go', 'install', './...')
return code: 2``` (please see full log attached)

Any thoughts on how to resolve this?

[pre-commit.txt](https://github.com/parca-dev/parca/files/9496080/pre-commit.txt)

**Comment by maxbrunet** [comment]
> Any thoughts on how to resolve this? Fwiw, if I comment out the jsonnet-format hook, everything runs fine, so it appears to be due to that.

Looks like Jsonnet 0.18 depends on an old version of `golang.org/x/sys` which does not work with Go 1.18+ and darwin arm64. The fix should land in the next `go-jsonnet` release.

In the meantime, you can use the `SKIP=jsonnet-format` environment variable with `pre-commit run` and `git commit`

**Review by maxbrunet — COMMENTED** [review]
I think you can add the `CHANGELOG.md` files to `.prettierignore`, since they are auto-generated, they'll keep coming back

```patch
diff --git a/.prettierignore b/.prettierignore
index e5c208e8..9e9695cf 100644
--- a/.prettierignore
+++ b/.prettierignore
@@ -7,3 +7,6 @@ coverage/
 
 # Exclude protobuf generated files
 ui/packages/shared/client
+
+# Exclude generated changelogs
+ui/**/CHANGELOG.md
```

maxbrunet (approved):
Looking all tidy! :sparkles::broom:

A final tip for EOF and trailing whitespaces, you can install the [editorconfig](https://editorconfig.org/) plugin for your IDE and the config [here](https://github.com/parca-dev/parca/blob/main/.editorconfig) should take care of them before `pre-commit` even needs to run

manojVivek (approved):

PR #1615, opened 2022-08-25
pkg/debuginfo: Mark already existing metadata as uploaded

brancz:
There can be confusing circumstances, when for example debuginfos have already been uploaded in the past, but there was no metadata. Then the metadata will continue to not be created, while the debuginfo server acknowledges that it already knows about the debuginfos.

This ensures that if the debuginfo server realizes that it already has debuginfos, that it marks them as uploaded in the metadata.

### DIFF
--- a/pkg/debuginfo/store.go
+++ b/pkg/debuginfo/store.go
@@ -214,6 +214,10 @@ func (s *Store) upload(ctx context.Context, buildID, hash string, r io.Reader) e
 		if hash != "" && metadataFile != nil {
 			if metadataFile.Hash == hash {
 				level.Debug(s.logger).Log("msg", "debug info already exists", "buildid", buildID)
+				if err := s.metadata.MarkAsUploaded(ctx, buildID, hash); err != nil {
+					err = fmt.Errorf("failed to update metadata after uploaded: %w", err)
+					return status.Error(codes.Internal, err.Error())
+				}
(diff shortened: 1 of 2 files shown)


pkg/debuginfo/store.go
```diff
@@ -214,6 +214,10 @@ func (s *Store) upload(ctx context.Context, buildID, hash string, r io.Reader) e
 		if hash != "" && metadataFile != nil {
 			if metadataFile.Hash == hash {
 				level.Debug(s.logger).Log("msg", "debug info already exists", "buildid", buildID)
+				if err := s.metadata.MarkAsUploaded(ctx, buildID, hash); err != nil {
```
javierhonduco, line 217: > There can be confusing circumstances, when for example debuginfos have already been uploaded in the past, but there was no metadata. 

If the metadata was never created, won't this fail all the time?

https://github.com/parca-dev/parca/blob/41a8b3003d8cf411cd57a36ac2c8201456c36112/pkg/debuginfo/metadata.go#L144-L148

brancz: Good catch!

javierhonduco (approved):
LGTM!

kakkoyun (changes requested):
LGTM.

@javierhonduco's point still persists though.

brancz:
Long superseeded by lots of other work.

PR #1638, opened 2022-08-31
ci: fix version parsing for container builds

### LINKED ISSUE #1637: Incorrect version in container builds
Related to #542 and #537 - though it looks like the release process has changed since then.

The containers being published on merges to main are displaying the wrong version numbers.

For example, at the time of writing the latest image is: `main-390a0a60`

```
❯ docker run -p7070:7070 --rm ghcr.io/parca-dev/parca:main-390a0a60 /parca --log-level debug
...
level=debug name=parca ts=2022-08-31T14:19:30.379067196Z caller=main.go:49 msg="parca initialized" version=v0.0.1-390a0a6 commit=390a0a605790e0207d95a623f384b315ef15bf5e config="&{parca.yaml all debug :7070 []  false  0 0 false false 8196 536870912 data false simple 3 badger api.pprof.me:443 [https://debuginfod.elfutils.org] 5m0s /tmp    false false map[]}"
```

Note `version=v0.0.1-390a0a6` which is also displayed in the UI.

I took a look through the build process for the container in the workflow and couldn't see anything obviously wrong, the `GORELEASER_CURRENT_TAG`  variable looks the same as the other builds, but somehow something is getting lost.

jnsgruk:
Fixes #1637 

I think by not specifying the fetch depth, the `checkout` action was calling `git fetch --no-tags`

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -57,6 +57,8 @@ jobs:
     steps:
       - name: Check out code into the Go module directory
         uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # tag=v3.0.2
+        with:
+          fetch-depth: 0
 
       - name: Set up Go
         uses: actions/setup-go@268d8c0ca0432bb2cf416faae41297df9d262d7f # tag=v3.3.0


jnsgruk:
Ok so looking at the logs for this [run](https://github.com/parca-dev/parca/runs/8117037658?check_suite_focus=true#step:5:2) vs the [last commit on `main`](https://github.com/parca-dev/parca/runs/8114993419?check_suite_focus=true#step:5:7), I think this has solved it.

kakkoyun (approved):
LGTM.
Good catch and fix! Thanks :heart:

kakkoyun:
I wonder if we have similar issues with any other actions?

We definitely miss the same for the corresponding agent action.

Issue #1276, opened 2022-06-30
symbolizer: Add metrics of `symbolizer` package


kakkoyun:
- Success and failures
- Reasons
- How much does it take to finish a round?
- How much does it take to store the symbolized locations?

PR #1640, opened 2022-08-31
Count symbolizer attempts and errors

marselester:
Hi! I would like to tackle https://github.com/parca-dev/parca/issues/1276 issue, but I am not sure I got it right.

> Success and failures
> Reasons

I assumed that one should count symbolization attempts and pay attention to failure reasons.

> How much does it take to finish a round?

Did you mean a duration of `runSymbolizationCycle()` or something else?

> How much does it take to store the symbolized locations?

Did you mean a duration or a size of locations `s.metastore.CreateLocationLines()`?


pkg/symbolizer/symbolizer.go
```diff
@@ -33,6 +34,11 @@ import (
 
 type Symbolizer struct {
 	logger log.Logger
+	// attempts counts the total number of symbolization attempts.
+	attempts prometheus.Counter
+	// errors counts the total number of symbolization errors, partitioned by an error reason
+	// such as failure to fetch unsymbolized locations.
+	errors *prometheus.CounterVec
```
marselester, line 41: I used two counters, although one can do the job. Please let me know which option you prefer.

pkg/symbolizer/symbolizer.go
```diff
@@ -89,8 +114,10 @@ func (s *Symbolizer) runSymbolizationCycle(ctx context.Context) {
 			Limit:  s.batchSize,
 			MinKey: prevMaxKey,
 		})
+		s.attempts.Inc()
 		if err != nil {
 			level.Error(s.logger).Log("msg", "failed to fetch unsymbolized locations", "err", err)
+			s.errors.With(prometheus.Labels{"reason": "fetch_unsymbolized_locations"}).Inc()
```
marselester, line 120: If `fetch_unsymbolized_locations` is too verbose, I will try to come up with a shorter name.

marselester:
Hm, golangci-lint times out `Timeout exceeded: try increasing it by passing --timeout option`.

kakkoyun:
> Did you mean a duration of `runSymbolizationCycle()` or something else?

Yes, exactly this.

> Did you mean a duration or a size of locations `s.metastore.CreateLocationLines()`?

Ditto :+1:

pkg/symbolizer/symbolizer.go
kakkoyun: This looks perfect :+1:

pkg/symbolizer/symbolizer.go
```diff
@@ -89,8 +114,10 @@ func (s *Symbolizer) runSymbolizationCycle(ctx context.Context) {
 			Limit:  s.batchSize,
 			MinKey: prevMaxKey,
 		})
+		s.attempts.Inc()
```
kakkoyun, line 117: So the counters actually count per batch. This is worth mentioning in the description.

kakkoyun: We should also consider this when we add the histogram for the duration.

pkg/symbolizer/symbolizer.go
kakkoyun: I think it's fine :+1:

kakkoyun (commented):
LGTM. Awesome work! Thanks a lot for your contribution.

I've commented inline. They're just minor things.

marselester:
Thanks! I've added histograms as well.

pkg/symbolizer/symbolizer.go
```diff
@@ -55,22 +69,57 @@ type DebugInfoFetcher interface {
...
 	debuginfo DebugInfoFetcher,
 	symbolizer *symbol.Symbolizer,
 	debuginfodCacheDir string,
 	debuginfoCacheDir string,
 	batchSize uint32,
 ) *Symbolizer {
-	return &Symbolizer{
+	attemptsTotal := prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Name: "symbolizer_attempts_total",
```
kakkoyun, line 82:
```suggestion
			Name: "symbolizer_symbolication_attempts_total",
```

pkg/symbolizer/symbolizer.go
```diff
@@ -55,22 +69,57 @@ type DebugInfoFetcher interface {
...
+			Help: "Total number of symbolization attempts in batches.",
+		},
+	)
+	errorsTotal := prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "symbolizer_errors_total",
```
kakkoyun, line 88:
```suggestion
			Name: "symbolizer_symbolication_errors_total",
```

pkg/symbolizer/symbolizer.go
```diff
@@ -55,22 +69,57 @@ type DebugInfoFetcher interface {
...
+			Help: "Total number of symbolization errors in batches, partitioned by an error reason.",
+		},
+		[]string{"reason"},
+	)
+	duration := prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name:    "symbolizer_duration_seconds",
```
kakkoyun, line 95:
```suggestion
			Name:    "symbolizer_symbolication_duration_seconds",
```

pkg/symbolizer/symbolizer.go
```diff
@@ -55,22 +69,57 @@ type DebugInfoFetcher interface {
...
+	reg *prometheus.Registry,
 	metastore pb.MetastoreServiceClient,
 	debuginfo DebugInfoFetcher,
 	symbolizer *symbol.Symbolizer,
 	debuginfodCacheDir string,
 	debuginfoCacheDir string,
 	batchSize uint32,
 ) *Symbolizer {
-	return &Symbolizer{
+	attemptsTotal := prometheus.NewCounter(
```
kakkoyun, line 80:
```suggestion
	attemptsTotal := promauto.With(reg).NewCounter(
```

pkg/symbolizer/symbolizer.go
```diff
@@ -55,22 +69,57 @@ type DebugInfoFetcher interface {
...
+		prometheus.CounterOpts{
+			Name: "symbolizer_attempts_total",
+			Help: "Total number of symbolization attempts in batches.",
+		},
+	)
+	errorsTotal := prometheus.NewCounterVec(
```
kakkoyun, line 86:
```suggestion
	errorsTotal := promauto.With(reg).NewCounterVec(
```

pkg/symbolizer/symbolizer.go
```diff
@@ -55,22 +69,57 @@ type DebugInfoFetcher interface {
...
+		prometheus.CounterOpts{
+			Name: "symbolizer_errors_total",
+			Help: "Total number of symbolization errors in batches, partitioned by an error reason.",
+		},
+		[]string{"reason"},
+	)
+	duration := prometheus.NewHistogram(
```
kakkoyun, line 93:
```suggestion
	duration := promauto.With(reg).NewHistogram(
```

pkg/symbolizer/symbolizer.go
```diff
@@ -55,22 +69,57 @@ type DebugInfoFetcher interface {
...
+		prometheus.HistogramOpts{
+			Name:    "symbolizer_duration_seconds",
+			Help:    "How long it took in seconds to finish a round of the symbolization cycle in batches.",
+			Buckets: []float64{0.001, 0.01, 0.1, 0.3, 0.6, 1, 3, 6, 9, 20, 30, 60, 90, 120},
+		},
+	)
+	storeDuration := prometheus.NewHistogram(
```
kakkoyun, line 100:
```suggestion
	storeDuration := promauto.With(reg).NewHistogram(
```

pkg/symbolizer/symbolizer.go
```diff
@@ -55,22 +69,57 @@ type DebugInfoFetcher interface {
...
+		prometheus.HistogramOpts{
+			Name:    "symbolizer_store_duration_seconds",
+			Help:    "How long it took in seconds to store a batch of the symbolized locations.",
+			Buckets: []float64{0.001, 0.01, 0.1, 0.3, 0.6, 1, 3, 6, 9, 20, 30, 60, 90, 120},
+		},
+	)
+	reg.MustRegister(attemptsTotal, errorsTotal, duration, storeDuration)
```
kakkoyun, line 107: I would rather suggest using `promauto` package and registering in-line to be consistent with the rest of the code base.

kakkoyun (commented):
Awesome work!

I just realized we are missing the action name in the metric name `symbolizer ` is being subsystem name. (And I recently discovered `symbolication` is a more correct term).

And let's use `promauto` package to be consistent with the rest of the code base.

pkg/symbolizer/symbolizer.go
marselester: Good point, I have found a few examples in parca-agent repository.

marselester:
Good points! I also added `parca_` prefix to the metrics.

pkg/symbolizer/symbolizer.go
```diff
@@ -55,22 +70,64 @@ type DebugInfoFetcher interface {
...
 	debuginfo DebugInfoFetcher,
 	symbolizer *symbol.Symbolizer,
 	debuginfodCacheDir string,
 	debuginfoCacheDir string,
 	batchSize uint32,
 ) *Symbolizer {
-	return &Symbolizer{
+	attemptsTotal := promauto.With(reg).NewCounter(
+		prometheus.CounterOpts{
+			Namespace: "parca",
```
kakkoyun, line 83: I think we should omit `Namespace: "parca"`. This will result in the metric name being `parca_symbolizer_symbolication_attempts_total`. We don't use `parca` as a prefix elsewhere. 
Also, I don't mind collecting everything in the name. The same goes for metrics.

kakkoyun (approved):
Awesome :tada: 

Only one thing left and we can merge it.

pkg/symbolizer/symbolizer.go
marselester: Alright, that's done :)

kakkoyun (approved):

Issue #1192, opened 2022-06-24
FE: Allowed to query if no profile type is selected


lilic:
If I go to parca and select the label name that is suggested for me e.g. `app`, while leaving profile type as blank and I hit search, I get an empty screen. I was expecting a warning that I didn't select any profiles (as that was the case) or an error of some sort.

This is the URL that is generated: https://demo.parca.dev/?expression_a=%7B%7D&from_a=1656084616216&to_a=1656085516216&merge_a=false&time_selection_a=relative:minute|15&currentProfileView=icicle

PR #1645, opened 2022-09-01
ui: Improve querying experience

### LINKED ISSUE #1439: ui: Improve querying experience on Parca
- [x] Ensure that querying using the `MatchersInput` component is disabled until a profile type has been selected.
- [x] Add a loader to the MatchersInput component
- [x] Prevent overflow of the label names and values by adding a fixed height
- [x] Suggest similar label names and values when typing in the query browser. This means whenever a user types in a label name or value, the suggestion box will now be re-ordered to show the closest suggestion.
- [x] The `MatchersInput` component should show the next suggestion even when a user is manually typing in a query i.e. not clicking through the suggestions list.
- [ ] The backend expects the label value to be in quotes, so the `MatchersInput` component automagically adds a "" to surround the label value if omitted by the user.

yomete:
This PR addresses the following:

- [x] Ensures that querying using the `MatchersInput` component is disabled until a profile type has been selected. #1192 
- [x] Added a loader to the MatchersInput component for slow networks/huge data responses.
- [x] Prevent overflow of the label names and values (blocking the metrics graph and icicle graph) by adding a fixed height.
- [x] Suggest similar label names and values when typing in the query browser. This means whenever a user types in a label name or value, the suggestion box will now be re-ordered to show the closest similar suggestion.
- [x] The `MatchersInput` component now shows the next suggestion even when a user is manually typing in a query i.e. not clicking through the suggestions list.
- [x] The backend expects the label value to be in quotes, so the `MatchersInput` component automagically adds a "" to surround the label value if omitted by the user.

Resolves #1439


brancz:
Reeeeeally nice improvements!! So much better than before! I can still trigger a few problems, particularly when using backspace a lot, but I think we can take care of more as follow ups!

brancz (approved):

Issue #1461, opened 2022-07-30
Validate no duplicate labels on writes


brancz:
Our storage does not allow multiple labels with the same name, so we should validate and reject requests that have duplicate label names.

PR #1658, opened 2022-09-03
pkg/profilestore: reject requests that have duplicate label names

heylongdacoder:
Fixes: #1461


pkg/scrape/scrape.go
```diff
@@ -466,8 +465,6 @@ mainLoop:
 					Value: l.Value,
 				})
 			}
-			// Must ensure label-set is sorted
-			sort.Sort(tl)
```
heylongdacoder, line 470: since profilestore will sort the label-set, I think this can be removed?

brancz (approved):
Thank you! lgtm

PR #1632, opened 2022-08-30
fix(ui): type all the things (strict mode)

maxbrunet:
More type annotations!

Closes #1579

PR #1664, opened 2022-09-05
Use connect-go and connect-web for APIs

metalmatze:
This was my 20% Friday project and I'll continue.

It makes Parca use [Connect](https://connect.build/) which makes creating the Server for both HTTP/JSON and gRPC/protobuf a lot easier. 

The Targets page of Parca already works via the newly generated APIs with HTTP/JSON directly.
And some of the Query service APIs also work. This is the most work left, converting the APIs.

Since we use a couple of gRPC interceptors we need to write these for the time being ourselves. It should be too complex. In fact, for [Pyrra](https://github.com/pyrra-dev/pyrra) I've already started to write an interceptor that creates the known `grpc_server_*` and `grpc_client_*` metrics. 
AFAIK the buf people are working on an Open Telemetry interceptor too. I'll ask them about it.


ui/packages/shared/client/src/parca/query/v1alpha1/query_pb.ts
```diff
@@ -0,0 +1,2015 @@
...
+   * @generated from field: int64 flat = 3;
+   */
+  flat = protoInt64.zero;
+
+  /**
+   * diff is the diff value between two profiles
+   *
+   * @generated from field: int64 diff = 4;
+   */
+  diff = protoInt64.zero;
```
maxbrunet, line 889: That looks nice, it automatically handles the `int64` support: https://github.com/bufbuild/protobuf-es/blob/main/packages/protobuf/src/proto-int64.ts

The config was manual with the other plugin: https://github.com/timostamm/protobuf-ts/blob/master/MANUAL.md#bigint-support

The strict mode PR (#1632) surfaces some potential issues around it, and you might need to handle them, a large comment in the plugin's code seems to explain how

buf.gen.yaml
```diff
@@ -12,6 +12,11 @@ plugins:
     out: gen/proto/go
     opt: paths=source_relative
 
+  # renovate: datasource=github-releases depName=bufbuild/connect-go
+  - plugin: connect-go
```
maxbrunet, line 16: Aliases! I'd still pin the version of the plugins to ensure it is reproducible

maxbrunet (commented):
Exciting to see movement on this PR!


> Since we use a couple of gRPC interceptors we need to write these for the time being ourselves. It should be too complex. In fact, for [Pyrra](https://github.com/pyrra-dev/pyrra) I've already started to write an interceptor that creates the known `grpc_server_*` and `grpc_client_*` metrics. 
> AFAIK the buf people are working on an Open Telemetry interceptor too. I'll ask them about it.

btw `connect-opentelemetry-go` has had its first release: https://buf.build/blog/connect-opentelemetry-go

PR #1665, opened 2022-09-05
pkg/symbol: Prevent redundant ELF file openings

marselester:
Hello,

I was looking at the symbol package and noticed that there might be a potential performance gain if an ELF file was opened once instead of N times, i.e., there will be less [`elf.NewFile()`](https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/debug/elf/file.go;l=243) calls.

It's just an idea though. The downside is that the code gets less readable due to extra `f.Open()`/`f.Close()`.


pkg/debuginfo/store.go
```diff
@@ -223,7 +223,12 @@ func (s *Store) upload(ctx context.Context, buildID, hash string, r io.Reader) e
 			return status.Error(codes.Internal, err.Error())
 		}
 
-		if err := elfutils.ValidateFile(objFile); err != nil {
+		f, err := elfutils.Open(objFile)
+		if err == nil {
+			err = f.Validate()
```
kakkoyun, line 228: I'd expect to check for validity if there's no error opening the file.

pkg/debuginfo/store.go
```diff
@@ -223,7 +223,12 @@ func (s *Store) upload(ctx context.Context, buildID, hash string, r io.Reader) e
...
+		}
+		if err != nil {
+			f.Close()
```
kakkoyun, line 231: I'd also expect the file is opened only without an error returned from the Open function.

pkg/symbol/elfutils/elfutils.go
```diff
@@ -25,6 +25,32 @@ import (
 	"github.com/nanmu42/limitio"
 )
 
+// File represents an open ELF file.
+type File struct {
```
kakkoyun, line 29: Can we get away without introducing a new type? It'd be amazing just to use `elf.File`?

kakkoyun (commented):
Looking good. We might need to meditate on the APIs just a little more, maybe?

The idea was to have self-contained functions to work anywhere.
We can expose the equivalent functions that accept files as well. Or, as you did, we can add the functionality to a type. But rather than introducing a new type, how about using the existing one like `elf.File` or https://github.com/parca-dev/parca/blob/main/pkg/symbol/elfutils/debuginfofile.go

We also have a similar abstraction in parca-agent https://github.com/parca-dev/parca-agent/tree/main/pkg/objectfile that could give an idea.

pkg/symbol/elfutils/elfutils.go
marselester: Sure, I've updated the PR.

pkg/debuginfo/store.go
```diff
@@ -234,12 +235,19 @@ func (s *Store) upload(ctx context.Context, buildID, hash string, r io.Reader) e
 		}
 
 		// Valid.
-		hasDWARF, err := elfutils.HasDWARF(objFile)
+		f, err := elf.Open(objFile)
 		if err != nil {
 			level.Debug(s.logger).Log("msg", "failed to check for DWARF", "err", err)
```
kakkoyun, line 240:
```suggestion
			level.Debug(s.logger).Log("msg", "failed to open object file", "err", err)
```

pkg/debuginfo/store.go
```diff
@@ -433,18 +441,25 @@ func (s *Store) FetchDebugInfo(ctx context.Context, buildID string) (string, deb
 	}
 
 	if source != debuginfopb.DownloadInfo_SOURCE_DEBUGINFOD {
-		hasDWARF, err := elfutils.HasDWARF(objFile)
+		f, err := elf.Open(objFile)
 		if err != nil {
 			level.Debug(logger).Log("msg", "failed to check for DWARF", "err", err)
```
kakkoyun, line 446:
```suggestion
			level.Debug(logger).Log("msg", "failed to open object file", "err", err)
```

pkg/symbol/elfutils/elfutils.go
```diff
@@ -126,32 +114,9 @@ func IsSymbolizableGoObjFile(path string) (bool, error) {
 	return false, errors.New("failed to detect .gopclntab section or section has no bits")
 }
 
-// IsGoObjFile checks whether the specified executable or library file is generated by Go toolchain.
-func IsGoObjFile(path string) (bool, error) {
```
kakkoyun, line 130: I assume this has not been used anywhere else, right?

pkg/symbol/symbol.go
```diff
@@ -201,13 +205,21 @@ func (s *Symbolizer) liner(m *pb.Mapping, path string) (liner, error) {
 // newLiner creates a new liner for the given mapping and object file path.
 func (s *Symbolizer) newLiner(buildID, path string) (liner, error) {
 	logger := log.With(s.logger, "file", path, "buildid", buildID)
-	hasDWARF, err := elfutils.HasDWARF(path)
+
+	f, err := elf.Open(path)
+	if err != nil {
+		level.Error(logger).Log("msg", "failed to open binary", "err", err)
+		return nil, ErrLinerCreationFailedFile
```
kakkoyun, line 212: Let's just wrap and return the error we received. No need to obfuscate the existing error. Also, this deprecates the log line above. We should only handle errors once.

kakkoyun (commented):
Looking good. I've left my comments.

My other concern, even though not directly related to this PR, was that the `store.upload` function became harder to read with multiple levels of if and else's. We need to write some tests and refactor this method to make it more maintainable.

And also, we'll need a subsequent PR on the agent side to adapt to these changes.

pkg/symbol/elfutils/elfutils.go
marselester: Yeah, I couldn't find any references in parca and parca-agent repositories.

pkg/symbol/symbol.go
marselester: Makes sense 👍

kakkoyun (approved):
LGTM! Thanks :tada:

PR #1675, opened 2022-09-06
Run `buf mod update`

### LINKED ISSUE #1674: Update buf googleapis reference
We recently made a change to our managed [buf.build/googleapis/googleapis](https://buf.build/googleapis/googleapis) repository before certifying BSR as v1. You'll need to run `buf mod update` in proto/ to update your buf.lock.

[googleapis](https://github.com/googleapis/googleapis) contains over 3800 files, mostly relating to Google's core APIs. This causes numerous issues for most users, such as timeouts when installing packages with huge swaths of unused code to pull in just a few files. Not only is this a lot of code over the network and on disk, but it can cause issues in editors that try to parse all of that code!

However, there are only ~30 files used by 99.999% of developers, and these files are the most common dependency in the Protobuf ecosystem. For example:

- google.type, which defines useful messages such as DateTime and Money.
- google.rpc, for interacting with gRPC.
- google.api, for defining APIs, such as with grpc-gateway.

We considered these commonly used packages and included a subset of them into the new, slim googleapis repository.

We *hate* breaking anyone, even in an alpha, and our apologies that it's causing any difficulties here. The issues with googleapis were serious enough that we felt that we had to do this in this extraordinary circumstance. We're coming out of beta extremely soon, and once we do, we take our responsibility not to break customers (free users or otherwise) very seriously, we appreciate you riding the rough road for the moment!

stefanvanburen:
Fixes #1674.

### DIFF
--- a/.github/workflows/proto-pr.yaml
+++ b/.github/workflows/proto-pr.yaml
@@ -61,4 +61,4 @@ jobs:
         with:
           input: 'proto'
           # The 'main' branch of the GitHub repository that defines the module.
-          against: 'https://github.com/${GITHUB_REPOSITORY}.git#branch=main,ref=HEAD~1,subdir=proto'
+          against: 'https://github.com/${GITHUB_REPOSITORY}.git#branch=main,subdir=proto'
--- a/Makefile
+++ b/Makefile
@@ -109,6 +109,7 @@ proto/all: proto/vendor proto/format proto/lint proto/generate
 proto/lint:
 	# docker run --volume ${PWD}:/workspace --workdir /workspace bufbuild/buf lint
 	buf lint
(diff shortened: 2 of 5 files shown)


kakkoyun:
The failure of the PR check is because of the breaking API check. And that's caused by the removable of the referred module version from the buf registry. So it's safe to merge the PR if we're satisfied with the generated files. Consecutive checks will be green after we merge this.

```
❯ buf build .git#branch=main
Failure: buf.build/googleapis/googleapis:80720a488c9a414bb8d4a9f811084989: does not exist
```
https://buf.build/googleapis/googleapis/commits/main

kakkoyun (approved):

manojVivek:
@all-contributors Please add @stefanvanburen for code.

PR #1678, opened 2022-09-07
pkg/scrape: fix manager reload potential deadlock

heylongdacoder:


kakkoyun (approved):
LGTM.

Great catch! Thanks for the fix :+1:

PR #1686, opened 2022-09-08
ui: Exporting `Callgraph` as a dynamically imported component using React.Lazy

manojVivek:


yomete (approved):
Just the one TS error but apart from that, LGTM!

manojVivek:
> Just the one TS error but apart from that, LGTM!

Yes, just pushed a commit fixing that. 😅

PR #1688, opened 2022-09-08
Callgraph enhancements

monicawoj:
- applies different color scheme (tested for accessibility using color blindness simulator Chrome plugin "Spectrum")
- displays nodes as rectangles rather than circles
- displays function name on nodes that meet a certain value threshold
- added zoom and panning of the callgraph
- nodes are currently sized by cumulative value, but this will likely change to size by flat value once that data is available
- debounces node search to avoid multiple immediate re-renders of the callgraph/icicle graph


yomete (approved):
Amazing work! 👏🏾 👏🏾 👏🏾

PR #1691, opened 2022-09-08
pkg/symbol: Add docs

marselester:
Hi! I've added some docs for the symbol package. Not sure I got it right though.


kakkoyun (approved):
LGTM. Awesome! Thanks for this!

PR #1696, opened 2022-09-10
parca.yaml: replace object_storage with debug_info

s-urbaniak:
Currently, the example parca.yaml uses `object_storage` while `debug_info` seems more correct. This fixes it


s-urbaniak:
nevermind, i see this changed in master :-)

PR #1769, opened 2022-09-26
ui: Circular dependency removal and typescript transpiling in the published packages

manojVivek:
This PR contains:
- [x] Removed circular dependency between `@parca/profile` and `@parca/components` packages by moving the profiling related components from `@parca/components` to `@parca/profile`.
- [x] Typescript files are transpiled into `dist` before the packages are published.


manojVivek:
Please ignore the "UI documentation" build failure on Vercel for this PR. We need to add a step to build the packages before building the storybook, I'll do that once this branch is merged, as the commands needed for that are in this branch right now.

brancz (approved):

PR #1770, opened 2022-09-26
share: Share grpc query exposes profile description and grpc versioning

manojVivek:
Includes:
- [x] `ProfileTypes` query exposes profile description.
- [x] `share.proto` changes to comply with `buf lint`.


proto/parca/share/v1alpha1/share.proto
```diff
@@ -0,0 +1,80 @@
...
+  // description of the profile uploaded.
+  string description = 2;
+}
+
+// QueryResponse is the returned report for the given query.
+message QueryResponse {
+  // report is the generated report
+  oneof report {
+    // flamegraph is a flamegraph representation of the report
+    parca.query.v1alpha1.Flamegraph flamegraph = 5;
```
brancz, line 69: I expected these to start at 1

brancz (approved):
I'm sort of surprised to see no changes to the components that are going to use the response from either of these API responses. Do they already only pass the fields? If that's the case then this lgtm!

proto/parca/share/v1alpha1/share.proto
manojVivek: Fixed the field numbers.

Yeah, Parca doesn't use the modified responses. It's only used in the sharing service right now. Will take care of those in the next PR once these are merged.

PR #1775, opened 2022-09-26
parca: cancel parcaserver application context when handling an interrupt

asubiotto:
Previously, the context that was used to call ListenAndServe was never cancelled.

cc @brancz

### DIFF
--- a/pkg/parca/parca.go
+++ b/pkg/parca/parca.go
@@ -584,35 +584,44 @@ func runScraper(
 		},
 	)
 
-	parcaserver := server.NewServer(reg, version)
-	gr.Add(
-		func() error {
-			return parcaserver.ListenAndServe(
-				ctx,
-				logger,
-				flags.Port,
-				flags.CORSAllowedOrigins,
-				flags.PathPrefix,
-				server.RegisterableFunc(func(ctx context.Context, srv *grpc.Server, mux *runtime.ServeMux, endpoint string, opts []grpc.DialOption) error {
-					scrapepb.RegisterScrapeServiceServer(srv, m)
-					if err := scrapepb.RegisterScrapeServiceHandlerFromEndpoint(ctx, mux, endpoint, opts); err != nil {
-						return err
-					}
-					return nil
-				}),
-			)
-		},
-		func(_ error) {
-			ctx, cancel := context.WithTimeout(ctx, 30*time.Second) // TODO make this a graceful shutdown config setting
-			defer cancel()
+	{
+		parcaserver := server.NewServer(reg, version)
+		serveCtx, cancelServe := context.WithCancel(ctx)
+		gr.Add(
+			func() error {
+				return parcaserver.ListenAndServe(
+					serveCtx,
+					logger,
+					flags.Port,
+					flags.CORSAllowedOrigins,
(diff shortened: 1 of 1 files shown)


pkg/parca/parca.go
```diff
@@ -584,35 +584,43 @@ func runScraper(
...
 
-			level.Debug(logger).Log("msg", "server shutting down")
-			err := parcaserver.Shutdown(ctx)
-			if err != nil && !errors.Is(err, context.Canceled) {
-				level.Error(logger).Log("msg", "error shutting down server", "err", err)
-			}
-		},
-	)
+				// Cancels ListenAndServe.
+				cancel()
```
brancz, line 611: we should only call this cancel after we've exceeded the graceful timeout, otherwise, requests may error in context canceled even if they would have completed within the graceful shutdown window

pkg/parca/parca.go
```diff
@@ -584,35 +584,43 @@ func runScraper(
...
+
+				// Create a new context for the server shutdown with a deadline.
+				ctx, cancel := context.WithTimeout(ctx, 30*time.Second) // TODO make this a graceful shutdown config setting
```
brancz, line 614: I think just for it to be easier to reason about I would name this `ctx` and `cancel` differently

pkg/parca/parca.go
asubiotto: Done.

pkg/parca/parca.go
asubiotto: Renamed the serve and shutdown contexts. Good thing too, the previous version of the code would've created a child context of the already canceled context for Shutdown, making it a noop.

pkg/parca/parca.go
```diff
@@ -584,35 +584,44 @@ func runScraper(
...
+			func() error {
+				return parcaserver.ListenAndServe(
+					serveCtx,
+					logger,
+					flags.Port,
+					flags.CORSAllowedOrigins,
+					flags.PathPrefix,
+					server.RegisterableFunc(func(ctx context.Context, srv *grpc.Server, mux *runtime.ServeMux, endpoint string, opts []grpc.DialOption) error {
+						scrapepb.RegisterScrapeServiceServer(srv, m)
+						if err := scrapepb.RegisterScrapeServiceHandlerFromEndpoint(ctx, mux, endpoint, opts); err != nil {
```
brancz, line 600: The code within this function reads to me the way that we should also pass the `serveCtx` here.

brancz (approved):
One last comment then lgtm

pkg/parca/parca.go
asubiotto: Not sure I understand. I think this `ctx` shadows the outer `ctx` and is a child of `serveCtx`

brancz: The way I read the code within `RegisterScrapeServiceHandlerFromEndpoint` is that the passed context will be used to serve requests. Therefore the right context to use here is `serveCtx` as once cancelled with the `cancelServe` function those execution paths should cancel and not only when the main/parent context gets cancelled at the very end just before exiting the process.

manojVivek:
@all-contributors Please add @asubiotto for code.

Issue #939, opened 2022-04-13
Show parca agents pushing data on targets page


brancz:
It is quite confusing to be able to query data but not be able to see any targets on the targets page. This is because the targets page currently only shows targets that are being scraped (pulled). It would be useful to also list the agents actively pushing data to Parca.

maxbrunet:
Done in #1819

PR #1819, opened 2022-10-04
pkg/profilestore: implement agents RPC

heylongdacoder:
Relates to: #939


heylongdacoder:
proto file comment still pending. I will do it once the the overall changes has been agreed. Thanks! :)

brancz (approved):
I don't think we need to add the bool to the WriteRawRequest message, otherwise, this approach looks good! If there is a lot of contention on the mutex we might want to optimize the access pattern there, but I think this is fine for a start!

pkg/profilestore/profilecolumnstore.go
```diff
@@ -139,10 +150,91 @@ func (s *ProfileColumnStore) WriteRaw(ctx context.Context, req *profilestorepb.W
...
+	writeErr := s.writeSeries(ctx, req)
+
+	if req.Agent && len(req.Series) != 0 {
+		ag := agent{
+			lastPush:         start,
+			lastPushDuration: time.Since(start),
+			lastError:        writeErr,
+		}
+		nodeName, err := nodeNameFromLabels(req.Series)
+		if err != nil {
```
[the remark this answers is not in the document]
heylongdacoder: That is a good idea! It is done. I also renamed the proto file RPC name to make it more consistent with the `targets` RPC. Documentations are added as well. Thanks for the review! :)

heylongdacoder:
I have one question. May I know what is the purpose of this line(https://github.com/parca-dev/parca/blob/main/pkg/profilestore/profilecolumnstore.go#L62)?

Do I need to add `var _ profilestorepb.AgentsServiceServer = &ProfileColumnStore{}` below it?

brancz:
You don't have to add that. It just ensures that `ProfileColumnStore` implements the `profilestorepb.ProfileStoreServiceServer` interface at compile time.

pkg/profilestore/profilecolumnstore.go
```diff
@@ -139,10 +154,98 @@ func (s *ProfileColumnStore) WriteRaw(ctx context.Context, req *profilestorepb.W
...
 	}
 
+	return nil
+}
+
+func (s *ProfileColumnStore) updateAgents(ip string, ag agent) {
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+
+	s.agents[ip] = ag
```
metalmatze, line 169: When updating the agents it probably makes sense to also clean up "old" agents that don't push anymore.
It's probably best to do this in a separate goroutine eventually, but for now, I think we can start cleaning them up here.
Essentially something like this:
```suggestion
	s.agents[ip] = ag
	
	for ip, a := range s.agents {
		if a.lastPush.Before(time.Now().Add(-5*time.Minute) {
			delete(s.agents, ip)
		}
	}
```
_I've written this code here on GitHub so it might not be 100% correct._

Already on https://demo.parca.dev I can see this becoming a problem very quickly, where there are always 4 agents at a time right now and we might deploy or restart them a couple of times daily.

metalmatze (commented):
Really looking forward to this feature!

pkg/profilestore/profilecolumnstore.go
heylongdacoder: Ahh, i forgot to handle this. Thanks for the review and pointing this out! Done.

metalmatze (approved):

kakkoyun (approved):

PR #1829, opened 2022-10-05
ui: Add a watch mode for `tsc`

yomete:
I noticed that we lost hot reloading (when making changes to any file in the `packages/shared/*`) locally and traced this down being to the fact that we specify the entry file, `main`, in the packages to be `dist/index.js`, and not the actual source code, which is in `src`.  The `dist` folder is where the transpiled Typescript code is stored, as we now [transpile](https://github.com/parca-dev/parca/pull/1769) our `@parca` components before publishing.

So to enable hot reloading for the shared `@parca` packages, I added a `watch` script that uses Lerna to run the `tsc --watch` command in all affected packages. The `tsc --watch` mode will listen for changes to the source file and automatically transpile them into the `dist` folder.

I've also edited the documentation to reflect this.


ui/packages/shared/components/package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
     "test": "jest --coverage --config ../../../jest.config.js ./src/*",
     "prepublish": "yarn build",
-    "build": "tsc && tailwindcss -o dist/styles.css --minify"
+    "build": "tsc && tailwindcss -o dist/styles.css --minify",
+    "watch": "tsc --watch && tailwindcss -o dist/styles.css --minify"
```
maxbrunet, line 10: I suppose `--watch` does not exit unless it is interrupted, so anything right of `&&` will never run.

`tailwindcss` also has a `--watch` flag, looks like it would need to run in parallel to `tsc`

yomete: Good catch. I'll modify that

yomete: Interestingly, after removing everything right of `&&`, Tailwind CSS changes still cause a hot reload of the app. I assume this is because the `dist` folder already contains the Tailwind CSS file itself, and we're merely now just changing classnames in components.

manojVivek: No, I think, it just works when you add a class that is already there in the css file. 
It'll break if we add a new class that is not used in any of the source files yet, as tailwind scans the source files and adds styles to the output css only for the classes that are found in the scanned files. 

We should put back the tailwind command for it to work properly with future changes as well.

yomete: got it. I'll open another PR for that.

PR #1862, opened 2022-10-10
Use tsc-watch to watch TypeScript files and compile Tailwind files.

yomete:
I added [tsc-watch](https://www.npmjs.com/package/tsc-watch) which allows us to run another command in addition to watching the TypeScript files. It has a `onSuccess` flag that lets you execute any command on every successful TypeScript compilation.


manojVivek (approved):
Looking great! 🚀

PR #1871, opened 2022-10-11
ci(ui): remove usage of set-output workflow command

maxbrunet:
It has been deprecated: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


manojVivek (approved):
Lgtm!

PR #1881, opened 2022-10-13
ui: Update Parca pkgs versions correctly

yomete:
We were incorrectly using wrong versions of some Parca packages. Noticed this while trying to update Parca components itself in the cloud app.


brancz (approved):

ui/packages/shared/components/src/DateTimeRangePicker/DateTimeRangePickerPanel/AbsoluteDatePicker/index.tsx
```diff
@@ -17,8 +17,6 @@ import {AbsoluteDate, DateTimeRange, getDateHoursAgo} from '../../utils';
 import Button from '../../../Button';
 import {UTCDateTimePicker} from '../../../DateTimePicker';
 
-import 'react-datepicker/dist/react-datepicker.css';
```
yomete, line 20: Removing this line here and placing it in the `ui/packages/app/web/src/App.tsx` file should hopefully solve the [css-npm](https://nextjs.org/docs/messages/css-npm) error seen on cloud.

manojVivek: But this would be an inconvenience to make the consuming project to import this css file explicitly. 

It's ok for now, in the future we can look into an alternative way to import those styles into `DatePicker` component only so that the consumers don't need to worry about it.

yomete: Yeah it was the most ok solution I could think of for now. Definitely will revisit in the future.

PR #1910, opened 2022-10-17
ci: Disabling cancel-in-progress for Publish UI components workflow

manojVivek:
Enabling cancel-in-progress for the publish UI component job is creating a bug where the actual publish of the package to npm is failing. Lerna pushes the updated git tags to GitHub before publishing to npm, which is triggering another run of the workflow and in turn cancelling the in-progress publishing to npm.
And the newly triggered run wouldn't publish it either, as it already has the git tags that indicate that the npm publishing is update to date. 

Disabling `cancel-in-progress` for this workflow would fix it.

### DIFF
--- a/.github/workflows/ui-publish-components.yml
+++ b/.github/workflows/ui-publish-components.yml
@@ -7,7 +7,7 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.number || github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: false
 
 jobs:
   skip-check:


metalmatze (approved):

maxbrunet (approved):
Adding `[skip ci]` to the Lerna release commits could also be a solution

manojVivek:
> Adding `[skip ci]` to the Lerna release commits could also be a solution

Oh yeah, I didn't think about that one. 
But looking at Lerna's docs now, it looks like we cannot configure the commit message with the `publish` command only. We need to do `version` with commit message and then `publish` with `from-git` flag. 
The current workflow seems clean, so will just stick to disabling `cancel-in-progress` for now and revisit later when there is a need.

PR #1930, opened 2022-10-19
Fix elf based symbolization

javierhonduco:
There were two issues:
- There was an off-by-one, which the first commit fixes + we were using the dynamic symbols, too, which I don't belive it's correct (but I will double check in a bit);
- Symbols were not demangled, which affects C++ and Rust

## Test plan

- Added some more unittests
- Cross checked against GNU's addr2line
- End-to-end visual checks below

**Before**
(note that symbols are not correct, and also, mangled)
image

**After**:

image

Signed-off-by: Francisco Javier Honduvilla Coto <javierhonduco@gmail.com>


brancz (approved):
Nice!

javierhonduco:
Vercel failed, and even approving to run on this PR didn't do the trick

kakkoyun:
Amazing ❤️

PR #1932, opened 2022-10-19
ui: Use BigInt to represent 64 bit addresses

javierhonduco:
This is necessary as the maximum number that JS can represent is 2^53 [0].And show the addresses in the popover in hexadecimal.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt

- [0] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER
## Test plan

image

Signed-off-by: Francisco Javier Honduvilla Coto <javierhonduco@gmail.com>

### DIFF
--- a/ui/packages/shared/profile/src/GraphTooltip/index.tsx
+++ b/ui/packages/shared/profile/src/GraphTooltip/index.tsx
@@ -103,10 +103,10 @@ const TooltipMetaInfo = ({
             <td className="w-4/5 break-all">
               <CopyToClipboard
                 onCopy={onCopy}
-                text={' 0x' + hoveringNode.meta.location.address.toString()}
+                text={hexifyAddress(hoveringNode.meta.location.address)}
               >
                 <button className="cursor-pointer">
-                  {' 0x' + hoveringNode.meta.location.address.toString()}
+                  {hexifyAddress(hoveringNode.meta.location.address)}
                 </button>
               </CopyToClipboard>
             </td>
--- a/ui/packages/shared/profile/src/utils.ts
+++ b/ui/packages/shared/profile/src/utils.ts
@@ -18,7 +18,7 @@ export const hexifyAddress = (address?: string): string => {
   if (address == null) {
     return '';
   }
-  return `0x${parseInt(address, 10).toString(16)}`;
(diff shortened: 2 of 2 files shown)


brancz:
cc @manojVivek @yomete I think there was an option in protobuf-ts to use BigInt by default to represent integers? I suppose we should do that?

manojVivek (approved):
Nice catch! 🔬

manojVivek:
> cc @manojVivek @yomete I think there was an option in protobuf-ts to use BigInt by default to represent integers? I suppose we should do that?

Yeah looks like we can configure it and the default is `BigInt` already: https://github.com/timostamm/protobuf-ts/blob/master/MANUAL.md#bigint-support

brancz:
Looks like we explicitly configure to not do that though https://github.com/parca-dev/parca/blob/2b67154d490d9565d77ad4f11783f60045f46dca/buf.gen.yaml#L25

brancz:
Looks like I introduced that, if I recall correctly that was because that was how the previous protobuf library was doing it and it allowed me to do less work. We should probably change that.

brancz (approved):
I do think we should switch the protobuf default, but better to merge this fix now and worry about the larger fix as a follow-up in the future. As far as I'm aware this should also be the only place where we care about this, so we might not need to do anything as a follow up, it just seems like the better thing to do long term.

javierhonduco:
Happy to either merge this PR as a quick fix, or close it if implementing the direct deserialisation of these values as BigInt isn't too hard 😄

maxbrunet:
> Looks like I introduced that, if I recall correctly that was because that was how the previous protobuf library was doing it and it allowed me to do less work. We should probably change that.

Yup, I was review that when I enabled TS strict mode, some browser (Safari iirc) did not support it, but from the MDN link support look good in recent versions

FYI I also mention it here: https://github.com/parca-dev/parca/pull/1664#discussion_r963078411

It seems to me the cleanest way would to migrate from https://github.com/timostamm/protobuf-ts to https://github.com/bufbuild/protobuf-es. Maybe this can be extracted from the connect-web PR

PR #1950, opened 2022-10-21
dev: update jsonnet version to 635d04f

asubiotto:
This commit fixes make dev/setup failing due to an older version of golang.org/x/sys in github.com/google/go-jsonnet. The version was bumped in https://github.com/google/go-jsonnet/pull/633.


asubiotto:
Pinning to a non-released commit is not ideal but without this building doesn't work. I should maybe add a comment that we should set it back to a release once the fix is available in GA.

PR #1959, opened 2022-10-22
fix(config): ensure path_prefix comes with default pprof_config

### LINKED ISSUE #1956: path_prefix behaviour
I have a prometheus with a path, i.e. `foo.bar/prometheus` and tried to configure Parca with:

```
      - job_name: "prometheus"
        scrape_interval: "15s"
        static_configs:
          - targets: ['foo.bar.svc.cluster.local:9090']
        profiling_config:
          path_prefix: /prometheus
```

Which ignored the path prefix.

After I've configured it like:

```
      - job_name: "prometheus"
        scrape_interval: "15s"
        static_configs:
          - targets: ['foo.bar.svc.cluster.local:9090']
        profiling_config:
          path_prefix: /prometheus
          pprof_config:
            memory:
              enabled: true
              path: /debug/pprof/allocs
```

It's now scraping every profile. 

Would be nice if I can omit a default pprof_config and just only set path_prefix (and document this, I figured this out from your .go testing files :)

maxbrunet:
Fixes #1956

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -190,8 +190,10 @@ func (c *ScrapeConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		return err
 	}
 
-	if unmarshalled.ProfilingConfig == nil || unmarshalled.ProfilingConfig.PprofConfig == nil {
+	if unmarshalled.ProfilingConfig == nil {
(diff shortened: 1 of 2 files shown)


brancz (approved):
lgtm!

PR #1976, opened 2022-10-25
Use deduplicated tables for strings, mappings, locations and functions  

metalmatze:
Up until now the flamegraphs contained all the data for each span/trace in the flamegraph. 
This comes at a big cost of bandwidth and also latency - both for building it on the server side, but especially when sending it across the wire. 

Going forward we'll have a `REPORT_TYPE_FLAMEGRAPH_TABLE` that is going to deduplicate the metadata and then each span is going to reference it's metadata in the table/array by index. 
The biggest gain comes from sending each strings only once. In the future there might be the possibility of using compression or radix trees to further reduce the amount of strings, definitely not a big issue for now.

So far I've seen profiles being only 10% in size of what they've been before or to rephrase we can send 10x bigger profiles with the same size limits. 🎉


proto/parca/metastore/v1alpha1/metastore.proto
```diff
@@ -298,4 +313,10 @@ message Mapping {
 
   // has_inline_frames indicates whether the mapping has associated inline frames.
   bool has_inline_frames = 10;
+
+  // fileStringIndex is the index in the string table to the file name associated with the mapping.
+  uint32 file_string_index = 11;
```
yomete, line 318: curious to what the difference is between `filename_string` and `file_string`

yomete (approved):
lgtm on the frontend side

kakkoyun (approved):
LGTM ❤️

thorfour (approved):

ui/packages/shared/profile/src/IcicleGraph.tsx
```diff
@@ -49,6 +54,10 @@ interface IcicleGraphNodesProps {
 
 interface IcicleGraphRootNodeProps {
   node: FlamegraphRootNode;
+  strings: string[];
```
manojVivek, line 57: As a future improvement, we can create a `FramegraphContext` and access the tables through that so that we don't have to pass them as props to every component.

manojVivek (approved):
Lgtm!

proto/parca/metastore/v1alpha1/metastore.proto
metalmatze: This is just following whatever the meta structure already had, adding a `_string_index` suffix.

ui/packages/shared/profile/src/IcicleGraph.tsx
metalmatze: That would be amazing! I'll create an issue to track it.

PR #1999, opened 2022-10-27
ui: Synchronize clicking on Top table and Icicle graph

yomete:
Clicking on any row in the Top table will now highlight the same span in the Icicle graph.


manojVivek (approved):
Lgtm! 🚀

PR #2030, opened 2022-11-03
ui: Date picker timezone fix and minor UX improvement

### LINKED ISSUE #2023: Display time in UTC in both summary and input fields
As can be seen in the image below, the timezone seems to be inconsistent across the input and summary fields for the time selector dialog which can be confusing.

image

manojVivek:
Fixes #2023


monicawoj (approved):
🎉

PR #2032, opened 2022-11-03
Sending perf data upstream via ParcaContext

yomete:
With this changes we can now utilize the `ParcaContext` context to send perf data. For starters, it utilizes both the Profiler API from React and the afterframe pkg to measure perf.  I'm still not sure which of them to use but for now I'm playing around with both of them to see which is better for our use case.

Here's a [gist](https://gist.github.com/yomete/73e856481062d78fef77a91e4052ac44) of how we'd use this in other apps.


ui/packages/app/web/src/pages/testpage.tsx
```diff
@@ -0,0 +1,61 @@
...
+    parcaContextDefaultValue
+  );
+
+  console.log('parcaContextState', parcaContextState.trackedPerfEvents);
+
+  return (
+    <ParcaContextProvider
+      value={{
+        loader: <p>loading...</p>,
+        dispatch: dispatchToParcaContext,
```
manojVivek, line 52: Instead of passing two fields, for state and setState for the passing the data, should we simplify it to a simple callback whenever there is a perf data? 

Like:
```
<ParcaContextProvider
      value={{
        loader: <p>loading...</p>,
        perf: {
          onRender: // handler that can handler React Profiler API.
          ... // other perf handler that we need to support
         }
      }}
>
...
</ParcaContextProvider>
```
I feel there is there would be less need to send back the data into the child components? 

Wdyt?

manojVivek (approved):
Lgtm!

Issue #1633, opened 2022-08-30
ui: stop using TypeScript Enums


maxbrunet:
While working on #1632, I noticed Enums are awkward to work with and often result in odd typing like `as keyof typeof SomeEnum`.

[eslint-plugin-typescript-enum](https://www.npmjs.com/package/eslint-plugin-typescript-enum) could be used and has more explanations in its README.

PR #2045, opened 2022-11-07
ui: Add label-value autocompletion based on parser

brancz:


ui/packages/shared/parser/src/index.ts
```diff
@@ -45,6 +45,54 @@ function matcherTypeFromString(matcherTypeString: string): MatcherType {
...
+
+  while (currentState.right.rule.name !== 'labelName') {
+    currentState = currentState.left;
+  }
+
+  return currentState.right.data.value;
+}
+
+export interface LiteralSuggestion {
+  type: 'literal';
```
monicawoj, line 63: let's maybe save these type scripts to a Typescript enum? Looks like some of the values are re-used so can help us avoid misspellings. 

For example,
```
export enum SuggestionType {
  Literal = 'literal',
  Matcher = 'matcherType',
  LabelName = 'labelName',
  LabelValue = 'labelValue',
  ProfileName = 'profileName'
}
```

ui/packages/shared/parser/src/index.ts
```diff
@@ -249,12 +308,12 @@ export class Query {
       // We're not going to skip suggesting to type a whitespace character.
       if (!(nextSymbol.type !== undefined && nextSymbol.type === 'space')) {
         if (nextSymbol.literal !== undefined) {
-          const suggestion = {type: 'literal', value: nextSymbol.literal as string};
+          const suggestedValue = nextSymbol.literal as string;
           if (
-            suggestions.findIndex(s => s.type === 'literal' && s.value === suggestion.value) === -1
+            suggestions.findIndex(s => s.type === 'literal' && s.value === suggestedValue) === -1
```
monicawoj, line 313: If we save the suggestion types as an enum (as suggested here: https://github.com/parca-dev/parca/pull/2045/files#r1015431013) we could reuse those values in the following manner:

```suggestion
            suggestions.findIndex(s => s.type === SuggestionType.Literal && s.value === suggestedValue) === -1
```

ui/packages/shared/profile/src/ProfileMetricsGraph/index.tsx
```diff
@@ -42,36 +42,39 @@ export const useQueryRange = (
...
-): IQueryRangeResult => {
-  const [isLoading, setIsLoading] = useState<boolean>(false);
-  const [error, setError] = useState<any>(null);
-  const [response, setResponse] = useState<QueryRangeResponse | null>(null);
+): IQueryRangeState => {
+  const [state, setState] = useState<IQueryRangeState>({
+    response: null,
+    isLoading: true,
+    error: null,
+  });
```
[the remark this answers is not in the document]
brancz: not just cleaner, before we *never* cleared the error state

ui/packages/shared/parser/src/index.ts
brancz: they're not reused they're used to select the right concrete type from the `Suggestion` union type, because of the union type, they already act as as enum at compile time

ui/packages/shared/profile/src/ProfileMetricsGraph/index.tsx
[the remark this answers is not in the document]
monicawoj: yes yes, great catch! By "cleaner" I meant exactly that! Not just how the code looks, but more-so how it works - saving loading/response/error in a single "state" rather than separate because we can't have two of these simultaneously. If we are loading, we want to clear the previous results/error, if we have a result or error, we are not loading, so and so forth. Definitely helps us avoid logical errors like the one you mentioned of never clearing the error state

ui/packages/shared/parser/src/index.ts
monicawoj: Ok, gotcha! Just wanted to make sure a potentially misspelled suggestion type string would be flagged as a ts error, if needed

monicawoj (approved):

ui/packages/shared/parser/src/index.ts
maxbrunet: FYI TS `enum` is to be avoided, it is not fully type safe, there are better alternatives, please see #1633 for details

yomete (approved):
ditto with what Monica said but apart from that lgtm! amazing!
we can also merge as it is and then update in another PR

ui/packages/shared/parser/src/index.ts
brancz: Yeah this would only work if we used the enum, but based on the above comments, this is not actually duplicate, as there would be a type error if we misspelled this.

PR #2049, opened 2022-11-07
Callgraph edits

monicawoj:
This PR includes two small fixes:
- zoom in/out (bug where zoom through trackpad was just zooming in)
- search for empty string (all nodes were hidden when search value was empty string)


yomete (approved):
lgtm!

Tssted the zoom and it works perfectly 👍🏾 :shipit:

PR #2055, opened 2022-11-08
pkg/query: Fix flamegraph function deduplication

brancz:
We were resetting all function IDs to be `""`, which caused functions to override each other seemingly randomly.


metalmatze (approved):

metalmatze:
There are some tests that need to be updated. 
I remember now why I set `Id=""`, it was to not send those strings over the wire too. Maybe we need to somehow remove them later in the calls, probably after the aggregation.

brancz:
Yep, just pushed a fix to do exactly that.

PR #2062, opened 2022-11-09
ft: `Filter by function` functionality added to ProfileView

manojVivek:


monicawoj (approved):
This looks good to me! As mentioned on Discord, I'm working on the visualization top bar redesign, which may slightly shift some things around in terms of UI (which we can of course discuss), but the functionality looks on point!

pkg/query/columnquery.go
```diff
@@ -196,6 +221,7 @@ func (q *ColumnQueryAPI) renderReport(ctx context.Context, p *profile.Profile, t
 			Report: &pb.QueryResponse_Top{Top: top},
 		}, nil
 	case pb.QueryRequest_REPORT_TYPE_CALLGRAPH:
+		fmt.Println("callgraph")
```
brancz, line 224: debugging leftover I assume?

pkg/query/columnquery.go
```diff
@@ -142,9 +143,33 @@ func (q *ColumnQueryAPI) Query(ctx context.Context, req *pb.QueryRequest) (*pb.Q
...
+		p = filterProfileData(p, *req.FilterQuery)
+	}
+
 	return q.renderReport(ctx, p, req.GetReportType())
 }
 
+func filterProfileData(p *profile.Profile, filterQuery string) *profile.Profile {
+	filteredSamples := []*profile.SymbolizedSample{}
+	for _, s := range p.Samples {
+		var lines []profile.LocationLine
```
brancz, line 156: We don't need to copy here. I would suggest the inside of this loop should be extracted to a separate function that's just called `keepSample` and if it returns true, then we append to `filteredSamples`. All the `keepSample` function then does is loop over all lines of all locations and on the first match it returns `true` and otherwise if no location line contained the search string, then it returns false.

brancz (commented):
I like this feature a lot! A few small comments, but I think we're very close!

pkg/query/columnquery.go
manojVivek: Ahh, yes, my bad.

pkg/query/columnquery.go
manojVivek: Done

brancz (approved):
Love it! lgtm

PR #2063, opened 2022-11-09
pkg/query: Fix locations with no lines

brancz:
Previously we were skipping locations that had no lines in the response.


kakkoyun (approved):
LGTM 🚀

PR #2065, opened 2022-11-09
pkg/query: Fix comparing of two locations with only addresses

brancz:
Previously there was a panic, when two locations were being compared that had no lines.


kakkoyun (approved):
LGTM

PR #2066, opened 2022-11-09
pkg/symbol/elfutils: Fix handling of stripped Go binaries

brancz:
Go binaries that are stripped, for example built with:

```
go build -ldflags "-w -s" main.go
```

Have an empty symbol table section, which some additional build steps may strip, since it's empty.

Before this patch, those Go binaries would run into an error case when attempting to load the symbol table.

Since the only truly necessary section for symbolization is the `.gopclntab` section, this patch proposes to only check for this section.


kakkoyun (approved):
LGTM

PR #2101, opened 2022-11-14
pkg/debuginfo: Always attempt to reupload corrupted files

brancz:


metalmatze (approved):

metalmatze:
I wonder if we should cut a v0.14.1 with this?

brancz:
Let's see if this improves the errors we're seeing on the demo instance. If this improves it, then we should I think.

PR #2110, opened 2022-11-15
ui: Make the perf value in ParcaContext optional

yomete:


ui/packages/shared/components/src/ParcaContext/index.tsx
```diff
@@ -19,7 +19,7 @@ interface Props {
   perf: {
```
manojVivek, line 19: Should we make this a `perf?:` instead of the `| null` at the end? 
I believe, the `| null`, would still enforce a `perf: null` in the passed object.

yomete: TIL: you can do this `perf?`. that sgtm

manojVivek (approved):
Looks great!

PR #2114, opened 2022-11-15
Change log levels to not log recoverable errors as error level

metalmatze:
While running Parca locally with parca-agent these logs made it seem like there were errors that needed to be fixed. While I think they are useful as debug logs, I don't think Parca should constantly log these messages.


brancz (approved):
Thank you!! This is indeed a huge improvement!

PR #2115, opened 2022-11-15
Profile vis state in URL

monicawoj:


monicawoj:
Initially, I believed that we would be able to entirely remove the Profile visualization state from the Redux store, however, this is not the case. _If_ we were using Parca as a standalone tool, the URL entirely could serve as the state - which would be optimal for shareability. However, because Parca is also a plugin to other services (such as Grafana, which can show multiple Parca panels), we 1) do not want to pollute the URL in other services and 2) a single URL will not be enough to hold the state of multiple Parca instances. Therefore, the updated approach will be to save the profile UI state in the Redux store, as is currently the case, and use a useEffect hook to update the URL query params whenever a "state" value in the Redux store changes so that the URL also always reflects the current state of the application. The URL updating should be behind some sort of flag so that we can turn it on/off as needed (for example, turning off if we do not want to pollute a third-parties service when Parca is used as a plugin).

ui/packages/shared/functions/src/hooks/useURLState.tsx
```diff
@@ -0,0 +1,60 @@
...
+import {useNavigate} from 'react-router-dom';
+import {parseParams, convertToQueryParams} from '../index';
+import {useAppDispatch, useAppSelector, setGlobalStateValue} from '@parca/store';
+
+// every time the param value changes, the URL is updated
+export const useURLState = (paramName: string, withURLUpdate: boolean = true) => {
+  const navigate = useNavigate();
+  const dispatch = useAppDispatch();
+  const router = parseParams(window.location.search);
+  const storeValue = useAppSelector(state => state.profile[paramName]);
```
manojVivek, line 24: Shall we move this selector into the slice and import it from there so that the select logic is not dispersed across components?

Similar to this: https://github.com/parca-dev/parca/blob/9efbbdb380472db6b27702da2a7a61e341352af0/ui/packages/shared/store/src/slices/profileSlice.ts#L51

ui/packages/shared/functions/src/hooks/useURLState.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  // 1. set the initial state to the store value or the URL value
+  const [value, setValue] = useState<string | string[] | undefined>(
```
manojVivek, line 27: Do we need this additional local state here? We can just return the dispatching logic as the `setValue` of this hook. 

Like:
```
const setValue = (val) => {
  const isEmpty = val => val === undefined || val === null || val === '';
  // 2a. update the local store value
  if (storeValue !== value) {
    dispatch(setGlobalStateValue({key: paramName, value}));
  }
};
```

ui/packages/shared/functions/src/hooks/useURLState.tsx
```diff
@@ -0,0 +1,60 @@
...
+// limitations under the License.
+
+import {useState, useEffect, Dispatch} from 'react';
+import {useNavigate} from 'react-router-dom';
+import {parseParams, convertToQueryParams} from '../index';
+import {useAppDispatch, useAppSelector, setGlobalStateValue} from '@parca/store';
+
+// every time the param value changes, the URL is updated
+export const useURLState = (paramName: string, withURLUpdate: boolean = true) => {
+  const navigate = useNavigate();
```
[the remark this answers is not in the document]
monicawoj: Not too difficult to prop drill and this seems like the simpler solution in my opinion

ui/packages/shared/functions/src/hooks/useURLState.tsx
monicawoj: You're right, this isn't necessary. Removing and will just use dispatch as the returned setValue

yomete (approved):
lgtm! amazing stuff 👏🏾

PR #2121, opened 2022-11-16
ProfileIcicleGraph Component performance benchmark

manojVivek:
Sample report:
```############## Benchmark Report ##############
Run at: 16/11/2022, 15:32:38
Results:
┌─────────┬──────────────────────────┬─────────────┬────────┬───────┬───────┬────────┬────────┬────────┐
│ (index) │           name           │ sampleCount │  mean  │  min  │  max  │  p50   │  p75   │  p90   │
├─────────┼──────────────────────────┼─────────────┼────────┼───────┼───────┼────────┼────────┼────────┤
│    0    │ 'ProfileIcicleGraph-10M' │     32      │ 6.512  │ 5.133 │ 7.462 │ 6.467  │ 6.733  │ 7.308  │
│    1    │ 'ProfileIcicleGraph-1M'  │     32      │  2.4   │ 1.923 │ 2.949 │ 2.432  │ 2.564  │ 2.649  │
│    2    │ 'ProfileIcicleGraph-20M' │     32      │ 15.568 │ 12.5  │  18   │ 15.833 │ 16.333 │ 16.833 │
└─────────┴──────────────────────────┴─────────────┴────────┴───────┴───────┴────────┴────────┴────────┘


```


metalmatze:
I can see really big files being checked into the repository. `ui/packages/shared/profile/src/ProfileIcicleGraph/benchmarks/benchdata/parca-10m.json` and others.
What we do in Go is to generate these test cases in a deterministic fashion and before running the actual benchmark we reset any timers and other observers for the benchmark. Maybe something similar is possible?

### For the future

It's definitely nice that the Results table gets pretty printed. I would keep that. Additionally, however, I would look for a tool that allows you to compare benchmarks across various runs. 
Again, Go has the [benchstat](golang.org/x/perf/cmd/benchstat) tool that will take two text files and print a comparison of each benchmark run. 
https://pkg.go.dev/golang.org/x/perf/cmd/benchstat#hdr-Example

Usually, we take these outputs and post them along with the PR that tries to improve performance. Maybe there are nicer services in JS that automate parts of this?
https://github.com/polarsignals/frostdb/pull/247

manojVivek:
> I can see really big files being checked into the repository. `ui/packages/shared/profile/src/ProfileIcicleGraph/benchmarks/benchdata/parca-10m.json` and others.
> What we do in Go is to generate these test cases in a deterministic fashion and before running the actual benchmark we reset any timers and other observers for the benchmark. Maybe something similar is possible?

Hmm, I'll see what can be done for this. 

> For the future
> It's definitely nice that the Results table gets pretty printed. I would keep that. Additionally, however, I would look for a tool that allows you to compare benchmarks across various runs.

Yes, I thought about it too and kept it for the future. Will create an issue and keep track of it.

manojVivek:
Pushed a commit to remove the data files and populate them on the fly.

ui/packages/shared/profile/src/ProfileIcicleGraph/index.tsx
```diff
@@ -35,6 +35,8 @@ const ProfileIcicleGraph = ({
   const compareMode = useAppSelector(selectCompareMode);
   const {ref, dimensions} = useContainerDimensions();
 
+  console.log('graph', JSON.stringify(graph));
```
yomete, line 38: you probably missed this but there's a stray `console.log` here.

metalmatze (approved):
Very nice!

ui/packages/shared/profile/src/ProfileIcicleGraph/index.tsx
manojVivek: Done, taken care of.

PR #2122, opened 2022-11-16
Fix semantics and rename http flag

kakkoyun:
* Fix semantics for http address flag
* Deprecate port flag
* Unify debuginfo flags

### DIFF
--- a/pkg/parca/parca.go
+++ b/pkg/parca/parca.go
@@ -76,10 +76,12 @@ const (
 )
 
 type Flags struct {
-	ConfigPath         string   `default:"parca.yaml" help:"Path to config file."`
-	Mode               string   `default:"all" enum:"all,scraper-only" help:"Scraper only runs a scraper that sends to a remote gRPC endpoint. All runs all components."`
-	LogLevel           string   `default:"info" enum:"error,warn,info,debug" help:"log level."`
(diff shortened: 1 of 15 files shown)


pkg/parca/parca.go
```diff
@@ -133,6 +135,11 @@ func Run(ctx context.Context, logger log.Logger, reg *prometheus.Registry, flags
 		defer closer()
 	}
 
+	if flags.Port != "" {
+		level.Warn(logger).Log("msg", "flag --port is deprecated, use --http-address instead")
```
brancz, line 139: We broke a bunch of flags in this release I think adding one more doesn’t make a difference

brancz (approved):
lgtm

yomete (approved):

pkg/parca/parca.go
metalmatze: Agreed.

metalmatze (approved):

javierhonduco (approved):
🎉

pkg/parca/parca.go
kakkoyun: 👍

kakkoyun: I'd done it but then reverted; it broke the snap. I think this is one of the oldest flags so let's don't break the backwards compatibility on this one.

PR #2124, opened 2022-11-16
pkg/scrape: Allow profiles to be treated as already normalized

brancz:
While this may be necessary for some pprof HTTP endpoints, we have encountered others now that are already normalized, in which case normalizing them again is incorrect.

This is save to be added like this without changing existing behavior, as the value defaults to false which is the currently used value before this patch.


metalmatze (approved):
Makes sense!

PR #2135, opened 2022-11-17
ui/perf: Attach the flamegraph sample count to markInteraction

yomete:


yomete:
I'll merge this in as soon as it's green so I can test this on cloud

PR #2136, opened 2022-11-17
Add debug flag for local binary builds

kakkoyun:


brancz:
Could we add a comment what these flags do?

kakkoyun:
> Could we add a comment what these flags do?

I'll do this in a subsequent PR. Sorry I missed the comment.

PR #2147, opened 2022-11-18
Unify debuginfo naming

kakkoyun:

⚠️ Breaking changes for the API ⚠️


kakkoyun:
pre-commit failure is because of the breaking gRPC API change. 
Decision to merge it up to other maintainers.

brancz:
I'm ok with this, but we should do a release of all components at once that fixes these across the board.

PR #2162, opened 2022-11-22
ci(pre-commit): apply automatic fixes

maxbrunet:
This is a new action which will apply pre-commit automatic fixes to the branch (e.g. formatting)

image

https://pre-commit.ci/lite.html

PR #2167, opened 2022-11-22
chore(pre-commit): add buf generate hook

maxbrunet:
With #2162, fixes can be apply automatically, so hopefully this would fix Renovate PRs (and unless they need rebasing they should be able to merge automatically too) :grin:

I'll see if I can add a `buf-generate` hook upstream to avoid hacking another.


kakkoyun (approved):
LGTM. Let's see it in action.

PR #2192, opened 2022-11-28
ui: Matcher Input box improvement

manojVivek:
This is a PR to convert the matcher query input box from `input` to `textarea`. 

This also has config changes to create PR preview deployments for Parca UI on Vercel that connects to demo.parca.dev of the API.


yomete (approved):
lgtm!

PR #2194, opened 2022-11-28
parca: minor fixes

Sylfrena:
- profilestore: remove debug value log
  - parcacol: add logging

### DIFF
--- a/pkg/parcacol/querier.go
+++ b/pkg/parcacol/querier.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/apache/arrow/go/v8/arrow"
 	"github.com/apache/arrow/go/v8/arrow/array"
+	"github.com/go-kit/log"
 	"github.com/polarsignals/frostdb/query"
 	"github.com/polarsignals/frostdb/query/logicalplan"
 	"github.com/prometheus/prometheus/model/labels"
@@ -51,12 +52,14 @@ type Engine interface {
 }
 
 func NewQuerier(
+	logger log.Logger,
 	tracer trace.Tracer,
 	engine Engine,
 	tableName string,
 	metastore metastorepb.MetastoreServiceClient,
 ) *Querier {
 	return &Querier{
+		logger:    logger,
 		tracer:    tracer,
 		engine:    engine,
 		tableName: tableName,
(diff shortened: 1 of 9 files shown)


pkg/parcacol/querier.go
```diff
@@ -543,6 +548,8 @@ func (q *Querier) QuerySingle(
 func (q *Querier) findSingle(ctx context.Context, query string, t time.Time) (arrow.Record, string, profile.Meta, error) {
 	requestedTime := timestamp.FromTime(t)
 
+	level.Info(q.logger).Log("bamboozling_requested_timestamp", requestedTime, "query", query)
```
brancz, line 551: lol .. also I don't think we need this log line anymore

pkg/parcacol/ingest.go
```diff
@@ -114,6 +114,8 @@ func (ing Ingester) IngestProfile(ctx context.Context, ls labels.Labels, p *prof
 	buf.Reset()
 	defer ing.bufferPool.Put(buf)
 
+	level.Info(ing.logger).Log("profile_timestamp", p.Meta.Timestamp)
```
brancz, line 117: no longer needed

brancz (commented):
just two small things, I don't think we need the log lines anymore, but I think we should keep the loggers so we can easily add log lines whenever we need to

pkg/parcacol/querier.go
Sylfrena: I... thought I had deleted this one. We definitely don't need it. 🙈

Sylfrena:
okay I don't understand why the docs check is failing- precommit passed for me- is there something that we are supposed to run if we change flags?

brancz:
Looks like you fixed it. In case you edited it manually, I think there is make target called `make README.md` that templates the flags into the readme.

Did you maybe forget to push a fix? I can still see both log lines, but the comments are resolved.

Sylfrena:
> Looks like you fixed it. In case you edited it manually, I think there is make target called `make README.md` that templates the flags into the readme.
> 
> Did you maybe forget to push a fix? I can still see both log lines, but the comments are resolved.
I did remove them here https://github.com/parca-dev/parca/pull/2194/commits/17c2eb10164d70b14ce509dcb34191f4cf5038ba
but they came back in the last commit? I don't know what happened there really, I will just repush things

brancz (approved):

PR #2198, opened 2022-11-29
ui: Fix escaped characters bug

yomete:
This PR fixes a bug where an inclusion of escaped characters in the `parca.yaml` file causes the profiler not to work as expected. To fix, we replace single `\` in the `labelValues` string with doubles `\\` if available and this is done on 3 levels:

1. When the label values are fetched for a label name during autocomplete suggestion
2. When a user clicks on any sample in the MetricsGraph to view a profile and
3. When a user clicks on any of the labels on the MetricsGraph tooltip
 

Resolves [#1060](https://github.com/parca-dev/parca-agent/issues/1060)


ui/packages/shared/profile/src/MatchersInput/index.tsx
```diff
@@ -79,7 +79,12 @@ const MatchersInput = ({
 
       call.response
         .then(response => {
-          setLabelValues(response.labelValues);
+          // replace single `\` in the `labelValues` string with doubles `\\` if available.
+          const newValues = response.labelValues.map(value =>
+            value.includes('\\') ? value.replace('\\', '\\\\') : value
```
manojVivek, line 84: I belive the `string.replace` will only replace the first matching pattern, so it wouldn't handle a second occurrence of `\\` in the string. 
I am not sure if multiple escape characters is a possibility here, but just in case.

manojVivek: Also, Should we extract this into a `sanitizeLabelValue` function in `@parca/functions` package and use it?

yomete: Yeah I just figured that was a possibility. making a change now

manojVivek: It's great that the CodeQL scanner captured it too. 😀

brancz (approved):
Nice work!! Works perfectly now. lgtm 👍 🎉

rafaeljesus:
is there a docker image / release I could use with this fix?

brancz:
Yep! There's always an image built from `main`. We're actually running `main-9b7a8c94` on the demo cluster right now, which has a ton of other really good improvements, I would suggest to try that one!

PR #2201, opened 2022-11-30
pkg/parcacol: Improve sparse labelset ingestion

metalmatze:
This is especially important when requests with various series exist where some series might not have labels in their labelset, however, the overall buffer written to storage has these label names available as column.

### DIFF
--- a/pkg/parcacol/sample.go
+++ b/pkg/parcacol/sample.go
@@ -47,6 +47,7 @@ func NormalizedProfileToParquetBuffer(w io.Writer, schema *dynparquet.Schema, ls
 		r = SampleToParquetRow(
 			schema,
 			r[:0],
+			names,
 			pprofLabels,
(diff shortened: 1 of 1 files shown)


pkg/parcacol/sample.go
```diff
@@ -156,29 +157,32 @@ func SampleToParquetRow(
 
 		// All remaining cases take care of dynamic columns
 		case ColumnLabels:
-			for _, label := range ls {
-				row = append(row, parquet.ValueOf(label.Value).Level(0, 1, columnIndex))
+			lset := ls.Map() // Convert labels to map for quicker lookup in loop.
```
brancz, line 160: this seems very expensive, maybe we can pass a map in the first place and let the caller decide how to optimize these allocations?

brancz (approved):

PR #2202, opened 2022-11-30
pkg/parcacol: Group all profiles of a write request into one buffer

brancz:
```
$ benchstat old.txt new.txt
name                              old time/op    new time/op    delta
ProfileColumnStoreWriteSeries-10    2.74ms ± 4%    0.00ms ± 6%  -99.99%  (p=0.000 n=10+10)

name                              old alloc/op   new alloc/op   delta
ProfileColumnStoreWriteSeries-10    4.05MB ± 0%    0.00MB ± 0%  -99.99%  (p=0.000 n=10+10)

name                              old allocs/op  new allocs/op  delta
ProfileColumnStoreWriteSeries-10     32.3k ± 0%      0.0k ± 0%  -99.98%  (p=0.000 n=10+10)
```


pkg/profilestore/profilecolumnstore.go
```diff
@@ -135,8 +141,144 @@ func (s *ProfileColumnStore) writeSeries(ctx context.Context, req *profilestorep
...
+}
+
+func sortedKeys(m map[string]struct{}) []string {
+	keys := make([]string, 0, len(m))
+	for k := range m {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	return keys
+}
```
metalmatze, line 188:
```suggestion
func sortedKeys(in map[string]struct{}) []string {
	if len(in) == 0 {
		return []string{}
	}

	out := maps.Keys(in)
	sort.Strings(out)
	return out
}
```
Sadly the sort.Sort isn't generic yet, but when it's going to be, there will only be one of this ever.

metalmatze:
YAY! :tada: :rocket: 
Soooo happy this has the same impact for Parca.

thorfour (approved):
LGTM

pkg/parcacol/ingest.go
```diff
@@ -18,211 +18,277 @@ import (
...
+	defer ing.schema.PutBuffer(pBuf)
 
-	normalizedProfiles, err := ing.normalizer.NormalizePprof(ctx, name, names, p, normalized)
-	if err != nil {
-		return fmt.Errorf("normalize profile: %w", err)
-	}
+	var r parquet.Row
+	for _, s := range series {
+		for _, normalizedProfiles := range s.Samples {
+			for _, p := range normalizedProfiles {
```
metalmatze, line 100: So this is the biggest change I would still propose to this.
Given that we built this nested `series>profiles>samples` construct earlier on, but then use it here for just a bit of reusing labels, I'd propose to flatten this earlier already.
Basically, the `Ingest` takes in `[]normalizedSamples` and we only have one loop to iterate over.

PR #2205, opened 2022-12-01
pkg/parcacol: NormalizeWriteRawRequest returns flat slice of samples

metalmatze:
This will get rid of the nested `series > profiles > samples`. We need to pass the labels and meta into each normalizedSample but in the end, don't need to loop more than once to ingest.

This makes distributing samples by hashing them a lot easier too.

```
name                              old time/op    new time/op    delta
ProfileColumnStoreWriteSeries-24     451ns ±17%    1430ns ±11%  +216.89%  (p=0.000 n=10+10)

name                              old alloc/op   new alloc/op   delta
ProfileColumnStoreWriteSeries-24      563B ± 0%     4568B ± 0%  +711.87%  (p=0.000 n=10+10)

name                              old allocs/op  new allocs/op  delta
ProfileColumnStoreWriteSeries-24      7.00 ± 0%      7.00 ± 0%      ~     (all equal)
```
The performance is quite a bit worse, but given we just came from a -99% decrease in #2202, I'd say the ergonomics are more important than the performance here.

### DIFF
--- a/pkg/parcacol/ingest.go
+++ b/pkg/parcacol/ingest.go
@@ -23,11 +23,9 @@ import (
 	"sync"
 
 	"github.com/go-kit/log"
-	"github.com/go-kit/log/level"
 	"github.com/klauspost/compress/gzip"
 	"github.com/polarsignals/frostdb/dynparquet"
 	"github.com/prometheus/common/model"
-	"github.com/prometheus/prometheus/model/labels"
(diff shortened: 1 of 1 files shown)


metalmatze:
Seems like the improvements I was hoping for over what we had on `main` don't really make a big difference downstream. So closing this.

PR #2220, opened 2022-12-05
Use scrape interval correctly

fabxc:
Use the scrape interval rather than the timeout to determine the profile duration on scraping. Since taking the profile will take strictly longer than the duration the timeout must be allowed to be larger as well. By default 3s are added to the interval based on real-world obvservations.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -254,13 +254,13 @@ func (c *ScrapeConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 
 	// Validate the scrape and timeout internal configuration. When /debug/pprof/profile scraping
 	// is enabled we need to make sure there is enough time to complete the scrape.
-	if c.ScrapeTimeout > c.ScrapeInterval {
-		return fmt.Errorf("scrape timeout must be smaller or equal to inverval for: %v", c.JobName)
+	if c.ScrapeTimeout == 0 {
+		c.ScrapeTimeout = c.ScrapeInterval + model.Duration(3*time.Second)
+	}
+	if c.ScrapeTimeout <= c.ScrapeInterval {
+		return fmt.Errorf("scrape timeout must be greater than the interval: %v", c.JobName)
 	}
 
-	if c.ScrapeTimeout == 0 {
-		c.ScrapeTimeout = c.ScrapeInterval
-	}
 	if cfg, ok := c.ProfilingConfig.PprofConfig[pprofProcessCPU]; ok {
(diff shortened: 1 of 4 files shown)


brancz:
We just need you to sign the CLA (it's really just a one-time DCO), then this lgtm! Thank you!

brancz:
Looks like we have a test that also needs to be adapted.

pkg/config/config.go
```diff
@@ -250,13 +250,13 @@ func (c *ScrapeConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 
 	// Validate the scrape and timeout internal configuration. When /debug/pprof/profile scraping
 	// is enabled we need to make sure there is enough time to complete the scrape.
```
maxbrunet, line 252: This comment would be wrong

pkg/config/config.go
```diff
@@ -250,13 +250,13 @@ func (c *ScrapeConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
-	if c.ScrapeTimeout > c.ScrapeInterval {
-		return fmt.Errorf("scrape timeout must be smaller or equal to inverval for: %v", c.JobName)
-	}
-
 	if c.ScrapeTimeout == 0 {
-		c.ScrapeTimeout = c.ScrapeInterval
+		c.ScrapeTimeout = c.ScrapeInterval + model.Duration(3*time.Second)
 	}
+	if c.ScrapeTimeout <= c.ScrapeInterval {
+		return fmt.Errorf("scrape timeout must be greater than the interval: %v", c.JobName)
```
maxbrunet, line 257: hmm I think the timeout was required to be shorter to avoid potentially still being scraping the target when we start the next interval

It is the same logic in Prometheus: https://github.com/prometheus/prometheus/blob/v2.40.5/config/config.go#L293-L295

maxbrunet: An "infinitely" large timeout could be configured, resulting in scrapes pilling up on unresponsive targets until exhausting resources on the server

pkg/config/config.go
fabxc: How so? Because it's not only applying this for `/profile`?

maxbrunet: Sorry, I miss read, bad case of Mondays :sweat_smile:

pkg/config/config.go
maxbrunet: Ah I see, nevermind, sounds about right, please carry on :slightly_smiling_face:

metalmatze:
Awesome! :tada: I started CI.

manojVivek:
@all-contributors Please add @fabxc for code.

PR #2227, opened 2022-12-06
ui/packages/app/web: Hide agent or scraper target sections if non-existing

metalmatze:
Previously we still showed the Parca Agents section, even though that Parca never received a single request from a Parca Agent. 
before

By checking if there are any agents in the array we can hide the section entirely if there are none.

after


kakkoyun (approved):
LGTM

PR #2230, opened 2022-12-06
pkg/debuginfo: Cache positive exists lookups

brancz:
Previously we issued lots of repeated lookups in object storage. Since this data is immutable, once this result is positive, it will never change again, so we can cache the response forever.

How we got here: I noticed that our object storage bill was about ~50% class A operations (the other ~50% are egress network traffic). A quick Prometheus query revealed that 92% of all object storage requests are `iter` requests by the debuginfo API:

Screenshot 2022-12-06 at 10 51 31

Then I queried some continuous profiling data to discover which stacks are spending the most CPU time (therefore most often called) that call the iter function (screenshot is from Polar Signals Cloud):

Screenshot 2022-12-06 at 10 50 42

This shows clearly that the majority of times the `Exists` gRPC call calls the `iter` function, more specifically even the `find` function that I added caching to in this patch. So I checked what we can do about that, and it turns out all the positive answers can be cached therefore saving what I believe must be the overwhelming majority of API calls.

Once merged I can observe this for a day and check whether we get the saving that I anticipate.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,6 +25,7 @@ require (
 	github.com/grpc-ecosystem/go-grpc-middleware/v2 v2.0.0-rc.3
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.14.0
+	github.com/hashicorp/golang-lru/v2 v2.0.1
 	github.com/ianlancetaylor/demangle v0.0.0-20220517205856-0058ec4f073c
 	github.com/improbable-eng/grpc-web v0.15.0
 	github.com/klauspost/compress v1.15.12
--- a/pkg/debuginfo/store.go
+++ b/pkg/debuginfo/store.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
+	lru "github.com/hashicorp/golang-lru/v2"
 	"github.com/nanmu42/limitio"
 	"github.com/thanos-io/objstore"
 	"github.com/thanos-io/objstore/client"
@@ -79,6 +80,8 @@ type Store struct {
 
 	metadata         MetadataManager
 	debuginfodClient DebugInfodClient
+
+	existsCache *lru.ARCCache[string, struct{}]
 }
 
 // NewStore returns a new debug info store.
@@ -90,6 +93,11 @@ func NewStore(
 	bucket objstore.Bucket,
 	debuginfodClient DebugInfodClient,
 ) (*Store, error) {
+	existsCache, err := lru.NewARC[string, struct{}](100_000)
+	if err != nil {
+		return nil, err
+	}
+
 	return &Store{
 		tracer:           tracer,
 		logger:           log.With(logger, "component", "debuginfo"),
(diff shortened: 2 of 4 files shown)


kakkoyun (approved):
LGTM 🚀

pkg/debuginfo/store.go
```diff
@@ -97,6 +101,9 @@ func NewStore(
 		cacheDir:         cacheDir,
 		metadata:         metadata,
 		debuginfodClient: debuginfodClient,
+
+		existsCacheMtx: &sync.RWMutex{},
+		existsCache:    map[string]struct{}{},
```
javierhonduco, line 106: I'm worried about these caches being unbounded. While it's not very likely that the Agent will run for an extremely long time without restarts, could we set a limit on it with perhaps an LRU eviction policy?

pkg/debuginfo/store.go
```diff
@@ -382,7 +389,16 @@ func validateInput(id string) error {
...
+	_, found := s.existsCache[key]
+	s.existsCacheMtx.Unlock()
+
+	if found {
+		// We can't cache the negative result, because the file might be
+		// uploaded in the meantime.
+		return true, nil
+	}
+
 	err := s.bucket.Iter(ctx, key, func(_ string) error {
```
javierhonduco, line 402: Perhaps I'm missing something but why don't we use `get` (not sure what the name is) instead?

brancz: What would we get? The `metadata.json` or the debuginfo file? This essentially just checks that _anything_ is there.

javierhonduco: I would suggest the debuginfo file, as it is what we care about the most. We handle a missing metadata file, but a missing debuginfo would be pretty bad

javierhonduco: From our cloud provider:
- class A: storage.objects.list
- class B: storage.*.get

Class B seems like it's 10x cheaper 😄 
https://cloud.google.com/storage/pricing

pkg/debuginfo/store.go
javierhonduco: Mmm, let's maybe do 10k? Sorry if I spoke too soon:

Assuming a key size of 20 characters, that hashmaps have no overhead, not counting for the locking, and that strings take 8 bytes per char (not really true):

(20 characters * 8 bytes per char + 1 byte for the value )* 100k elements = ~16MB

pkg/debuginfo/store.go
kakkoyun: TIL :)

go.mod
```diff
@@ -25,6 +25,7 @@ require (
 	github.com/grpc-ecosystem/go-grpc-middleware/v2 v2.0.0-rc.3
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.14.0
+	github.com/hashicorp/golang-lru/v2 v2.0.1
```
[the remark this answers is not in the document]
brancz: Yeah, I considered using goburrow/cache, but it caches the result regardless of the result, so we would need to error when an object is not found to prevent the result from being cached. I didn't think that made sense so I preferred to use a cache directly like this.

PR #2247, opened 2022-12-08
fix query bar spacing bug

### LINKED ISSUE #2243: [ui] Query bar UI gets squished on smaller width screens
If the user enters compare mode or is working on a smaller width screen, the top bar (query, filters, search) the query bar gets squished entirely. This should not be the case.

image

monicawoj:
resolves #2243


monicawoj:
addresses #2243

manojVivek (approved):
Looking good!

PR #2265, opened 2022-12-12
ui: Flamegraph tooltip performance fix

manojVivek:
Since this is an interaction-based use case, was unable to use the benchmarking tool, but here is the chrome fps meter readings for comparison. 

FPS meter before:
(~20fps)

https://user-images.githubusercontent.com/1283424/206965601-6efe96c4-e09f-4839-8bd3-2a98e61cbab1.mov

FPS meter after:
(~45fps)

https://user-images.githubusercontent.com/1283424/206965782-3503b1e7-d531-4e01-9910-452d8c6ffbc2.mov


brancz (approved):
Love when relatively small changes have huge effect like this!

PR #2283, opened 2022-12-13
parcacol: change stacktrace column to plain encoding

asubiotto:
The stacktrace column was previously encoded as a RLE_DICTIONARY. This has been seen to cause memory bloat since the cardinality of stacktrace values is quite high.


asubiotto:
cc @brancz @thorfour

PR #2285, opened 2022-12-13
ui: Add an `onError` callback to ParcaContext

yomete:
This is a precursor for determining when an expired token is used to query for profiles in the cloud app.


yomete:
Merging this as soon as it's green so I can test that it works in the cloud app.

ui/packages/shared/profile/src/ProfileMetricsGraph/index.tsx
```diff
@@ -89,13 +89,15 @@ const ProfileMetricsGraph = ({
...
   const isLoaderVisible = useDelayedLoader(isLoading);
-  const {loader} = useParcaContext();
+  const {loader, onError} = useParcaContext();
 
   if (isLoaderVisible) {
     return <>{loader}</>;
   }
 
   if (error !== null) {
+    onError?.metricsGraphError(error);
```
manojVivek, line 99: @yomete Maybe we should move this call into an effect on `error` variable, as this will invoke the callback for every render of the component.

yomete:
Thanks for the suggestions @manojVivek I'll make the modifications in a new PR.

PR #2293, opened 2022-12-14
ui: Make the `onError` callback more generic

yomete:


yomete:
@manojVivek here are the suggested changes from the previous [PR](https://github.com/parca-dev/parca/pull/2285)

PR #2299, opened 2022-12-15
*: Flamegraph data trimming

### LINKED ISSUE #2269: Trim the graph to eliminate stack traces that have insignificant value.
Reference to the default trim parameters used in pprof: https://github.com/google/pprof/blob/c488b8fa1db3fa467bf30beb5a1d6f4f10bb1b87/internal/driver/config.go#L65-L66

manojVivek:
- [x] Trimming nodes that are below the threshold
~~- [ ] Garbage collect the values corresponding to the trimmed nodes from the tables~~

Fixes #2269


proto/parca/query/v1alpha1/query.proto
```diff
@@ -224,6 +224,9 @@ message QueryRequest {
 
   // filter_query is the query string to filter the profile samples
   optional string filter_query = 6;
+
+  // disable_trimming disables trimming of insignificant nodes in the report
+  optional bool disable_trimming = 7;
```
brancz, line 229: if we allow disabling it in the API at all, then I think we should allow passing the exact parameters to trim with

manojVivek: Yeah right, you mean the `NodeCutOffFraction` value, right?

manojVivek: This is done.

pkg/query/flamegraph_table.go
```diff
@@ -475,3 +481,85 @@ func equalsByNameTable(tables TableGetter, a, b *querypb.FlamegraphNode) bool {
...
+	total := graph.Total
+
+	threshold := int64(float64(thresholdRate) * float64(total))
+	var children FlamegraphChildren = trimFlamegraphNodes(ctx, tracer, graph.Root.Children, threshold)
+	newTotal := int64(0)
+	newDiff := int64(0)
+	if len(graph.Root.Children) > 0 {
+		newTotal = children.Cumulative()
+		newDiff = children.Diff()
+	} else {
```
brancz, line 518: I don't understand this else statement. If there are no children in the root node it should return 0 anyway, so this else should not be needed?

pkg/query/flamegraph_table.go
```diff
@@ -475,3 +481,85 @@ func equalsByNameTable(tables TableGetter, a, b *querypb.FlamegraphNode) bool {
...
+		return nil
+	}
+	total := graph.Total
+
+	threshold := int64(float64(thresholdRate) * float64(total))
+	var children FlamegraphChildren = trimFlamegraphNodes(ctx, tracer, graph.Root.Children, threshold)
+	newTotal := int64(0)
+	newDiff := int64(0)
+	if len(graph.Root.Children) > 0 {
+		newTotal = children.Cumulative()
```
brancz, line 516: I think we should include the difference in cumulative due to node trimming in the proto response. I think it would be very confusing to click on a sample on a graph with let's say 100 CPU samples, and then only see 85 cumulative in the root node.

proto/parca/query/v1alpha1/query.proto
brancz: What if we always set the `node_trim_threshold` in the UI and if it's unset (or the default value of 0.0) then it's disabled?

proto/parca/query/v1alpha1/query.proto
```diff
@@ -224,6 +224,12 @@ message QueryRequest {
...
+
+  // node_trim_threshold is the threshold for trimming insignificant nodes in the report
```
brancz, line 231: From this comment it is unclear to me that this is a fraction or percentage. By this comment I would expect this to be a total value, which I don't think is true.

ui/packages/app/web/src/components/ui/UserPreferences/UserPreferenceItem.tsx
```diff
@@ -0,0 +1,82 @@
...
+        <input
+          className="bg-inherit border disabled:cursor-not-allowed"
+          id={id}
+          type="checkbox"
+          checked={value}
+          onChange={e => onChange(e.target.checked)}
+          disabled={disabled}
+        />
+      );
+    case 'number':
```
brancz, line 38: interesting idea, but I think this would break shareability, if you have a different configuration from me and you share a link with me I will see different data, if we expose this value (which I don't think we should today), then we should make it part of exploring the data (and therefore part of the URL)

proto/parca/query/v1alpha1/query.proto
manojVivek: Yes, that sounds simpler. Will do that.

pkg/query/flamegraph_table.go
manojVivek: Hmm. True, I was using this for the `FlamegraphNode` type where if there is no `Children` would mean leaf node and will have to take its own value. Sounds unnecessary for the Root node, will make it `0`.

pkg/query/flamegraph_table.go
manojVivek: In that case, how do we account for the difference in the Child nodes? As the sum of child nodes would still only be 85 right?

ui/packages/app/web/src/components/ui/UserPreferences/UserPreferenceItem.tsx
manojVivek: Yeah ok, and we can make this a value added to the URL whenever the user is in the ProfileExplorer and the query formed with the value from the URL taking higher precedence over the value from the preference so that the shareability is maintained. 

But I would prefer moving this into a separate issue and implement it once #2115 is merged. As it has some nice utilities for dealing with the URL params state. Any concerns?

pkg/query/flamegraph_table.go
brancz: Sorry I meant ultimately the UI should show something like "Showing 85 out of 100 samples"

brancz: So actually I meant that we should include the previous and trimmed total value in the response

ui/packages/app/web/src/components/ui/UserPreferences/UserPreferenceItem.tsx
brancz: I would suggest then let's just use a hard-coded value for a start. (I'm not even sure it ever doesn't need to be)

pkg/query/flamegraph_table.go
manojVivek: Got it.

manojVivek: Is `samples` the right terminology for all profile types? 
Also would it be the right place to show this message or should we hide it under some `i` icon?

Screenshot 2022-12-21 at 12 01 21 PM

ui/packages/app/web/src/components/ui/UserPreferences/UserPreferenceItem.tsx
manojVivek: I tried to see if I can implement the URL state without that PR, but it looks like a lot of duplicate work between these two. So, I'll hold it off.

So, if I understand it correctly, meanwhile, you would like this option removed from the user preference and work with just the hardcoded `0.005`cut-off value?

proto/parca/query/v1alpha1/query.proto
manojVivek: This is done.

pkg/query/flamegraph_table.go
manojVivek: This is done.

ui/packages/app/web/src/components/ui/UserPreferences/UserPreferenceItem.tsx
manojVivek: This is done.

brancz (approved):

PR #2325, opened 2022-12-21
ui: Fix displaying targets error and handle scraper-only mode correctly

brancz:
In scraper-only mode the agents gRPC service is not served, therefore returning `UNIMPLEMENTED`, which causes the UI not to be able to render the reply for the targets being scraped.

Additionally this patch also actually renders the errors that occur to allow for better troubleshooting when they do occur.


metalmatze (approved):
Oh, nice catch!

PR #2393, opened 2023-01-10
ui: Fix minor CSS bugs

yomete:
This fixes two minor CSS bugs.

We've had this weird overlay whenever selecting the profile types dropdown. 
image

This happens because the profile types are not yet returned from the API but the dropdown is still trying to display the items, and the element responsible for displaying the items happens to have a border which is why we see that line. To fix that I changed the logic so that the element responsible for showing the dropdown items is visible only when we have data from the API.

Secondly, I adjusted the checkbox + label in the user preferences modal to be aligned at the start of the flex.
**Before**
image

**After**
image


manojVivek (approved):
✨

PR #2398, opened 2023-01-11
fix height of button section of input element

### LINKED ISSUE #2397: [ui] fix styling of "filter by function" button (Firefox)
CSS issue in the "Filter by function" button where the arrow button portion of the input element does not match the height of the rest of the input element

The issue seems to only be present in Firefox.

monicawoj:
resolves #2397


yomete (approved):
lgtm!

PR #2403, opened 2023-01-12
Visualization Panels drag n drop feature

monicawoj:
Allows you to drag and drop the different visualization views (switch from right to left and vice-versa) when there are 2 panels. The URL is also updated when the positions of the panels are swapped.


yomete (approved):
sooo cool! 🚢 🇮🇹

manojVivek:
Monica, looks like there is a bug when I drag a panel and drop it in the same position. It still swaps them. 

https://user-images.githubusercontent.com/1283424/212598082-fb58efba-deb1-4549-89a4-99024775650c.mov

monicawoj:
> Monica, looks like there is a bug when I drag a panel and drop it in the same position. It still swaps them.
> 
>  Screen.Recording.2023-01-16.at.9.49.38.AM-1.mov

Thanks for catching this, Manoj! I'll look into it!

brancz:
Looooove it! lgtm from a user perspective!

Teeny tiny nit: Since each panel already has the X to close it, I don't think the reset button is needed. I don't feel strongly at all though, it would just be one less thing on the screen.

ui/packages/shared/profile/src/ProfileView/index.tsx
```diff
@@ -263,17 +263,81 @@ export const ProfileView = ({
...
+                                    snapshot.isDragging ? 'bg-gray-200' : 'bg-white'
+                                  )}
+                                >
+                                  <div className="w-full flex justify-end pb-2">
+                                    <div className="w-full flex justify-between">
+                                      <div
+                                        className={cx(isMultiPanelView ? 'visible' : 'invisible')}
+                                        {...provided.dragHandleProps}
+                                      >
+                                        <DragIndicator />
```
manojVivek, line 308: Shall we replace this icon with the [dots drag handle icon](https://icones.netlify.app/collection/all?s=material-symbols:drag-indicator)? `<Icon icon="material-symbols:drag-indicator" />`



I feel that icon is more universal, and this one looks a bit confusing, like the hamburger menu.

monicawoj: > 

Yesss! That's a great point. I think the reset button is extraneous in this case - will remove it

monicawoj:
> Looooove it! lgtm from a user perspective!
> 
> Teeny tiny nit: Since each panel already has the X to close it, I don't think the reset button is needed. I don't feel strongly at all though, it would just be one less thing on the screen.

Yesss! That's a great point. I think the Reset button doesn't add any new functionality here. Will remove it

PR #2423, opened 2023-01-18
Flamegraph trimming bugfix, Disabling flamegraph trimming by default & fixed the compare legend visibility

manojVivek:


yomete (approved):

pkg/query/flamegraph_table.go
```diff
@@ -543,15 +543,14 @@ func trimFlamegraphNodes(ctx context.Context, tracer trace.Tracer, nodes []*quer
 		if node.Cumulative < threshold {
 			continue
 		}
+		var oldChildren FlamegraphChildren = node.Children
```
asubiotto, line 546:
```suggestion
		oldChildren := node.Children
```

pkg/query/flamegraph_table.go
```diff
@@ -543,15 +543,14 @@ func trimFlamegraphNodes(ctx context.Context, tracer trace.Tracer, nodes []*quer
...
+		flat := node.Cumulative - oldChildren.Cumulative()
 		var children FlamegraphChildren = trimFlamegraphNodes(ctx, tracer, node.Children, threshold)
```
asubiotto, line 548:
```suggestion
		children := trimFlamegraphNodes(ctx, tracer, node.Children, threshold)
```

pkg/query/flamegraph_table.go
manojVivek: I was trying to do this but was unable to as `node.Children` ie `[]*querypb.FlamegraphNode` is not an interface. 

So, I created an alias type `FlamegraphChildren` and added some utility functions like below:

```
type FlamegraphChildren []*querypb.FlamegraphNode

func (n FlamegraphChildren) Cumulative() int64 {
	cumulative := int64(0)
	for _, child := range n {
		cumulative += child.Cumulative
	}
	return cumulative
}
```

I was unable to typecast and access these utility functions, so defined a variable with this type and used those for accessing the functions. 

There could a different way for this problem, please let me know.

asubiotto: ah, then what you have is fine, another option is:
```
oldChildren := FlamegraphChildren(node.Children)
```

manojVivek: Cool, thanks for the tip!

PR #2444, opened 2023-01-19
pkg/profile: Remove pointer

brancz:
This pointer was causing about 15% of CPU time of the query path purely because it needs to be allocated on the heap.


kakkoyun (approved):
LGTM

brancz:
I think there are ways to save even more if keeping this pointer. Going to close this for now.

PR #2456, opened 2023-01-22
chore: switch otel-collector image to release channel

maxbrunet:
The upstream automation had been broken for months, but now that it is fixed (open-telemetry/opentelemetry-collector-contrib@1c12e7240b9cdddc0c5c7627c9ea1c58f1e26829), an image is published on every commit to the project's `main` branch. This makes Renovate fairly noisy, so I suggest we switch to the release image. It will also give us a better visibility of the changes.

If for some reason the dev image is still preferred, an alternative would be to set its updates to a weekly schedule.


maxbrunet:
Merging to avoid merge conflicts with Renovate and stop noise. Please feel free to comment and follow up :slightly_smiling_face:

PR #2471, opened 2023-01-24
fix: change Reset zoom button text to be Reset view

yomete:


brancz:
If we're already touching this, can we have it aligned with the visualization-chooser-dropdown?

brancz (commented):
Love it!

PR #2497, opened 2023-01-30
pkg/debuginfo: Fix instantiation of upstream debuginfod servers

brancz:
Turns out our debuginfod integration never worked ...


javierhonduco (approved):
This explains a lot of things!! 😄

PR #2381, opened 2023-01-09
flamegraph: Binary based color stack

manojVivek:
Resolves #2145.

PR #2501, opened 2023-01-30
ui: Add a button to clear search string after clicking on a table row

yomete:
1. A clear selection button will now show when a user clicks on a table row to select nodes on other visualisations. 
2. The clear selection button only appears when there are two dashboard visualizations next to each other.

image

image


brancz:
Might not be a bug that was introduced on this PR, but when I click on the table it doesn't appear to highlight the correct nodes in the icicle graph.

Screenshot 2023-01-31 at 11 13 27

Otherwise love this addition!

yomete:
> Might not be a bug that was introduced on this PR, but when I click on the table it doesn't appear to highlight the correct nodes in the icicle graph.
> 
> Screenshot 2023-01-31 at 11 13 27
> 
> Otherwise love this addition!

interesting. i'll take a look at that.

monicawoj:
> > Might not be a bug that was introduced on this PR, but when I click on the table it doesn't appear to highlight the correct nodes in the icicle graph.
> > Screenshot 2023-01-31 at 11 13 27
> > Otherwise love this addition!
> 
> interesting. i'll take a look at that.

@yomete not sure, but this may be due to the URL state changes I introduced a while back, would be good to start by checking to see if the table is using the search string value from the URL or the local state value... If you run into any questions/problems let me know!

manojVivek:
> @yomete not sure, but this may be due to the URL state changes I introduced a while back, would be good to start by checking to see if the table is using the search string value from the URL or the local state value... If you run into any questions/problems let me know!

@yomete I have this bug already fixed in the binary-based colour stacks for flamegraph PR here: #2381.

yomete:
> > @yomete not sure, but this may be due to the URL state changes I introduced a while back, would be good to start by checking to see if the table is using the search string value from the URL or the local state value... If you run into any questions/problems let me know!
> 
> @yomete I have this bug already fixed in the binary-based colour stacks for flamegraph PR here: #2381.

Noice. I'll review that one and merge this after.

ui/packages/shared/profile/src/TopTable/index.tsx
```diff
@@ -142,22 +140,27 @@ export const TopTable = ({
 
   const onRowClick = useCallback(
     (row: TopNode) => {
+      // If there is only one dashboard item, we don't want to select a span
+      if (dashboardItems.length <= 1) {
```
manojVivek, line 144: Should we also not show the pointer cursor for the table items when this condition matches? 
As that gives a wrong impression that it is clickable but nothing happens on click.

yomete: good idea! i'll implement that

manojVivek (approved):

manojVivek (approved):
Lgtm, let's revisit the absolute positioned if it bites back again!

PR #2507, opened 2023-01-31
pkg/parcacol: Handle NULL strings in Values API call

metalmatze:
This previously resulted in random strings showing up in the dropdown... Best case it was an empty string.


brancz (approved):
Ooooooooooo nice find!!

PR #2511, opened 2023-02-01
ui: Fix false positive Profile has no samples message in visualisations

yomete:
This will fix the current behaviour of the visualizations showing the `Profile has no samples` message for half a second before populating the visualisations with data. The code responsible for showing the message didn't take into account the value of the `loading` property. So I added that as a check before displaying that message.

I also removed the `disabled` property from buttons on ProfileView toolbar i.e. the share profile and download pprof buttons. They also flickered whenever selecting a new profile mostly because we had the disabled property set to the status of the data being loaded.

**demo.parca.dev**
demo

This **PR**
pr

TL:DR No more `Profile has no samples` message and less flickers in the profile view when selecting profiles.

One more thing: I still am not a fan on the very quick flicker that happens even now after this PR, as you can see in the GIF and demo itself, when you click on a profile in the metric graph, the viz still goes empty before showing the data. I tried to find where in the code is responsible for that but i was not successful. Any pointers would be appreciated. I'm thinking a loader might be better than the quick flicker. Or what do y'all think?

### DIFF
--- a/ui/packages/shared/profile/src/IcicleGraph.tsx
+++ b/ui/packages/shared/profile/src/IcicleGraph.tsx
@@ -26,12 +26,17 @@ import {
 } from '@parca/client/dist/parca/metastore/v1alpha1/metastore';
 import type {HoveringNode} from './GraphTooltip';
 import GraphTooltip from './GraphTooltip';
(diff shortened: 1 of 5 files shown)


brancz (approved):
🥳

brancz (approved):
Niiiiice works flawlessly!

PR #2516, opened 2023-02-02
Merges with Step Duration instead of Single Profile requests

### LINKED ISSUE #2289: [ui] Add support for step in querying
As is, when queries are made for a large duration, we are receiving several thousand data points for a given query. In order to reduce load and make for a more pleasant UI / visualizations, we would like to introduce a "step" query parameter that will dictate the interval between points. The step duration will be calculated on the frontend, based on the start and end timestamp to query for a desired amount of points (1000), as opposed to an endless amount of points each at a specified interval. 

Once the changes for Parca’s backend have been pushed to the "step" branch - the following items need to be introduced in the UI:
- [x] (Internally) Replace usage of SingleProfileSelection with MergeProfileSelection
- [x] Dynamically show the merge button based on the delta or not (whether [ProfileTypeResponse](https://buf.build/parca-dev/parca/docs/main:parca.query.v1alpha1#parca.query.v1alpha1.ProfileType) returns boolean). Show if delta boolean is true.
- [x] Calculate the step in a refactored helper function to use across the project consistently. We want to calculate the step duration based on a desired 1000 points
- [x] Once we click on point in the metrics graph: For delta types: send a merge request with the starting timestamp as the clicked point’s timestamp and the end timestamp to the start timestamp + step duration (in seconds). For non-delta types: Send a merge request with the start timestamp as the clicked point’s timestamp and the end timestamp being the same (essentially a single profile)
- [x] Adapt the URL state to reflect these new UI states: Never hide the Metrics Graph if merge=true in the URL. Every merge request should set merge to true. Next to the metrics start and end timestamp the profile itself will have a start and end timestamp (which is the start+step)
- [x] [Update as of 1 Feb 2023]: there will be no merge button, independent of delta and non-delta. 
- [x] [Update as of 1 Feb 2023]: For delta profiles, run merge immediately where the start and end time of the query matches the metrics graph time span. Show icicle graph of the merged result. 
- [x] [Update as of 1 Feb 2023]: For non-delta profile types, do not automatically run a merge request. Merges with different start and end times do not make sense for non-delta profiles. Only show the metrics graph. Once the user clicks on a point, the merge request is run with start and end time equal to the timestamp of the clicked point.

monicawoj:
resolves #2289


metalmatze:
Seems like merging is now blocked on 
```
Error: src/ProfileExplorer/index.tsx(80,9): error TS2552: Cannot find name 'error'. Did you mean 'onError'?
Error: src/ProfileExplorer/index.tsx(80,32): error TS2552: Cannot find name 'error'. Did you mean 'onError'?
Error: src/ProfileExplorer/index.tsx(81,17): error TS2304: Cannot find name 'error'.
Error: src/ProfileExplorer/index.tsx(83,7): error TS2552: Cannot find name 'error'. Did you mean 'onError'?
Error: src/ProfileExplorer/index.tsx(155,7): error TS2552: Cannot find name 'error'. Did you mean 'onError'?
Error: src/ProfileExplorer/index.tsx(155,30): error TS2552: Cannot find name 'error'. Did you mean 'onError'?
Error: src/ProfileExplorer/index.tsx(162,44): error TS2304: Cannot find name 'error'.
```

ui/packages/shared/functions/src/time.ts
```diff
@@ -11,7 +11,109 @@
...
+export const unitsInTime = {
+  [TimeUnits.Nanos]: {multiplier: 1, symbol: 'ns'},
+  [TimeUnits.Micros]: {multiplier: 1e3, symbol: 'µs'},
+  [TimeUnits.Milliseconds]: {multiplier: 1e6, symbol: 'ms'},
+  [TimeUnits.Seconds]: {multiplier: 1e9, symbol: 's'},
+  [TimeUnits.Minutes]: {multiplier: 6 * 1e10, symbol: 'm'},
+  [TimeUnits.Hours]: {multiplier: 60 * 60 * 1e9, symbol: 'h'},
+  [TimeUnits.Days]: {multiplier: 60 * 60 * 24 * 1e9, symbol: 'd'},
+  [TimeUnits.Weeks]: {multiplier: 60 * 60 * 24 * 7 * 1e9, symbol: 'w'},
+  [TimeUnits.Years]: {multiplier: 60 * 60 * 24 * 365 * 1e9, symbol: 'y'},
```
metalmatze, line 53: Not sure if worth adding as a comment but wanted it to mention once. The Go team, for example, decided against having weeks and years as constants. Especially the year one is wrong once every 4 years. :grimacing: It works well enough for our use case and Prometheus does the same, just be aware of the fact.

metalmatze (approved):
Fantastic work Monica! :+1: 
Code-wise we should be able to merge this really soon, once CI is green. 
Everything else we should tackle in smaller individual PRs, I think.

ui/packages/shared/functions/src/time.ts
monicawoj: ok, that's good to know! You're right with leap years, it could all be off. Maybe I'll just add a TODO in the code while I'm fixing whats blocking the build so we can flag this and return to it down the road

monicawoj:
> Seems like merging is now blocked on
> 
> ```
> Error: src/ProfileExplorer/index.tsx(80,9): error TS2552: Cannot find name 'error'. Did you mean 'onError'?
> Error: src/ProfileExplorer/index.tsx(80,32): error TS2552: Cannot find name 'error'. Did you mean 'onError'?
> Error: src/ProfileExplorer/index.tsx(81,17): error TS2304: Cannot find name 'error'.
> Error: src/ProfileExplorer/index.tsx(83,7): error TS2552: Cannot find name 'error'. Did you mean 'onError'?
> Error: src/ProfileExplorer/index.tsx(155,7): error TS2552: Cannot find name 'error'. Did you mean 'onError'?
> Error: src/ProfileExplorer/index.tsx(155,30): error TS2552: Cannot find name 'error'. Did you mean 'onError'?
> Error: src/ProfileExplorer/index.tsx(162,44): error TS2304: Cannot find name 'error'.
> ```

yes, this is a tricky issue. working on resolving this!

monicawoj:
@metalmatze all checks are green! but when I click a point in the Vercel deployment the `An error occurred: expected column stacktrace, got 0 columns` appears. Could you take a look? Are we still ok to merge here?

pkg/parcacol/querier.go
```diff
@@ -312,22 +352,27 @@ func (q *Querier) QueryRange(
...
-	timestampColumnIndex := 0
-	timestampColumnFound := false
-	valueColumnIndex := 0
-	valueColumnFound := false
+	type columnIndex struct {
+		index int
+		found bool
+	}
+	// Add necessary columns and their found value is false by default.
+	columnIndices := map[string]columnIndex{
```
brancz, line 360: @metalmatze how come this is not a struct? I get the convenience of the map access below, but I think we give up compile-time checks that we're actually accessing the correct fields, and not fields that may be unexpected. A switch/case setting the appropriate field of a struct would allow for less mistakes I think.

ui/packages/shared/profile/src/ProfileMetricsGraph/index.tsx
```diff
@@ -13,22 +13,23 @@
...
 interface ProfileMetricsGraphProps {
   queryClient: QueryServiceClient;
   queryExpression: string;
   profile: ProfileSelection | null;
   from: number;
   to: number;
-  select: (source: ProfileSelection) => void;
   setTimeRange: (range: DateTimeRange) => void;
   addLabelMatcher: (key: string, value: string) => void;
+  onPointClick: (timestamp: any, labels: any, queryExpression: string) => void;
```
yomete, line 32: If we can add the appropriate types to the props here, that'd be great.

yomete (approved):

metalmatze:
The calculations now look correct as in `percentage of CPU` spend. 
parca
This can be roughly compared with Prometheus metrics too. 
prometheus

Essentially, before I had the `1000 *` thrown into the code but the `100 *` is actually correct to get to percentages between 0-100.

brancz (approved):
I'm happy with merging as is as long as we make the additional fix we discussed @metalmatze.

This would mean that the query range API from now on will return two values:

1) The current non-normalized total value (as the flamegraph will also have it in its root.
2) The normalized per-second value as calculated in this PR.

The UI will then need to display both of these values as well for delta-profiles.

ui/packages/shared/profile/src/ProfileMetricsGraph/index.tsx
monicawoj: true, those anys were meant to be temporary! I'll change that

metalmatze:
Opened #2596, #2597, and #2598 as follow-ups.

pkg/parcacol/querier.go
```diff
@@ -381,12 +388,6 @@ func (q *Querier) queryRangeDelta(ctx context.Context, filterExpr logicalplan.Ex
 		}
 	}
 
-	for name, index := range columnIndices {
```
brancz: I think this check was useful. Maybe we pre-populate the indexes with `-1` to check if we have found an index for each of them. Otherwise we silently err on the side of using index 0.

metalmatze: I had *int before but the -1 is nicer. Will update.

brancz (approved):

PR #2520, opened 2023-02-03
pkg/parca: Add e2e test for Go PGO

brancz:
Add a test to ensure that the data exported by Parca can be used by Go's newly added profile-guided optimizations (PGO).

While testing this I realized that the functionality currently expects the samples count to be on index position 1 (understandable since that's where the runtime-generated profiles have it). But Parca takes the profiles apart and only returns a single sample type, therefore it currently panics, which is why this patch includes the small fix for the pprof profile. An issue has already been filed upstream: https://github.com/golang/go/issues/58292

### DIFF
--- a/pkg/parca/parca_test.go
+++ b/pkg/parca/parca_test.go
@@ -14,12 +14,15 @@
 package parca
 
 import (
+	"bytes"
 	"compress/gzip"
 	"context"
 	"crypto/tls"
 	"errors"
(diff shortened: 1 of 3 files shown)


metalmatze (approved):

kakkoyun (approved):
LGTM.

I think we just need to appease the linter and we should merge it.

Issue #2540, opened 2023-02-06
ui: Upgrade to Node 18


yomete:
Parca fails to compile on the default node version (18) on macOS via Homebrew. It's currently set to Node 16

metalmatze:
Here's why it fails:
```
error eslint-plugin-jsdoc@38.0.6: The engine "node" is incompatible with this module. Expected version "^12 || ^14 || ^16 || ^17". Got "19.6.0"
error Found incompatible module.
```

maxbrunet:
There are a lot of major updates available for JS deps, see [dashboard](https://togithub.com/parca-dev/parca/issues/1226).
Some are likely related to supporting the new LTS 


Although for this one, it appears to be [`@grafana/toolkit`](
https://github.com/grafana/grafana/blob/v9.3.6/packages/grafana-toolkit/package.json#L85):

```console
$ yarn why eslint-plugin-jsdoc
yarn why v1.19.0
[1/4] Why do we have the module "eslint-plugin-jsdoc"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
info 
=> Found "eslint-plugin-jsdoc@38.0.6"
info Reasons this module exists
   - "_project_#@parca#grafana-datasource#@grafana#toolkit" depends on it
   - Hoisted from "_project_#@parca#grafana-datasource#@grafana#toolkit#eslint-plugin-jsdoc"
   - Hoisted from "_project_#@parca#grafana-panel#@grafana#toolkit#eslint-plugin-jsdoc"
   - Hoisted from "_project_#@parca#grafana-datasource#@grafana#toolkit#@grafana#eslint-config#eslint-plugin-jsdoc"
Done in 1.17s.
```

Made a quick PR grafana/grafana#62979

maxbrunet:
Also I remember this [mhart/alpine-node](https://github.com/mhart/alpine-node) image is not maintained:

https://github.com/parca-dev/parca/blob/dbc7a0630b6968f99b24dda812600e7d22f204a0/ui/Dockerfile.dev#L1

Should we change it for the official one?

maxbrunet:
Done in #2563

PR #2543, opened 2023-02-06
chore(deps): use official nodejs image for development

maxbrunet:
[mhart/alpine-node#deprecated](https://github.com/mhart/alpine-node#deprecated)

(The `# vim: ft=dockerfile` comment adds syntax highlighting in GitHub)

Related to #2540


brancz (approved):
I don’t even remember anymore why we ever used that base image.

PR #2567, opened 2023-02-09
ui: Added screen with info when no data is received on the server

### LINKED ISSUE #2514: When there is no data, modify the UI to show some pointers to check if the data ingestion is happening
Screenshot_2023-02-02_at_09 10 02

manojVivek:
Fixes #2514 

Screenshot:
Screenshot_2023-02-09_at_1 08 01_PM

### DIFF
--- a/ui/packages/shared/profile/src/ProfileExplorer/index.tsx
+++ b/ui/packages/shared/profile/src/ProfileExplorer/index.tsx
@@ -11,14 +11,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {QuerySelection} from '../ProfileSelector';
+import {QuerySelection, useProfileTypes} from '../ProfileSelector';
 import {ProfileSelection, ProfileSelectionFromParams, SuffixParams} from '..';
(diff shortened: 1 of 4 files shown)


ui/packages/shared/profile/src/ProfileExplorer/index.tsx
```diff
@@ -126,6 +134,14 @@ const ProfileExplorerApp = ({
     return selectProfile(p, '_b');
   };
 
+  if (profileTypesLoading) {
+    return <>{loader}</>;
+  }
+
+  if (profileTypesData?.types.length === 1) {
```
monicawoj, line 141: just to confirm, do we only want to show the no data prompt when the length === 1? Would `profileTypesData?.types.length` ever return 0 or -1?

manojVivek: My bad, was using 1 for testing and pushing that too. 😐  
Fixing it now.

monicawoj (approved):

PR #2571, opened 2023-02-10
ui: Add a tooltip when a MetricsGraph tooltip label is hovered on.

### LINKED ISSUE #2474: Show full label when hovering
When holding shift and hovering with the mouse over a label it would be nice to show a tooltip or something to be able to see the full value of a label.

Screenshot 2023-01-25 at 10 04 24

In this screenshot I was hovering over the `cgroup_name` label but can't see the full value unless I click it, which will fire off a new query but I don't even know yet if that's what I want to do.

yomete:
Resolves #2474 

label-tooltip-hover

### DIFF
--- a/.github/workflows/codeql-analysis-go.yml
+++ b/.github/workflows/codeql-analysis-go.yml
@@ -74,9 +74,9 @@ jobs:
           cache: yarn
 
       - name: Initialize CodeQL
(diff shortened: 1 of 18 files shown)


manojVivek (approved):
LGTM!

PR #2586, opened 2023-02-14
chore: Remove unused CSS files & sort imports using Prettier

yomete:
This PR:

1. Initially the idea was to convert the CSS files in `ui/packages/shared/profile/src/` to be Tailwind stylings. But we actually don't use these files anymore. So I went ahead to remove the unused CSS files in the `ui/packages/shared/profile/src/` folder.
2. Uses the [@ianvs/prettier-plugin-sort-imports pkg](https://www.npmjs.com/package/@ianvs/prettier-plugin-sort-imports) to automatically sort imports based on the import order [here](https://github.com/parca-dev/parca/pull/2584/files#diff-06653bc733c61efc5323bfdf6e31ae9e0a48fab9044c37cdbb79443f815b9af7R6-R13).


ui/packages/shared/client/src/google/api/http.ts
```diff
@@ -16,16 +16,19 @@
...
-import type { IBinaryWriter } from "@protobuf-ts/runtime";
-import { WireType } from "@protobuf-ts/runtime";
-import type { BinaryReadOptions } from "@protobuf-ts/runtime";
-import type { IBinaryReader } from "@protobuf-ts/runtime";
-import { UnknownFieldHandler } from "@protobuf-ts/runtime";
-import type { PartialMessage } from "@protobuf-ts/runtime";
-import { reflectionMergePartial } from "@protobuf-ts/runtime";
-import { MESSAGE_TYPE } from "@protobuf-ts/runtime";
-import { MessageType } from "@protobuf-ts/runtime";
+import {
```
manojVivek, line 19: I think we should revert the deleted `.prettierignore` file, as we don't need the files in `ui/packages/shared/client` formatted since it's a generated package.

.pre-commit-config.yaml
```diff
@@ -76,3 +76,6 @@ repos:
     hooks:
       - id: prettier
```
manojVivek, line 77: Looks like this prettier hook runs, ignoring the config in `ui/.prettierrc`? 🤔

manojVivek (commented):
Just a mirror inconsistency with the prettier config, looks good otherwise!

ui/packages/shared/client/src/google/api/http.ts
yomete: So it wasn't moved, I only moved it to be in the `ui/` folder as that's where the prettier config, the `.prettierrc`  file lives.

yomete: And the now moved file also config for ignoring the client files.

monicawoj (approved):
lgtm! very glad to have the import sort in place!!

PR #2600, opened 2023-02-15
ui: Flamegraph tooltip improvements

manojVivek:


ui/packages/shared/profile/src/utils.ts
```diff
@@ -50,3 +50,11 @@ export const truncateString = (str: string, num: number): string => {
 
   return str.slice(0, num) + '...';
 };
+
+export const truncateStringReverse = (str: string, num: number): string => {
```
monicawoj, line 54: I think this is a good example of something that could live in the @parca/utils package once that is in place :) cc @yomete

monicawoj:
In general, looks great to me! 

My only concern is the empty whitespace in the tooltip when hovering over the root node (or a node with very little information to show).. I don't assume the user will spending much time hovering over root, but just wanted to bring it up and get some others' thoughts on this.
Screenshot 2023-02-15 at 22 19 33

manojVivek:
> My only concern is the empty whitespace in the tooltip when hovering over the root node (or a node with very little information to show).. I don't assume the user will spending much time hovering over root, but just wanted to bring it up and get some others' thoughts on this.

Yes, I had the concern as well. Should we try listing all fields everywhere and mark that there's `no data` wherever we don't have the info? 

@brancz Let me know what do you think as well.

brancz:
> Yes, I had the concern as well. Should we try listing all fields everywhere and mark that there's no data wherever we don't have the info?

I think this is a good idea.

yomete (approved):

PR #2616, opened 2023-02-16
pkg/parcacol: Translate empty `""` matcher values to NULL

metalmatze:
This feature needs the NULL query support in FrostDB: https://github.com/polarsignals/frostdb/pull/352.

This finally allows once to query things like 

`cgroup_name!=""` in SQL `cgroup_name IS NOT NULL` to show all profiles that run within a cgroup.
`namespace=""` in SQL `namespace IS NULL` to show all profiles that don't belong to a Kubernetes pod.
`systemd_unit!=""` in SQL `systemd_unit IS NOT NULL` shows all profiles that are run as sytemd unit.

cc @kakkoyun who was asking earlier.


kakkoyun (approved):
LGTM ❤️ 💚 💙 💛 💜

brancz (approved):

PR #2637, opened 2023-02-21
pkg/symbolizer: Use symbols liner either HasSymtab or HasDynsym

zdyj3170101136:
continuation of the work in https://github.com/parca-dev/parca/pull/2491

### DIFF
--- a/pkg/symbolizer/symbolizer.go
+++ b/pkg/symbolizer/symbolizer.go
@@ -480,6 +480,9 @@ func (s *Symbolizer) symbolizeLocationsForMapping(ctx context.Context, m *pb.Map
 			if err := s.metadata.SetQuality(ctx, m.BuildId, dbginfo.Quality); err != nil {
 				return nil, nil, fmt.Errorf("set quality: %w", err)
 			}
+			if !dbginfo.Quality.HasDwarf && !dbginfo.Quality.HasGoPclntab && !(dbginfo.Quality.HasSymtab || dbginfo.Quality.HasDynsym) {
+				return nil, nil, fmt.Errorf("check debuginfo quality: %w", ErrNoDebuginfo)
+			}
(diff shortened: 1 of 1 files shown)


pkg/symbolizer/symbolizer.go
```diff
@@ -480,9 +480,6 @@ func (s *Symbolizer) symbolizeLocationsForMapping(ctx context.Context, m *pb.Map
 			if err := s.metadata.SetQuality(ctx, m.BuildId, dbginfo.Quality); err != nil {
 				return nil, nil, fmt.Errorf("set quality: %w", err)
 			}
-			if !dbginfo.Quality.HasDwarf && !dbginfo.Quality.HasGoPclntab && !(dbginfo.Quality.HasSymtab && dbginfo.Quality.HasDynsym) {
```
v-thakkar, line 483: It'll be still nice to keep this check. We can do the same as you've done at other places - we can check for either .symtab or .dynsym instead of both.

v-thakkar (changes requested):
Thanks for the PR, other changes look good to me apart from the one mentioned below.

Also, can you please add more description in the PR about why we're doing this change? For example: here you can also mention that this is in continuation of the amazing work you did in #2491 etc.

pkg/symbolizer/symbolizer.go
```diff
@@ -556,7 +556,8 @@ func (s *Symbolizer) newLiner(filepath string, f *elf.File, quality *debuginfopb
 		}
 
 		return lnr, nil
-	case quality.HasSymtab && quality.HasDynsym:
+		// TODO CHECK plt
```
v-thakkar, line 559: Good point :+1:

v-thakkar (approved):
LGTM :tada:

PR #2638, opened 2023-02-21
ui: Add labels to query when clicking on metrics sample

### LINKED ISSUE #2617: Propagate labelset when clicking on a specific sample in the metrics graph
_Note: This happens since we merged #2516_

When clicking on a specific sample in the metrics graph it shows a profile for _all_ series at that timestamp. 

Screenshot from 2023-02-16 19-52-28

The query in the payload is `parca_agent_cpu:samples:count:cpu:nanoseconds:delta{systemd_unit!=""}`. It's missing the specific series' label set I was interested in. Like `comm="java"` and all the others one can see in the tooltip. 

To fix this we need to propagate the series' entire label set in the query. 

cc @monicawoj

metalmatze:
Closes #2617 

Right now it's missing the label propagation from the input field though. Any ideas how to propagate those?

### DIFF
--- a/ui/packages/shared/profile/src/ProfileSource.tsx
+++ b/ui/packages/shared/profile/src/ProfileSource.tsx
@@ -21,7 +21,7 @@ import {
   Timestamp,
 } from '@parca/client';
 import {formatDate} from '@parca/functions';
-import {ProfileType, Query} from '@parca/parser';
+import {Matcher, ProfileType, Query} from '@parca/parser';
 
 export interface ProfileSource {
   QueryRequest: () => QueryRequest;
@@ -80,13 +80,16 @@ export function ProfileSelectionFromParams(
     mergeTo !== undefined &&
     expression !== undefined
   ) {
-    return new MergedProfileSelection(
-      parseInt(mergeFrom),
-      parseInt(mergeTo),
-      ParseLabels(labels ?? ['']),
-      expression,
-      filterQuery
-    );
+    // TODO: Refactor parsing the query and adding matchers
+    let query = Query.parse(expression);
+    ParseLabels(labels ?? ['']).forEach(l => {
+      const [newQuery, changed] = query.setMatcher(l.name, l.value);
+      if (changed) {
+        query = newQuery;
+      }
+    });
+
+    return new MergedProfileSelection(parseInt(mergeFrom), parseInt(mergeTo), query, filterQuery);
   }
 
   return null;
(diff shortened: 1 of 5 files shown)


ui/packages/shared/profile/src/ProfileSource.tsx
```diff
@@ -56,6 +56,10 @@ export function SuffixParams(params: {[key: string]: any}, suffix: string): {[ke
 }
 
 export function ParseLabels(labels: string[]): Label[] {
+  if (labels.length === 0 || (labels.length === 1 && labels[0] === '')) {
```
monicawoj, line 59: instead of this, you could also strip the labels array of any empty string values and then check its length:

```
const nonEmptyLabels = labels.filter((str) => str !== '');
if (labels.length < 1)
```

ui/packages/shared/profile/src/ProfileSource.tsx
```diff
@@ -190,32 +194,44 @@ export class ProfileDiffSource implements ProfileSource {
...
 
+  query(): string {
+    const seriesQuery =
+      this.profName +
+      this.labels.reduce(
+        (agg: string, label: Label) => agg + `${label.name}="${label.value}",`,
+        '{'
+      );
+    const query = seriesQuery + '}';
+    console.log(query);
```
monicawoj, line 223: can remove console.log :)

ui/packages/shared/profile/src/ProfileSource.tsx
```diff
@@ -239,13 +255,13 @@ export class MergedProfileSource implements ProfileSource {
...
   ProfileType(): ProfileType {
-    return ProfileType.fromString(Query.parse(this.query).profileName());
+    return ProfileType.fromString(Query.parse(this.query()).profileName());
   }
 
   Describe(): JSX.Element {
     return (
       <a>
-        Merge of &quot;{this.query}&quot; from {formatDate(this.mergeFrom, timeFormat)} to{' '}
+        Merge of &quot;{this.query()}&quot; from {formatDate(this.mergeFrom, timeFormat)} to{' '}
```
monicawoj, line 264: can be written using template literals (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

`Merge of "${this.query()}" from ${formatDate(this.mergeFrom, timeFormat)} to...`

monicawoj: using `

ui/packages/shared/profile/src/ProfileSource.tsx
metalmatze: Like that much better! :+1:

brancz (commented):
I don't think this strategy is quite right. What we need to do is parse the query we sent to the metrics graph query, using the parser, and then call `setMatcher` for each of the labels known from the metrics sample: https://github.com/parca-dev/parca/blob/7a07bf57f7092b0ee9c1b742c495229ce1231ee7/ui/packages/shared/parser/src/index.ts#L385

Then we can use the resulting query for the query to the `Query` endpoint.

metalmatze:
Looking into using the `setMatchers` from the MetricsGraph returned `Label[]`. That should be way better indeed.

brancz (approved):

PR #2640, opened 2023-02-21
ui: QueryRange MetricSample returns double value

### LINKED ISSUE #2596: QueryRange should return two values
This would mean that the query range API from now on will return two values:

1) The current non-normalized total value (as the flame graph will also have it in its root.
2) The normalized per-second value as calculated in this PR.

The UI will then need to display both of these values as well for delta profiles.

_Originally posted by @brancz in https://github.com/parca-dev/parca/pull/2516#pullrequestreview-1298370649_

metalmatze:
For delta profiles, the percentage is calculated. However, the QueryRange API so far returned an `int64` losing the precision. As a hack, we also had to multiply `100`. 
This PR adds a `double valuePrecision` to the QueryRange API's MetricSample. 

The UI now shows values up to 2 decimals. 
Screenshot from 2023-02-21 18-50-54

Closes #2596 

The old `value` field stays around for backward compatibility. 
Once removed it probably makes sense to move the `100 *` to the frontend.


brancz (commented):
I don't think this is quite what I was imagining. I was imagining showing 2 values, 1 is the average per second value, and the other is the total over the step duration. The average per second value is the one we want to use for drawing the metrics graph, but the total value is what we still want to show in the hover pop up, as that's the value that will be consistent with the root of the icicle graph when clicked.

metalmatze:
Here are screenshots of individual profile types. 
I've double-checked the `Total` value. For all types, it equals the flame graph's cumulative total (root) again. :+1: 

CPU Samples:
cpu-samples
CPU Nanoseconds:
cpu-nanoseconds
Goroutines (non-delta)
goroutines

For the non-delta profiles it might make sense to _not_ show the Value and Total as they'll always be the same.

brancz (approved):
Very nice work!

Tiny nit, but I'm perfectly happy for this to be a follow-up: For non-delta profiles we should probably just show `Value` and not total, since it's always the same anyway.

PR #2643, opened 2023-02-22
pkg/debuginfo: close bucket reader

zdyj3170101136:
must close the reader of bucket, or there would be many goroutine.
截屏2023-02-22 下午2 33 24


brancz (approved):
lgtm

javierhonduco (approved):

PR #2644, opened 2023-02-22
ui: Fixed the unnecessary re-rendering of flamegraph

manojVivek:


ui/packages/shared/components/src/ConditionalWrapper/index.tsx
```diff
@@ -11,18 +11,24 @@
...
+  wrapperProps,
+  children,
+}: Props<T>): JSX.Element => {
   if (condition) {
-    return <Wrapper>{children}</Wrapper>;
+    return <WrapperComponent {...wrapperProps}>{children}</WrapperComponent>;
   }
 
-  return children;
+  return <>children</>;
```
yomete, line 33: Wouldn't this just return the text `children`? or we need to wrap in curly brackets

yomete (approved):
Nice! Can confirm it doesn't rerender on my end 👏🏾

brancz (approved):

ui/packages/shared/components/src/ConditionalWrapper/index.tsx
manojVivek: Yes. 🤦🏼‍♂️

Issue #2657, opened 2023-02-24
cpu profiler labels don't show up in parca


teleivo:
Hi! First, thank you for parca 😄 

I am trying to add [labels to my CPU profiles](https://github.com/DataDog/go-profiler-notes/blob/main/guide/README.md#cpu-profiler-labels) so I can filter profiles in parca. 

I read https://www.polarsignals.com/blog/posts/2021/04/13/demystifying-pprof-labels-with-go/ and the parca docs. I feel stuck 😅 

Here is a simplified example of my code

```go
package main

import (
        "context"
        "io"
        "log"
        "net/http"
        "net/http/pprof"
        rpprof "runtime/pprof"
        "strings"
)

func main() {
        mux := http.NewServeMux()
        mux.HandleFunc("/debug/pprof/", pprof.Index)
        mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
        mux.HandleFunc("/work", work)
        srv := http.Server{Addr: "localhost:6060", Handler: mux}
        log.Fatal(srv.ListenAndServe())
}

func work(w http.ResponseWriter, r *http.Request) {
        labels := rpprof.Labels("my", "label", "method", r.Method, "endpoint", r.URL.Path)
        rpprof.Do(r.Context(), labels, func(_ context.Context) {
                var sum int
                for i := 0; i < 1_000_000_000; i++ {
                        sum++
                }
                io.Copy(w, strings.NewReader("lots of work\n"))
        })
}
```

I then generate some load using

```sh
while sleep 0.1; do curl http://localhost:6060/work; done
```

The tags don't show up in parca in `pprof/profile`

![targets-labels](https://user-images.githubusercontent.com/4661144/221238174-b11f78f9-afed-49f8-9c3b-b0a85f806ef7.jpg)

They are also not in the profile I can download from parca and inspect using `go tool pprof`.

```sh
go tool pprof ~/Downloads/profile.pb.gz
File: main
Type: samples
Time: Feb 24, 2023 at 5:50pm (CET)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) tags
(pprof)
```

The tags are in if I directly get the profile using `go tool pprof`

```
go tool pprof http://127.0.0.1:6060/debug/pprof/profile?seconds=30
(pprof) tags
 endpoint: Total 21.2s
           21.2s (  100%): /work

 method: Total 21.2s
         21.2s (  100%): GET

 my: Total 21.2s
     21.2s (  100%): label
```

[...]

PR #2667, opened 2023-02-28
fix: export of custom pprof labels

teleivo:
work in progress to fix https://github.com/parca-dev/parca/issues/2657

* Add missing pprof_labels to the downloaded profile. Fix written by @brancz in https://github.com/parca-dev/parca/compare/pprof-labels-poc 😄 
* Normalization of labels is done in two places. Label normalization before ingest was adding the collision prefix 'exported_' even if no collision was present.
* Adapted `pkg/parcacol/normalizer_test.go` setup data as it did not seem to reflect reality. The `stringTable` does not contain the takenLabels key/values when running parca. `stringTable`  only contains strings from the profile. In case of a collision, the colliding key will be in the `stringTable` due to it being in the profile. In such a case it is not prefixed with `exported_` as this is done by parca.


pkg/parcacol/querier.go
```diff
@@ -83,7 +88,7 @@ func (q *Querier) Labels(
 
 	err := q.engine.ScanSchema(q.tableName).
 		Distinct(logicalplan.Col("name")).
-		Filter(logicalplan.Col("name").RegexMatch("^labels\\..+$")).
+		Filter(logicalplan.Col("name").RegexMatch("^"+ColumnLabels+"\\..+$|^"+ColumnPprofLabels+"\\..+$")).
```
brancz, line 91: the `.` in the constant won't quite behave the way we want here, it will match any character ... templating regexes is tricky I think it's easier to reason about if we just write it out in this case

pkg/parcacol/querier.go
```diff
@@ -102,7 +107,8 @@ func (q *Querier) Labels(
 				}
 
 				val := stringCol.Value(i)
-				seen[strings.TrimPrefix(val, "labels.")] = struct{}{}
+				l := strings.TrimPrefix(strings.TrimPrefix(val, ColumnLabelsPrefix), ColumnPprofLabelsPrefix)
```
brancz, line 110: based on the prefix we can tell what kind of profile it is and return that to the frontend, and only the frontend will unify it into a single list

pkg/parcacol/querier.go
```diff
@@ -129,8 +135,52 @@ func (q *Querier) Values(
 ) ([]string, error) {
 	vals := []string{}
 
+	// TODO(ivo) as we do not know if the labelName is a parca or pprof label
```
brancz, line 138: since we ensure on insertion that there is no conflict, we can do a single query that contains both columns in the distinct operation, we'll never get a value in both columns

pkg/parcacol/querier.go
```diff
@@ -169,7 +219,18 @@ func (q *Querier) Values(
...
-	ref := logicalplan.Col("labels." + matcher.Name)
+	labelExpr, err := matcherToBinaryExpression(matcher, logicalplan.Col(ColumnLabelsPrefix+matcher.Name))
+	if err != nil {
+		return nil, err
+	}
+	pprofLabelExpr, err := matcherToBinaryExpression(matcher, logicalplan.Col(ColumnPprofLabelsPrefix+matcher.Name))
+	if err != nil {
+		return nil, err
+	}
+	return logicalplan.Or(labelExpr, pprofLabelExpr), nil
```
brancz, line 230: Like I said on Discord, I'm not sure this is quite correct, we actually need to say something like "label.a exists and is equal to "b" and pprof_labels.a does not exist OR pprof.labels_a exists and is equal to "b" and labels.a does not exist"

Unfortunately, frostdb doesn't support this yet.

I think we can leave this part for last though, we might have it implemented by the time we fix all the other things! :)

pkg/query/pprof.go
```diff
@@ -152,9 +152,16 @@ func GenerateFlatPprof(ctx context.Context, ip *parcaprofile.Profile) (*profile.
 			s.Value = s.DiffValue
 		}
 
+		labels := make(map[string][]string, len(s.Label))
+		for k, v := range s.Label {
+			labels[k] = []string{v}
+		}
+
+		// TODO(ivo): should we add NumLabel?
```
brancz, line 160: if you want to take care of it, go for it! a separate PR would also be fine to keep this PR smaller

brancz (commented):
This is going into a really awesome direction, thank you so much for taking a stab at this!!

pkg/parcacol/querier.go
teleivo: perfect 😄

pkg/query/pprof.go
teleivo: another PR then 😄

pkg/parcacol/normalizer.go
```diff
@@ -167,7 +169,7 @@ func LabelNamesFromSamples(
 }
 
 // TODO: support num label units.
-func LabelsFromSample(takenLabelNames map[string]string, stringTable []string, plabels []*pprofpb.Label) (map[string]string, map[string]int64) {
+func LabelsFromSample(takenLabels map[string]string, stringTable []string, plabels []*pprofpb.Label) (map[string]string, map[string]int64) {
 	labels := map[string][]string{}
 	labelNames := []string{}
```
teleivo, line 174: @brancz we collect `labelNames` and go over them in sorted order later on. We return map `resLabels` so I wonder whether we need to do that. We don't do it in `LabelNamesFromSamples`. All tests (+ manual test) pass if I remove `labelNames` and just iterate over the `labels` keys.

pkg/parcacol/normalizer.go
```diff
@@ -188,11 +190,11 @@ func LabelsFromSample(takenLabelNames map[string]string, stringTable []string, p
 	resLabels := map[string]string{}
 	for _, labelName := range labelNames {
 		resLabelName := labelName
-		if _, ok := takenLabelNames[resLabelName]; ok {
-			resLabelName = "exported_" + resLabelName
+		if _, ok := takenLabels[resLabelName]; ok {
+			resLabelName = model.ExportedLabelPrefix + resLabelName
 		}
 		if _, ok := resLabels[resLabelName]; ok {
```
teleivo, line 196: Why is there another check done after the collision was handled? 🤔 

Is that in case I define labels `instance` and `exported_instance` in my pprof_labels when `instance` is already part of the labels?

pkg/parcacol/querier.go
teleivo: I don't quite understand 😄 

I assumed that since we removed the prefix for `labels.` we should do the same irrespective of where the prefix comes from.

Do you mean we should only trim the `labels.` prefix and the frontend is going to take care of removing the `pprof_labels.` prefix?

brancz:
Created https://github.com/polarsignals/frostdb/issues/394 and added it to the team's backlog so we'll hopefully have the last missing piece for this soon!

brancz:
https://github.com/polarsignals/frostdb/issues/394 is merged so we can now build the queries we want to here! 🥳

pkg/parcacol/normalizer.go
brancz: it's so we have a consistent order if we see the same labelset again, I agree we should have a test for this case, and we need to fix it in the other place, neither of that needs to happen in this PR though

brancz (commented):
I personally think we can leave the `LabelNames` and `LabelValues` calls out for now and leave them with the current behavior, and handle that in a follow-up, this is difficult enough without those :D your choice though

pkg/parcacol/normalizer.go
teleivo: got it, thank you!

pkg/parcacol/normalizer_test.go
```diff
@@ -21,55 +21,126 @@ import (
...
+		{
+			name: "colliding labels in descending order",
+			takenLabels: map[string]string{
+				"instance": "127.0.0.1:6060",
+			},
+			stringTable: []string{"", "instance", "17", "method", "GET"},
+			samples: []*pprofpb.Sample{
+				{
+					Label: []*pprofpb.Label{{
+						Key: 3,
```
teleivo, line 42: @brancz I just want to make sure I get this right 😅 I added some missing tests based on the ones that were already here. The `order` in `colliding labels in descending order` refers to the order of the `[]*pprofpb.Label` right? So the `label.Key` in descending order is `method, instance` in this case and reversed in the ascending case.

brancz:
Just rebased and tested this, and it works flawlessly. I'd be happy to merge this!

brancz:
Superseeded by #3365 

Thank you so much for writing this code @teleivo!

PR #2669, opened 2023-02-28
prefix the frostdb metrics

thorfour:


brancz (approved):

brancz:
Should this perhaps happen on the FrostDB side instead?

thorfour:
> 

We could. I've always felt the library shouldn't assume a namespace of the application using it.

brancz:
hmm I don't think I agree with that, that means that frostdb can't have docs of it's own or dashboards that would work regardless of the consumer

metalmatze:
Yep. I agree with Frederic on this one. All metrics coming from the FrostDB repository should have the frostdb prefix. If, Parca wants to, we could prefix them again, making them `parca_frostdb_wal...`, for example.

PR #2672, opened 2023-03-01
Update `buf.gen.yaml` to use new buf plugins

metalmatze:
Renovate bot is failing because the newer versions aren't available as the old custom plugins anymore.


kakkoyun (approved):

maxbrunet:
@metalmatze this regex needs to be updated too :)

https://github.com/parca-dev/parca/blob/8123277e8ec0efd7ba04c9fa54a5c114bf7e43cd/renovate.json#L47

PR #2676, opened 2023-03-02
Update renovate config on how to update `buf.gen.yaml`

metalmatze:
Is there some way to test these regexes? What's generally speaking a development flow on updating them and making sure they work as expected? :thinking: 

The problem is, that we're somewhat stuck with one plugin still using `remote: ` while the others moved over to `plugin: `. :man_shrugging: Not sure if there's a good solution right now.

cc @maxbrunet


renovate.json
```diff
@@ -44,7 +44,7 @@
       "description": "Update Buf plugins",
       "fileMatch": ["(^|/)buf\\.gen\\.yaml$"],
       "matchStrings": [
-        "# renovate: datasource=(?<datasource>.+?) depName=(?<depName>.+?)(?: (?:packageName)=(?<packageName>.+?))?(?: versioning=(?<versioning>.+?))?\\s*-?\\s*remote: ('|\")?.*:(?<currentValue>.+?)(-\\d+)?('|\")?\\s"
+        "# renovate: datasource=(?<datasource>.+?) depName=(?<depName>.+?)(?: (?:packageName)=(?<packageName>.+?))?(?: versioning=(?<versioning>.+?))?\\s*-?\\s*plugin: ('|\")?.*:(?<currentValue>.+?)(-\\d+)?('|\")?\\s"
```
maxbrunet, line 47: This bit can be removed, it was handling the revision, but it has moved to a dedicated key

```suggestion
        "# renovate: datasource=(?<datasource>.+?) depName=(?<depName>.+?)(?: (?:packageName)=(?<packageName>.+?))?(?: versioning=(?<versioning>.+?))?\\s*-?\\s*plugin: ('|\")?.*:(?<currentValue>.+?)('|\")?\\s"
```

maxbrunet (commented):
Thank you for taking care of that

> Is there some way to test these regexes? What's generally speaking a development flow on updating them and making sure they work as expected? :thinking: 

The config itself can be validated with `npx -p renovate renovate-config-validator`, but for the regexes, I do not know anything convenient, I usually test them in [regex101.com](https://regex101.com/).

> The problem is, that we're somewhat stuck with one plugin still using `remote: ` while the others moved over to `plugin: `. :man_shrugging: Not sure if there's a good solution right now.

`remote` is deprecated and I do not think `vtproto` will receive any update there, so this case can be ignored. We need to track this issue bufbuild/plugins#251. If we want the latest version of the `vtproto` plugin in the meantime, I guess we would have to manage its installation manually ourselves.

maxbrunet (approved):

PR #2677, opened 2023-03-02
Shareability fixes

monicawoj:
This PR addresses two issues that we were encountering:
1) Selected point in metrics graph & Flamegraph were not appearing despite this information being in the URL (shareability fix)
2) Flamegraph was not being updated for delta type profiles on time selection in the metrics graph


brancz (approved):
Nice job! Works perfectly now!

PR #2693, opened 2023-03-06
ui: Add a default visualization for the dashboard_items params

yomete:
We were not initializing a value for the `dashboard_items` param and this is causing pprof.me to crash.

This will ensure that other apps that uses the `ProfileView` component has a default value (`icicle`) to fall too


monicawoj (approved):
looks like there's one UI linter error to fix, but besides that looks great!!

PR #2696, opened 2023-03-06
EXPERIMENTAL arrow ingestion: create the arrow record from sorted buffer

thorfour:
Changed the record creation to happen from the sorted Parquet buffer to prevent the panic [seen in FrostDB](https://github.com/polarsignals/frostdb/issues/368)

So instead of writing the series directly to an Arrow record, it writes them to a Parquet buffer, sorts that Parquet buffer, and then reads the sorted rows back into an Arrow record.

While this is not ideal and it would likely be better to sort the series beforehand and then write the series directly to an arrow record as it was before, this is far less complex and requires less investment for the time being for an experimental feature. If we merge this I shall open an issue to track the desire to sort the series beforehand and remove the intermediary Parquet buffer step.


brancz (approved):
I understand this is just to test if that panic continues to occur after this, right? If so lgtm

brancz:
I would also be happy with that.

thorfour:
I think we should do both. FrostDB expects sorted rows, so this is a bug fix in the Parca side of things, but I agree that it would also be good to have the sanity check in Frost

asubiotto:
I agree we should do both long term, but if short-term the intent of this fix is to prevent panics in frostdb I think it's more effective to add the check/sorting on the frostdb side rather than parca. If the issue is that we don't have code to sort arrow records and don't want to invest in that right now, then we could probably convert to parquet, sort, and convert back even though it's not ideal performance wise.

thorfour:
This is intended to "do the right thing" on the Parca side where we write to FrostDB with sorted rows. Since we don't have a way to sort records right now, and we don't have the cycles to invest in that I think this corner cut is pragmatic while we explore ingesting arrow records. Not doing this means we would still need to implement a way to sort records on the Frost side which I'm trying to avoid investing in right now.

PR #2703, opened 2023-03-07
Dockerfile: Downloading grpc-health-probe with retries

manojVivek:
Would something like this would be fine to temporarily handle the `grpc-health-probe` binary download problem on GitHub actions?


brancz:
We should add to the comment that we added the retry because we’ve seen this fail *a lot*.

otherwise lgtm

manojVivek:
> We should add to the comment that we added the retry because we’ve seen this fail _a lot_.

Done.

brancz (approved):

PR #2710, opened 2023-03-08
pkg/parcacol: fix QueryRange

zdyj3170101136:
replace ts divide by 1000 by expand step/
otherwise it only means at least 1000 second(16 min) we get a point.

### DIFF
--- a/pkg/parcacol/querier.go
+++ b/pkg/parcacol/querier.go
@@ -301,6 +301,11 @@ func (q *Querier) QueryRange(
 	start := timestamp.FromTime(startTime)
 	end := timestamp.FromTime(endTime)
 
+	// With a scrape interval of 10s and a query range of 1d we'd query 8640 samples and at most return 960.
+	// Even worse for a week, we'd query 60480 samples and only return 1000.
+	if step < endTime.Sub(startTime)/1000 {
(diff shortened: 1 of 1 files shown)


pkg/parcacol/querier.go
```diff
@@ -600,9 +605,7 @@ func (q *Querier) queryRangeNonDelta(ctx context.Context, filterExpr logicalplan
 
 		// TODO: This still queries way too much data from the underlying database.
 		// This needs to be moved to FrostDB to not even query all of this data in the first place.
-		// With a scrape interval of 10s and a query range of 1d we'd query 8640 samples and at most return 960.
-		// Even worse for a week, we'd query 60480 samples and only return 1000.
-		tsBucket := ts / 1000 / int64(step.Seconds())
+		tsBucket := ts / int64(step.Seconds())
```
metalmatze, line 608: This worked as intended. We want to divide by 1000 as we store the timestamps as milliseconds. 
Then for each step, we only want to have one sample (the first one we find) to reduce the number of samples returned to the frontend. 

If you want a more fine grained metrics graph you need to zoom into the time range.

metalmatze:
I'm going to close this for now.

PR #2753, opened 2023-03-14
design-system: Storybook v7 setup and preliminary stories for the `Button` component

manojVivek:


ui/packages/shared/components/src/Button/Button.stories.tsx
```diff
@@ -0,0 +1,19 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Button from './';
+
+export default {component: Button, title: 'Components/Button'};
+export const Primary = {args: {variant: 'primary', children: 'Primary', disabled: false}};
+export const Neutral = {args: {variant: 'neutral', children: 'Neutral', disabled: false}};
+export const Link = {args: {variant: 'link', children: 'Link Text', disabled: false}};
```
monicawoj, line 19: is it possible to provide a user control in Storybook where we would be able to manually change the variant? For example a dropdown with all the available variant values?

manojVivek: Yes, that's in the plan.
Right now, the controls for native prop types work out of the box. The boolean and string works, the variants is an enum, and it doesn't work out of the box, I'll see what can be done for that.

manojVivek: Done, specifying the type explicitly worked here. Will proceed with that and then see if this can be inferred automatically some way
Screenshot 2023-03-16 at 4 43 31 PM
.

PR #2754, opened 2023-03-14
ui: Add perf monitoring for all visualizations

yomete:
This PR adds perf monitoring for
- Metrics graph
- Call graph and
- Top table


manojVivek (approved):
Lgtm!

Issue #2719, opened 2023-03-09
feature: searching in flamegraphs without recalculating the % of cycles


javierhonduco:
The new "filter by function" feature works great but unfortunately, it seems to create a "new" flamegraph, recomputing the % of the metric, such as CPU cycles.

This makes performance troubleshooting harder as it's not easy to understand what's the "global" resource usage for a particular frame/callstack.

Ideally, we would compute the percentages with the query, e.g. comm="parca", pid="4" but then filtering by function would leave the % as-is

PR #2758, opened 2023-03-14
ui: Improve showing filtered and trimmed nodes in ProfileIcicleGraph

metalmatze:
If nothing gets filtered and trimmed the icicle graph now looks like this (see the line at the bottom).
icicle-raw

With trimming enabled: 
icicle-trimmed

With filtering by function (for `sort`)
icicle-filtered

And finally with both trimming and filtering. 
icicle-filtered-trimmed


metalmatze:
Oh, seems like #2756 created a merge conflict. I'll rebase these tomorrow.

Other follow-ups are:

- Tooltip need to use the max unfiltered total value (solves #2719)
- Call graphs need the same capabilities.

brancz:
Super cool!! I think the indicating text should be above the icicle graph as it's unpredictable how large it may be so you may not even know if that text is there just because the icicle graph is too tall.

metalmatze:
> Super cool!! I think the indicating text should be above the icicle graph as it's unpredictable how large it may be so you may not even know if that text is there just because the icicle graph is too tall.

It was there before and I moved it below the graph. It's mostly debugging information that's there when needed but nothing that usually concerns users unless they look for it. 
I think we, as developers, are biased towards thinking the information is important because we think in terms of samples and filtering and trimming. It's still there, just not at the very top.

metalmatze:
I'm writing and updating the flamegraph_table_test.go cases now. :relieved:

brancz:
Alright happy with trying like this. I'm not convinced if I'm honest but I was wrong about function filtering as well, so I'm very happy to give it a shot and judge after using it a bit.

brancz (approved):

proto/parca/query/v1alpha1/query.proto
```diff
@@ -309,6 +309,9 @@ message Flamegraph {
 
   // untrimmed_total is the total weight of the flame graph before trimming
   int64 untrimmed_total = 9;
+
+  // unfiltered_total is the total weight of the flame graph before trimming
+  int64 unfiltered_total = 10;
```
brancz, line 314: hmm after thinking about this over lunch, should we maybe rather make this part of the overall response, and not the flamegraph itself? this seems like it's not to the specific visualization, but rather the raw data used to create it.

metalmatze: That is correct. We want the same for the call graph (see comment above). I mostly added it here because the `untrimmed` was there already. 
Additionally, if we were to add this. I'd be in favor of calling the fields `trimmed` and `filtered` and setting them to the difference. That way the raw flame graph would be `total + trimmed + filtered`. It's really weird to work with right now.

metalmatze: Looking at the proto again, would we move all `total` fields of all QueryResponse reports to the top too? 
I feel like this would be a bigger change, including the total field overall too. It might be out of the scope for this PR. However, given we want to be backward compatible, we might not want to merge this PR as is then.

yomete (approved):
lgtm!

proto/parca/query/v1alpha1/query.proto
metalmatze: Update the proto to the discussed changes.

manojVivek (approved):
Lgtm!

PR #2766, opened 2023-03-15
Show absolute and relative percentages in tooltip 

### LINKED ISSUE #2719: feature: searching in flamegraphs without recalculating the % of cycles

metalmatze:
If the flame graph has been filtered or trimmed the absolute percentage is shown first.

I really tried to make the wording as intuitive as possible. I'm still not sure it's the best though. Thoughts?
If the flame graph isn't filtered or trimmed just one percentage is shown. 
Screenshot from 2023-03-15 15-12-45

Closes #2719 
Depends on #2758


metalmatze:
> I would find it more obvious if it was always just the percentage of whatever is displayed, be that filtered or non-filtered. Plus it seems intuitive where if you reset the view it will show you the full percentage again.

That was what we had so far. The issue was originally opened because that wasn't enough. And I agree, I want to know what a function has used absolutely too.

metalmatze:
I'm still not sure about the wording.
Right now I'm learning towards `33.51k (41.66% / 71.76% relative)` as filtered isn't correct when no one filtered and samples got trimmed instead.

brancz:
The percentage should never change because of trimming. The total should still be total without trimming.

I wonder if `XY% total / YZ% filtered` could work?

metalmatze:
Yes, I think now, when changing the trimming as we discussed on Discord, the filtered will actually be the only time the graph's total changes, thus the only time the second percentage will be needed.

javierhonduco:
Wondering if removing the total count would be useful here, and just saw the percentage(s).

Ultimately we need to decide what's deciding the value for the denominator. I think that should be decided with the query system we have, but filtering/drilling down should not change the ratio and perhaps not even show the filtered one?

brancz:
I find the filtered value useful if I'm optimizing a specific user interaction like a database query. These are typically not representative in the total CPU time of a process, they rather manifest in user-perceived latency.

metalmatze:
Let's merge https://github.com/parca-dev/parca/pull/2815 first and then I'll take another look and post here. 
Thanks for the feedback so far.

metalmatze:
Updated the UI to show the percentage `of filtered` graph. 
Screenshot from 2023-03-22 11-56-15

PR #2767, opened 2023-03-15
fix "Add panel" and visualization type selectors

monicawoj:
Adding a 'relative' class to the outer 'div' element holding the select button and options should fix the absolute positioning issue that was taking place on scroll.


ui/packages/shared/components/src/Select/index.tsx
```diff
@@ -69,101 +69,92 @@ const Select = ({
...
-                                    </span>
-                                  ) : null}
-                                </>
-                              )}
-                            </Listbox.Option>
-                          ))}
-                        </div>
+    <Listbox value={selectedKey} onChange={onSelection}>
+      {({open}) => (
+        <div className="min-w-fit relative">
```
monicawoj, line 74: Other than cosmetic changes introduced to the code (for example, removing React fragment elements <></> where not necessary) this is the only significant change introduced in this PR.

yomete (approved):
lgtm!

manojVivek (approved):
Lgtm!

PR #2785, opened 2023-03-16
Show step duration in metrics graph tooltip

### LINKED ISSUE #2649: Show Step duration in MetricsGraph tooltip
With merging https://github.com/parca-dev/parca/pull/2640 we show Value and Total values in the MetricGraph tooltip.

The Value is the average per second value across the duration.
And Total is the total value across the duration.

Imagine we have a step duration of 10s. Within those 10s we have 3 samples and each has a profiling duration of `2s`, meaning each profile observed the process for 2s. The sum of this duration is `6s`. So for the final duration, we want to display that there were 1.14 total CPU seconds observed over the `duration` of 6s. 

In the example below we would want to have: 

|      |      |
| --- | --- |
| Value  | 0.11399ns |
| Total  | 1.14s  |
| Duration | 6s |
| At | Feb 22 ... |

cpu-nanoseconds

monicawoj:
Resolves #2649


manojVivek (approved):
Lgtm!

metalmatze (approved):
Works perfectly. 

We need to follow up with the units at some point. Although this isn't the only place. Probably we should go through most of Parca's UI and collect where to update units and what the correct ones would be.

PR #2788, opened 2023-03-16
pkg/query: When trimming nodes only add necessary entries to string table

### LINKED ISSUE #2763: Trim strings table when trimming flame graph
When generating the flame graph the string table gets populated.

https://github.com/parca-dev/parca/blob/7aefbe49fefbd3acf5935c8a1118aa20532edad2/pkg/query/flamegraph_table.go#L33-L42

Only afterward the flame graph gets trimmed. The string table is possibly populated with a bunch of strings we don't care about anymore. Those strings get marshaled and sent to the frontend increasing the payload size and having the frontend read useless data.

metalmatze:
On the one hand, we trim the flame graph to make it quicker to render, and on the other hand, to reduce the payload size. However, until now we still sent the full string table along. Potentially many strings node samples might have been trimmed from the graph.

Closes #2763


metalmatze:
Given that fewer nodes are trimmed now the effects on the string table are fewer too. Still something though:

Here is the old vs new bytes of the string table.
```
OLD   NEW   PERCENTAGE
12788 12440 0.9727869878010635
16776 16337 0.973831664282308
16776 16330 0.9734144015259895
```

metalmatze:
Yes agreed, let's see how much it shows up in our ... profiles :smiley:

PR #2811, opened 2023-03-20
Bump frost

thorfour:
This brings in the latest FrostDB instance which fixes a panic in large aggregations.

Updates the Aggregation querier functions to handle multiple callbacks which can now happen for very large aggregations.


brancz:
So far lgtm but there are some more changes necessary for bumping to latest frostdb.

brancz (approved):

PR #2815, opened 2023-03-21
Trim flame graph nodes relative to node's parent

metalmatze:
So far we have trimmed nodes based on the flame graph's absolute cumulative value. This is not what we want to do, given we might lose data when diving further into a flame graph. 
However, in relation to the parent's cumulative value, there may occasion where the direct child still has too few values and would never be rendered. Those we want to trim.

cc @manojVivek @brancz


metalmatze:
Probably a good follow-up. We show the percentage of trimmed values. However, with these changes, they are often below 1%. I've seen trimmed values make up from `0.3%` to `0.8%`. Looking at the implementation in the frontend, we use `new Intl.NumberFormat('en-US');` which doesn't support `toFixed(2)`.
cc @manojVivek 
Screenshot from 2023-03-21 16-52-21

brancz (approved):

PR #2816, opened 2023-03-21
ui: Delete the @parca/functions folder and its usage

yomete:
The package is already deprecated as seen [here](https://www.npmjs.com/package/@parca/functions)


yomete:
@manojVivek @monicawoj

manojVivek (approved):
Lgtm!

PR #2826, opened 2023-03-22
introduce deterministic metrics graph coloring

### LINKED ISSUE #2478: Deterministic graph colors
Hey there, great project!

I am trying out parce for a small project of mine to evaluate for further use. Setting it up locally using the binary worked like a charm and I am playing around with the UI for now.

# Issue Description
A small thing I noticed is that the colors of the graphs do not seem to be assigned in a deterministic way. This causes the colors of the graphs to switch when pressing the "Search" button a couple of times:

image
image

This can be quite confusing, especially if the graph contains even more graphs. I guess a production deployment of parca will use a more sophisticated setup including a grafana for visualisation, but it would still be nice to improve the out-of-the-box experience.

# Proposed Solution

Assign graph colors in a more deterministic way.

Since profiles are "tagged" with things like `instance` and `job` (and probably others), tweaking the color selection algorithm to order profiles by taking these tags into account should be a viable option.

monicawoj:
resolves #2478


ui/packages/shared/profile/src/MetricsGraph/index.tsx
```diff
@@ -58,6 +58,7 @@ export interface HighlightedSeries {
 interface Series {
   metric: Label[];
   values: number[][];
+  id: string;
```
metalmatze, line 61: In this instance, we should probably call it `labelset`. 
```suggestion
  labelset: string;
```

metalmatze (commented):
Very cool. Looking forward to giving it a try!

yomete (approved):

manojVivek (approved):

PR #2830, opened 2023-03-22
ui/packages/shared/hooks/src/useUserPreference: Enable trimming by default

metalmatze:
With the recent improvements to trimming, we want to enable trimming by default again.


metalmatze:
Given this only flips a boolean I'm going to enable auto merged :relaxed:

PR #2831, opened 2023-03-22
Fix string trimming table bugs

metalmatze:
There are some other bugs that still need to be fixed tomorrow, but at least it doesn't crash the UI anymore.

### DIFF
--- a/pkg/query/flamegraph.go
+++ b/pkg/query/flamegraph.go
@@ -87,6 +87,12 @@ func (fgi *FlamegraphIterator) At() *querypb.FlamegraphNode {
 	return peekNode.Children[fgi.stack.Peek().currentChild]
 }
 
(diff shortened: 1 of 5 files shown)


metalmatze:
Alright. A new approach to trimming the string table. 
The rendering bug is entirely gone (no more runtime or kernel stack within application functions).


Instead of copying all metadata over to a new table we now track which strings, locations, and functions are actually needed.
The ones that can be ignored are going to be set to `""` (strings) or `nil` instead of fully being gone. 
I would say, that's fine and close enough for saving bytes when sending via gRPC. 

It still has a noticeable impact on the string table bytes. 
```
OLD    NEW    PERCENTAGE
105522 100853 95.57533026288357
105522 101106 95.8150906919884
111749 107115 95.85320674010505
```

brancz (approved):
nice work!

PR #2844, opened 2023-03-24
feature/signedupload: Support s3

zdyj3170101136:


zdyj3170101136:
the golangci-lint is not success because of s3 code.
i copy it directly from objstore.
should i changed it?
@brancz 

截屏2023-03-24 上午11 21 49

v-thakkar:
Drive-by comment: Is S3 package written by you or you're using Thanos's package with some modifications? Either way, you'll need to add Parca License in all.

zdyj3170101136:
> Details

you can see the first commit.
the s3 package is init from objstore.
i only add SignedPUT function on s3.Bucket struct.

seems the three auto commit add parca license.

zdyj3170101136:
> Drive-by comment: Is S3 package written by you or you're using Thanos's package with some modifications? Either way, you'll need to add Parca License in all.

you can see the first commit.
the s3 package is init from objstore.
i only add SignedPUT function on s3.Bucket struct.

seems the three auto commit add parca license.

cmd/parca/main.go
```diff
@@ -36,6 +37,7 @@ func main() {
 	ctx := context.Background()
 	flags := &parca.Flags{}
 
+	debug.PrintStack()
```
kakkoyun, line 40: Shall we hide this behind a debug flag?

kakkoyun (commented):
Thanks for taking care of it 🙏 🤗

I would rather see the changes committed to https://github.com/thanos-io/objstore
We can be better open-source citizens. I could help you with the reviews. I'm still a maintainer of that repo.

brancz:
I’m not convinced this can land in the objstore package. I thought about this when introducing the signed uploads originally, but the reality is that these features work wildly different on each provider so I don’t think an abstraction that’s suitable for the objstore package is possible.

zdyj3170101136:
> Thanks for taking care of it 🙏 🤗
> 
> I would rather see the changes committed to https://github.com/thanos-io/objstore We can be better open-source citizens. I could help you with the reviews. I'm still a maintainer of that repo.

as @brancz said, it is not easy to merge this to objstore

PR #2847, opened 2023-03-24
pkg/debuginfod: Fix caching and follow redirects of debuginfod servers

### LINKED ISSUE #2240: Debuginfod cache stores, but doesn't read from cache
While working on #2237 I noticed that while the debuginfod cache stores the debuginfo in the specified object storage bucket, it does not actually attempt to retrieve it next time that piece of debuginfo is requested.

https://github.com/parca-dev/parca/blob/fc26fd6d8739d45b2d77d335c05584e9f9366c95/pkg/debuginfo/debuginfod.go#L98-L125

I might just take care of this as part of #2237, but I wanted to open the issue to not loose track of it.

brancz:
This patch does two main things aside from a small cleanup and simplification:

* Fix caching: Previously while we wrote debuginfo to the object storage cache, we never tried to read if from there in the future.
* Follow redirects if an upstream debuginfod server returns a HTTP 3xx response.

Fixes #2240

### DIFF
--- a/pkg/debuginfo/debuginfod.go
+++ b/pkg/debuginfo/debuginfod.go
@@ -15,6 +15,7 @@ package debuginfo
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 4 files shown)


kakkoyun (approved):
LGTM ❤️ 💚 💙 💛 💜

PR #2866, opened 2023-03-27
proto/parca/share: Add total and filtered to QueryResponse

metalmatze:
These changes have been made to the Parca QueryResponse too and the UI of pprof.me expects these to be there now. The same seems fine to do for the share proto.


javierhonduco (approved):
Pity we didn't catch this in CI

metalmatze:
> Pity we didn't catch this in CI

Indeed. Maybe that's one of the few downsides of being backward compatible. :smiley:

brancz (approved):

PR #2884, opened 2023-03-29
ui/ProfileView: Pass total and filtered into ProfileView

metalmatze:
Not exactly sure about the types with Number and BigInt. Where to convert.


manojVivek (approved):
Lgtm!

metalmatze:
As written in one of the TODOs, we should rethink how we get responses and then pass them down into individual components. Only the component that renders each individual report type should care about what type of report is getting passed on. All components above that should work with a "generic" QueryResponse.

PR #2885, opened 2023-03-29
modify workflow for chromatic

monicawoj:
- add missing skip-check job
- distinctly specify "exit-zero-on-changes" flag so that changed stories do not cause the check to fail (this option is true by default, but it will be helpful to have this distinctly in the workflow file so that we are aware of this)


yomete (approved):
lgtm!

PR #2910, opened 2023-04-01
pkg/symbol: Position independent executable

marselester:
I am working on a [DIY symbolizer](https://github.com/marselester/diy-parca-agent/tree/main/cmd/addr2func) and decided to share some of the results -- support of PIE. I've started with simple cases (see a [blog post](https://marselester.com/diy-cpu-profiler-position-independent-executable.html)) such as resolving function names using only `.symtab` section, i.e., I haven't touched dynamic libraries yet.

PIE (position independent executable) is used by default in gcc for security measures, i.e., address space layout randomization. This means that the executable segment is mapped to a random high memory address such as `0x5646e2188000` instead of the usual `0x401000` address (non PIE).


kakkoyun (commented):
Thanks for the fantastic PR!

I'll ask for your patience with this. I want to take some time to review this thoroughly. We have other open PRs that touch these parts.

marselester:
@kakkoyun thank you!

kakkoyun (commented):
Thanks a lot for the contribution ❤️ 

As we talked offline. These changes need more consideration. We might evolve this into something else.

marselester:
Thank you for taking time walking me through the related code in the Parca Agent!

marselester:
Here is a summary of address normalization. Please feel free to fill the gaps.

Parca Agent stores CPU samples in pprof format and uploads them to Parca Server. The format itself allows to store function names, but Agent doesn't symbolize anything and delegates this responsibility to Server. Moreover, unlike pprof tools, Agent doesn't write "raw" sampled addresses into profiles, it normalizes them first. This decision was made to offload Server, so it wouldn't need to normalize an address on each `PCToLines` call.

Normalization allows one to look up a function name in a symbol table using an address
which is within the ELF file's executable segment range (where `.text` resides).

```go
// GetBase determines the base address to subtract from virtual
// address to get symbol table address. For an executable, the base
// is 0. Otherwise, it's a shared library, and the base is the
// address where the mapping starts. The kernel needs special handling.
base, _ := elfexec.GetBase(&ef.FileHeader, ph, f.m.kernelOffset, f.m.start, f.m.limit, f.m.offset)
normalizedAddress := addr - base
```

See related Agent code:

- writing an address to pprof https://github.com/parca-dev/parca-agent/blob/d10880037c3cc5aee33bb35b9a94ba02c05b9ad3/pkg/profiler/cpu/cpu.go#L935
- `Normalize` https://github.com/parca-dev/parca-agent/blob/d10880037c3cc5aee33bb35b9a94ba02c05b9ad3/pkg/address/normalizer.go#L48
- `ObjAddr` https://github.com/parca-dev/parca-agent/blob/d10880037c3cc5aee33bb35b9a94ba02c05b9ad3/pkg/objectfile/object_file.go#L156
- `GetBase` https://github.com/parca-dev/parca-agent/blob/d10880037c3cc5aee33bb35b9a94ba02c05b9ad3/internal/pprof/elfexec/elfexec.go#L221

Before uploading an object file (an executable or a shared library), it is stripped off of ELF sections that don't help Server with symbolization. For example, `.text` section where machine instructions live is deleted, but the `.symtab` section that contains a symbol table is preserved.

Issue #2024, opened 2022-11-02
ui: Make time format consistent across summary and input fields


Sylfrena:
Currently, the dropdown time-selector menu uses two different time formats for the input field(`November 2`) and the summary field(`2 Nov`). We should use a consistent format across both.
![image](https://user-images.githubusercontent.com/35404119/199466211-9d63a627-1384-45bd-a07f-f4915922e254.png)

metalmatze:
Given that we use UTC and follow Prometheus mostly, we should probably change the formatting to `yyyy-MM-dd HH:mm:ss` (right now would be `2023-04-04 09:13:30`) across the board. We can still use relative formatting where applicable, but overall whenever using absolute values this is what we should settle on.

yomete:
Fixed in #2920

PR #2920, opened 2023-04-04
ui: DateTimePicker adds seconds and uses yyyy-MM-dd HH:mm:ss

metalmatze:
Both Prometheus and Grafana use the above format for absolute formatting. Personally, I think that makes a lot of sense, especially when using UTC all over the place. Given we also have #2024 too, I went ahead and made the change in the DateTimePicker already and not only adding seconds.

Additionally, I fixed some styling issues along the way. Most prominently, the `p-2` to `pt-4` which had the DateTimeRangePicker offset horizontally.


yomete (approved):
lgtm!

PR #2922, opened 2023-04-04
ui: Fix circumstances where there is only a single label

brancz:
In this case the query parameter type is not a list of strings, but rather a single string so that needs to be handled separately.

Found this while testing parca-push.


manojVivek (approved):
Lgtm!

metalmatze (approved):

PR #2926, opened 2023-04-04
add option to split arrow records on ingestion

thorfour:


asubiotto (approved):
Eventually we'll swap over to use more efficient arrow records (i.e. REE arrays) so this will be unnecessary, right?

thorfour:
> Eventually we'll swap over to use more efficient arrow records (i.e. REE arrays) so this will be unnecessary, right?

I don't think it would become unnecessary, but it may be that we could use much larger row count values.

PR #2934, opened 2023-04-05
ui-perf: Protobuf long type handled as `Bigint` instead of `string` on the UI

manojVivek:


brancz (approved):
Amazing work! Also super cool that this is frontend-only change. If a backend change was involved this would be much more difficult to roll out.

metalmatze:
Do we still plan to go through with this? Would be great!

manojVivek:
Ah, looks like it got slipped from merging. 😐 
Taking care of this now.

PR #2955, opened 2023-04-07
Hidden flag to disable address normalization for debugging

marselester:
This flag mirrors https://github.com/parca-dev/parca-agent/pull/1534 so we should be able to test the changes without interfering with release cycles on both sides.

PR #2978, opened 2023-04-12
Store address normalization flag within Location entity

marselester:
The symbolizer package retrieves unsymbolized locations (i.e., sampled addresses) from the metastore (Badger). The locations are of `*pb.Location` type. Since the symbolizer needs to know whether addresses are normalized, I've added `is_raw_address` field to the proto definition of the `Location` message. I couldn't call it `is_normalized` because we need a field name that is false by default since Parca normalizes the ingested addresses as of now.


pkg/parcacol/normalizer.go
```diff
@@ -321,10 +321,11 @@ func (n *MetastoreNormalizer) NormalizeLocations(
 		}
 
 		req.Locations = append(req.Locations, &pb.Location{
-			Address:   addr,
-			IsFolded:  location.IsFolded,
-			MappingId: mappingId,
-			Lines:     lines,
+			Address:      addr,
+			IsRawAddress: false,
```
marselester, line 325: We can set `IsRawAddress: !n.isAddrNormEnabled` once https://github.com/parca-dev/parca/pull/2955 is merged.

marselester:
@brancz sure, that would be the best. I was thinking to use [objectfile package](https://github.com/parca-dev/parca-agent/blob/d10880037c3cc5aee33bb35b9a94ba02c05b9ad3/pkg/objectfile/object_file.go) and call `normalizedAddr, err := objFile.ObjAddr(addr)` in `symbolizeLocationsForMapping` because it has access to mapping, ELF file, and sampled addresses. The objectfile is currently part of the agent. Since the server doesn’t import anything from the agent, the package either could be moved to the server (here) or into a separate repository.

```go
// symbolizeLocationsForMapping fetches the debug info for a given build ID and symbolizes it the
// given location.
func (s *Symbolizer) symbolizeLocationsForMapping(ctx context.Context, m *pb.Mapping, locations []*pb.Location) ([][]profile.LocationLine, liner, error) { ... }
```

marselester:
I gave up on making all the tests pass (`elfexec_test.go` and `object_file_test.go`) for my hand-made normalizer and decided to copy the `objectfile` and `elfexec` packages instead.

pkg/symbol/addr2line/symtab.go
```diff
@@ -57,25 +57,29 @@ func Symbols(logger log.Logger, filename string, f *elf.File, demangler *demangl
...
 
 func (lnr *SymtabLiner) PCRange() ([2]uint64, error) {
 	return lnr.searcher.PCRange()
 }
 
 // PCToLines looks up the line number information for a program counter (memory address).
-func (lnr *SymtabLiner) PCToLines(addr uint64) (lines []profile.LocationLine, err error) {
+func (lnr *SymtabLiner) PCToLines(addr uint64, isRawAddr bool) (lines []profile.LocationLine, err error) {
+	if isRawAddr {
+		if addr, err = lnr.objFile.ObjAddr(addr); err != nil {
```
kakkoyun, line 79: We need to discuss whether we want to pull this logic from the agent to here. 

The object file we have here could be just a debuginfo file, so it's highly likely that it does not have the required `ProgramHeader` and `Note` sections. This is a design choice we made. We only want to send what is strictly necessary over the wire from the hosts.

We should send the necessary information (base address) as part of the API (probably in the mapping),

I'm happy to base address calculations as part of the tests. However, we can vastly simplify them. It's a shame that we cannot directly use these facilities from the agent (cyclic dependencies are hard).

kakkoyun: Also, with the recent changes, we moved most of the logic to the `Mapping` layer, which is semantically more correct. This information that is needed here belongs to the mapping not to the object file.

kakkoyun (commented):
Fantastic work 💯 I have commented.

We are close to the finish line for the base address logic.
Thanks a lot for helping us out on this.

pkg/symbol/addr2line/symtab.go
marselester: Thank you! I will have a look at the recent changes in the agent.

marselester: I've added base address fields to protobufs so they could be passed from the agent and stored in Badger, and later used during symbolization.

pkg/parcacol/normalizer.go
```diff
@@ -234,7 +234,13 @@ type mappingNormalizationInfo struct {
 	offset int64
 }
 
-func (n *MetastoreNormalizer) NormalizeMappings(ctx context.Context, mappings []*pprofpb.Mapping, stringTable []string) ([]mappingNormalizationInfo, error) {
+// NormalizeMappings converts pprof mappings into the internal mapping format
+// and stores them in the metastore (Badger).
+func (n *MetastoreNormalizer) NormalizeMappings(ctx context.Context, mappings []*pprofpb.Mapping, stringTable []string, baseAddresses []uint64) ([]mappingNormalizationInfo, error) {
+	if len(baseAddresses) > len(mappings) {
```
kakkoyun, line 240:
```suggestion
	if len(baseAddresses) != len(mappings) {
```
Maybe an equality check could be better to cover the case where `baseAddresses` is less than `mappings`.

pkg/symbol/addr2line/go.go
```diff
@@ -89,6 +91,9 @@ func (gl *GoLiner) PCToLines(addr uint64) (lines []profile.LocationLine, err err
 
 	name := "?"
 	// TODO(kakkoyun): Do we need to consider the base address for any part of Go binaries?
```
kakkoyun, line 93: I think now this comment is deprecated, no?

proto/parca/profilestore/v1alpha1/profilestore.proto
```diff
@@ -60,6 +60,9 @@ message LabelSet {
 message RawSample {
   // raw_profile is the set of bytes of the pprof profile
   bytes raw_profile = 1;
+  // base_addresses contains base addresses for each pprof mapping
+  // found in the raw profile in pprof format.
+  repeated uint64 base_addresses = 2;
```
kakkoyun, line 65: Let's try to omit this from this level and try to use the data from the mapping level. It would be cleaner.
We can pass down the mapping to liners for example. That way we can omit passing `isRaw` to the `PCToLine`.

What do you think?

kakkoyun (commented):
It looks like it's about to be done ⌛ Awesome work 🎉 

I have a suggestion about how we wire things.

Also, I'd like to know what @brancz thinks as well.

pkg/symbol/addr2line/go.go
marselester: Removing it... I wasn't sure about this TODO comment and left it for review :).

proto/parca/profilestore/v1alpha1/profilestore.proto
marselester: > try to use the data from the mapping level

From what I understand the agent sends raw profiles in pprof format, so we can't add a base address field to pprof's `Mapping` message https://github.com/google/pprof/blob/main/proto/profile.proto#L147 since we don't control it. The server defines its own `Mapping` message to store received profiles in metastore. The conversion from pprof to an internal mapping format happens in `NormalizeMappings()` during pprof ingestion.

I thought that `RawSample` is the best candidate to pass base addresses without breaking backward compatibility 🤔.

marselester: @kakkoyun I've moved `isRaw` field from `Location` to `Mapping` protobuf message to omit passing `isRaw` to the `PCToLine`.

brancz:
I'm only now getting to looking deeper into this, and I feel what we should do is instead send [this information](https://github.com/parca-dev/parca-agent/blob/0685522a698a1e9ea84762cdbd57af33bb256786/pkg/elfreader/executable.go#L71-L102):

* `fh.Type`
* `loadSegment.Off`
* `loadSegment.Vaddr`

The backend can then perform the correct calculation as it needs with the raw information. Beyond that, if the executable was uploaded (which we can [now](https://github.com/parca-dev/parca/pull/3590) do) we can also perform this extraction/calculation server side (that would allow supporting pprof profiles produced by things like [perf_data_converter](https://github.com/google/perf_data_converter)).

Since this has become a priority for a customer I'm going to have a look at this this week.

marselester:
@brancz you're welcome!

PR #2979, opened 2023-04-13
api: `Total` field fixed for Top table response

manojVivek:


pkg/query/top.go
```diff
@@ -73,10 +75,11 @@ func GenerateTopTable(ctx context.Context, p *parcaprofile.Profile) (*pb.Top, er
 	}
 
 	top := &pb.Top{
-		List:     list,
-		Reported: int32(len(list)),
-		Total:    int32(len(list)),
-		Unit:     p.Meta.SampleType.Unit,
+		List:       list,
+		Reported:   int32(len(list)),
+		Total:      int32(len(list)),
```
brancz, line 80: I find this confusing. I think we should rather put the cumulative into the total value, and add a new field that's called "RowNumber" that more accurately reflects the meaning.

metalmatze (approved):

PR #3019, opened 2023-04-23
perf(grpc): use optimized code from vtprotobuf

maxbrunet:
The optimized code generated by `vtprotobuf` had actually been left unused until now:
[planetscale/vtprotobuf@`main`#using-the-optimized-code-with-rpc-frameworks](https://github.com/planetscale/vtprotobuf/tree/main#using-the-optimized-code-with-rpc-frameworks)

Since it uses the global state, I load it as earlier as possible before any gRPC client or server start, and do not do it under `pkg/` as dependents may use other gRPC clients/servers without the optimized VT code.


javierhonduco (approved):
LGTM

kakkoyun (approved):
LGTM 🚀

PR #3020, opened 2023-04-23
fix: change automaxprocs log level to debug

maxbrunet:
This is related to internals, and irrelevant to the end-user. I would not like it to appear as recommendation to use CPU limits.


kakkoyun (approved):
LGTM ❤️

PR #3055, opened 2023-05-02
Support stacktrace ID's as a slice of strings

thorfour:
Changes the `StacktraceID` from a `string` to a `[]string` which allows the stack trace to be stored as a slice of locations instead of just a single string.


metalmatze:
I still think we should potentially do this, just need to discuss we want to have another follow-up after that.

PR #3069, opened 2023-05-04
pkg/query: Create flamegraph_arrow

metalmatze:
This creates a new flame graph representation. Instead of creating the flame graph with Go structs directly, we utilize Apache Arrow to create a record including the rows of each location.

The way the arrow schema looks can be viewed in `pkg/query/flamegraph_arrow.go` and `pkg/query/flamegraph_arrow_test.go` contains an example of what we currently have. 

The API currently doesn't expose this type just yet. Only the enum type was added to proto to be able to call `RenderReport` with `QueryRequest_REPORT_TYPE_FLAMEGRAPH_ARROW`.


pkg/query/flamegraph_arrow.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/apache/arrow/go/v10/arrow/array"
+	"github.com/apache/arrow/go/v10/arrow/memory"
+	"github.com/polarsignals/frostdb/pqarrow/builder"
+	"go.opentelemetry.io/otel/trace"
+
+	"github.com/parca-dev/parca/pkg/profile"
+)
+
+const (
+	flamegraphFieldMappingStart   = "mapping_start"
```
asubiotto, line 30: I wonder if we can make this code cleaner by having a map from field name to some struct that encapsulates the arrow type and holds a reference to the builder.

metalmatze:
Next week, I'll bring this PR into a mergeable state. So we can make progress on smaller follow-up PRs.

The follow ups on the Go side should be:
1. Trimming
1. pprof label support
1. cleaning up (potentially making our record builder using generics)
1. Exposing the new type from our API (this is more likely part of the UI changes)

metalmatze:
There's one more fix I need to do in upstream Arrow (due to which the CI is going to fail here)

proto/parca/query/v1alpha1/query.proto
```diff
@@ -320,6 +323,21 @@ message Flamegraph {
...
+  bytes record = 1;
+
+  // unit is the unit represented by the flame graph
+  string unit = 2;
+
+  // height is the max height of the graph
+  int32 height = 3;
+
+  // trimmed is the amount of cumulative value trimmed from the flame graph.
+  int64 trimmed = 4;
```
brancz, line 338: do we strictly need this?

brancz (approved):
Nice work! Happy to merge this and iterate!

metalmatze:
The new changes break the tests again. I'll fix those tomorrow.

pkg/query/flamegraph_arrow.go
```diff
@@ -199,13 +198,13 @@ func generateFlamegraphArrowRecord(ctx context.Context, mem memory.Allocator, tr
 
 	cumulative := int64(0)
 	height := int32(0)
-	rootsRow := []uint32{}
-	children := make([][]uint32, len(p.Samples))
+	rootsRow := []int{}
+	children := make(map[int][]int, len(p.Samples))
```
brancz, line 202: This we should undo I think, but if you rather wanted to commit what works and improve I totally understand

PR #3071, opened 2023-05-04
pkg/query/flamegraph_arrow: Start aggregation of flamegraph

metalmatze:


brancz:
I’m not sure the strategy is right here. We should do aggregations on the first time building already. Rebuilding from scratch was one of the major flaws in the previous iterations.

metalmatze:
You're right. I'll rework this entirely.

PR #3095, opened 2023-05-08
chore(container): add commit timestamp to image tag

maxbrunet:
It will allow the automation to sort the tags alphabetically, and select the latest to update the Kubernetes manifest with. Example: `main-1611906956-2d3fcbd`

See "Staging CI/CD workflow" of [Automate image updates to Git | Flux](https://fluxcd.io/flux/guides/image-update/).

It also fixes `COMMIT` to always be a Git commit SHA, never a Git tag.


metalmatze (approved):
Happy to use the commit timestamp as that's reproducible rather than the build timestamp. :+1:

PR #3103, opened 2023-05-10
Add filtering function names to QueryRange (metrics) 

metalmatze:
The filterByFunction moves from just the flame graph up to the ProfileExplorer (the top bar where a query gets defined). The filter query is used not only by the flame graph but also by the QueryRange (metrics). We should make sure to test with table and call graph (or at least plan to use the same filters in the future too).

There are a few small TODOs left but I wanted to share:
- [ ] Tooltips show `NaNns` for duration


ui/packages/shared/utilities/src/index.ts
```diff
@@ -36,34 +36,39 @@ export const capitalize = (a: string): string =>
...
+
+const percentage = {
+  percentage: {divisor: 0.01, symbol: '%'}, // dividing by 0.01 is like multiplying by 100.
 };
 
 const knownValueFormatters = {
   bytes: unitsInBytes,
   nanoseconds: unitsInTime,
   count: unitsInCount,
+  percentage,
```
metalmatze, line 71: Need to remove this.

ui/packages/shared/utilities/src/index.ts
```diff
@@ -36,34 +36,39 @@ export const capitalize = (a: string): string =>
     .join(' ');
 
 interface Unit {
-  multiplier: number;
+  divisor: number;
```
metalmatze, line 39: We divide by this. Multiplier threw me off for a bit. Therefore updating.

ui/packages/shared/profile/src/MetricsGraph/MetricsTooltip/index.tsx
```diff
@@ -121,7 +121,7 @@ const MetricsTooltip = ({
                       <tr>
                         <td className="w-1/4">Value</td>
                         <td className="w-3/4">
-                          {valueFormatter(highlighted.valuePerSecond, sampleUnit, 5)}
+                          {valueFormatter(highlighted.valuePerSecond, sampleUnit, 5)}{' '}
```
monicawoj, line 124: I don't believe the space is necessary here.

ui/packages/shared/profile/src/ProfileMetricsGraph/index.tsx
```diff
@@ -45,7 +46,8 @@ export const useQueryRange = (
   client: QueryServiceClient,
   queryExpression: string,
   start: number,
-  end: number
+  end: number,
+  filterByFunction: string | undefined
```
monicawoj, line 50: same as above
```suggestion
  filterByFunction?: string
```

ui/packages/shared/profile/src/ProfileMetricsGraph/index.tsx
```diff
@@ -30,6 +30,7 @@ interface ProfileMetricsGraphProps {
   profile: ProfileSelection | null;
   from: number;
   to: number;
+  filterByFunction: string | undefined;
```
monicawoj, line 33: Instead of '| undefined' it should suffice to make this an optional prop like so:

```suggestion
  filterByFunction?: string;
```

ui/packages/shared/profile/src/ProfileSelector/index.tsx
```diff
@@ -201,13 +208,21 @@ const ProfileSelector = ({
...
             <MatchersInput
               queryClient={queryClient}
               setMatchersString={setMatchersString}
               runQuery={setQueryExpression}
               currentQuery={query}
             />
+            <div>
+              <Input
+                className="divider-y block w-full flex-1 rounded rounded-l-none bg-gray-50 px-2 py-2 text-sm outline-none focus:ring-indigo-800 dark:bg-gray-900"
+                placeholder="filter function name..."
```
monicawoj, line 221: Placeholder should be capitalized to match the remainder of our placeholders/text

```suggestion
                placeholder="Filter function name..."
```

ui/packages/shared/profile/src/ProfileSelector/index.tsx
```diff
@@ -201,13 +208,21 @@ const ProfileSelector = ({
...
+          <div className="flex w-full flex-1">
             <MatchersInput
               queryClient={queryClient}
               setMatchersString={setMatchersString}
               runQuery={setQueryExpression}
               currentQuery={query}
             />
+            <div>
+              <Input
+                className="divider-y block w-full flex-1 rounded rounded-l-none bg-gray-50 px-2 py-2 text-sm outline-none focus:ring-indigo-800 dark:bg-gray-900"
```
monicawoj, line 220: The current styles don't exactly match the styles of the remainder of the top bar (spacing, rounding). I think we should have the same spacing around this as the other items in the top bar. To do so, we should probably move the Input to outside of the div that is holding the MatchersInput element.

monicawoj: Unless we purposefully want it to look "attached" to the other filter input? In which case, we should modify that input to not be a text area. The bottom right corner of a textarea can be stretched/expanded. Once expanded the two filter inputs do not look related to one another.

monicawoj (commented):
Functionality wise looks good to me, but I think the styling/UX can be improved. As is, I'm not sure exactly whether the goal was to make the 'Filter by function' input directly next to the other query filter input or not. The styling seems a little off in terms of spacing and rounding of corners (rounded next to sharp). 

Additionally, if the user expands the query filter input, the filter by function input seems out of place.
Screenshot 2023-05-15 at 11 41 33

ui/packages/shared/profile/src/MetricsGraph/MetricsTooltip/index.tsx
metalmatze: Good catch! That's a left over of something I removed again.

ui/packages/shared/profile/src/ProfileSelector/index.tsx
metalmatze: Yeah, exactly. It's supposed to be attached to the matchers query. It's basically an additional field for the query. I don't think we can move the textarea to be an input. Maybe the other way around would work though. Make the filter input a textarea?

metalmatze: I guess given that it doesn't really match with the styles, I'll try to make them unattached for now.

monicawoj: yeah I think unattached is the best way to go for now!

metalmatze:
Resizing the matcher textarea now also looks less broken. :relieved: 

filter-size

monicawoj (approved):

PR #3139, opened 2023-05-16
build(container): add metadata labels

maxbrunet:
This will enrich the registry UI and tools like Renovate (changelog) with metadata

https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#labelling-container-images

Related to parca-dev/parca-agent#1633


kakkoyun (approved):
LGTM ❤️

PR #3143, opened 2023-05-17
deploy/pyrra: Update files to match what we run in demo.parca.dev.

metalmatze:
We haven't updated these example SLO config files in a while. 
Now, that we moved the latency SLO over to native histograms I took the opportunity. 

Check out a live example on: https://demo.pyrra.dev/?filter=%7Bteam=%22parca%22%7D


kakkoyun (approved):
LGTM 🎉 🌮

PR #3186, opened 2023-05-27
fix: serve everything under path prefix when set

maxbrunet:
Serve everything under path prefix when `--path-prefix` is set, including `/metrics` and `/debug/pprof`, and nothing directly from `/`


thorfour (approved):
LGTM

brancz:
I feel like this was this way for a reason before. I think the way it was intended to be used is to have a rewrite rule by any proxy in front of parca.

pkg/server/server.go
```diff
@@ -127,26 +127,17 @@ func (s *Server) ListenAndServe(ctx context.Context, logger log.Logger, addr str
 	grpc_health.RegisterHealthServer(srv, s.grpcProbe.HealthServer())
 
 	internalMux := chi.NewRouter()
-	if pathPrefix != "" {
-		internalMux.Mount(pathPrefix+"/api", grpcWebMux)
-	}
-	internalMux.Mount("/api", grpcWebMux)
```
maxbrunet, line 133: We could use a rewrite rule, but what's the purpose of this then? It feels like the other paths were overlooked. Otherwise, I'd remove this if-block and we do not change anything in the router when `pathPrefix` is set. WDYT?

pkg/server/server.go
```diff
@@ -127,26 +127,17 @@ func (s *Server) ListenAndServe(ctx context.Context, logger log.Logger, addr str
...
-			return
-		}
-		pprof.Index(w, r)
-	})
+	internalMux.Handle(pathPrefix+"/debug/pprof/*", http.StripPrefix(pathPrefix, http.HandlerFunc(pprof.Index)))
+	internalMux.Handle(pathPrefix+"/debug/pprof/fgprof", fgprof.Handler())
+	internalMux.HandleFunc(pathPrefix+"/debug/pprof/cmdline", pprof.Cmdline)
+	internalMux.HandleFunc(pathPrefix+"/debug/pprof/profile", pprof.Profile)
+	internalMux.HandleFunc(pathPrefix+"/debug/pprof/symbol", pprof.Symbol)
+	internalMux.HandleFunc(pathPrefix+"/debug/pprof/trace", pprof.Trace)
```
metalmatze, line 140: We can use `internalMux.Route(pathPrefix, func(r chi.Router) {}` and then register all those handlers with that passed `r` to not repeat ourselves adding the prefix manually.

metalmatze (commented):
I guess the idea was that because the API is part of our public API the prefix would be adjusted, whereas the internal APIs would stay the same. It would be more clear if those routers where served on different ports, but we have it on the same one.

Given we have it on the same port and use the same router in the end, I can see how being consistent with the prefix makes sense. :+1:

pkg/server/server.go
maxbrunet: Thanks! I wanted to do something like that, but did not know how

brancz:
Ah ok yes I think in the past we played around with separate ports. Lgtm then

PR #3196, opened 2023-05-29
fix(ui): list active targets only

### LINKED ISSUE #3195: UI shows droppped targets
image

I think it should only show "ACTIVE" targets like Prometheus.

https://demo.parca.dev/targets
https://demo.parca.dev/devel/targets

maxbrunet:
Fixes #3195


maxbrunet:
https://parca-58f9x06w7-parca-dev.vercel.app/targets looks better now imo

metalmatze (approved):
That makes sense, especially for scraped targets. :+1:

PR #3208, opened 2023-06-01
fix click, drag, and keydown behavior

monicawoj:
Resolves MetricsGraph issues regarding point selection, range selection, freezing the tooltip, and taking screenshots.

**Problem:**
We've had several reports of bugs when interacting with the MetricsGraph. Sometimes, the MetricsGraph goes into a mode where instead of clicking on a point it will zoom in, the MetricsGraph Tooltip doesn't work anymore either.

**Solution:**
Currently, we have special logic in place that freezes the Metrics Graph tooltip when the user holds down the 'Shift' key. However, 'Shift' is a special key that is used for certain shortcuts (such as taking a screenshots on Mac). The issue here is that when the shift key was pressed, followed by 'command' in order to take a screenshot, the application was paused as it entered screenshot mode and `isShiftDown` in our `useKeyDown` hook did not update on 'keyup'. The last captured event was 'keydown' of the shift key, therefore `isShiftDown` remained stuck in a 'down' state, even though it was no longer down.  Manually resetting `isShiftDown` state to false when the user hits the "Meta" key (in Mac = command, in Windows = Control) was able to solve this issue. 

Additionally, clicks outside of the tooltip have been disabled when shift is held down in order to ensure that only the tooltip can be clicked while the tooltip is frozen. This was able to resolve issues around unexpected zooming.

### DIFF
--- a/ui/packages/shared/components/src/KeyDownContext/index.tsx
+++ b/ui/packages/shared/components/src/KeyDownContext/index.tsx
@@ -15,10 +15,12 @@ import {ReactNode, createContext, useContext, useEffect, useMemo, useState} from
 
 export interface KeyDownState {
   isShiftDown: boolean;
+  keys: (string | number)[];
 }
 
 const DEFAULT_VALUE = {
   isShiftDown: false,
+  keys: [],
 };
 
 const KeyDownContext = createContext<KeyDownState>(DEFAULT_VALUE);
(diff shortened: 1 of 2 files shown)


metalmatze (approved):
Very nice! 👏 
I couldn't trigger it anymore now! 🎉 

Nice idea combining it with the Meta key 👍

brancz (approved):

manojVivek:
Great find, Monica! 👏🏼 👏🏼 👏🏼

PR #3230, opened 2023-06-05
fix(config): fix reloading config from Kubernetes ConfigMap

maxbrunet:
This properly handle reloading the configuration from a Kubernetes ConfigMap which uses a symlink, a case I had initially overlooked.

In this case, the reload is triggered by the deletion of the old version of the config, not the symlink deletion/re-creation as fsnotify follows them when creating the watcher.


kakkoyun (approved):
LGTM ❤️ 💚 💙 💛 💜

PR #3267, opened 2023-06-12
Fix Callgraph crash

monicawoj:
Lazy loading of the Callgraph component in the @parca/profile package appears to be making the app crash as soon as the user selects the 'Callgraph' view. 

The bug only appears after building the code, which indicates that it is likely a problem with NextJS's behavior of React.Lazy loaded components. Lazy loading of this component has thus been removed. 

If we decide that the Callgraph component _does_ in fact require lazy loading, we can explore optimization using next/dynamic (https://nextjs.org/docs/pages/building-your-application/optimizing/lazy-loading).


yomete (approved):
lgtm!

PR #3270, opened 2023-06-12
debuginfo/store: Use tracer.Start instead of SpanFromContext

kakkoyun:

Since parca-agent started to create spans from the client side, `SpanFromContext` has started to create problems. 
Even though operations successfully finish since OTel SDK propagates the context from/to the client-side agent receives a timed-out/closed context.
This was causing errors on the agent side. 
The PR ensures a new context is created, and the child context is closed when the operation is finished without touching the parent(incoming) context.

Fixes https://github.com/parca-dev/parca-agent/issues/1741


metalmatze:
I wonder if there are some implications here. Would we break the gRPC interceptor to our service implementation chain?

PR #3276, opened 2023-06-13
ui: Automatically use system settings for Parca's theme

yomete:
After demoing this during Parca Office hours, we came to the consensus that the system settings theme mode should be chosen automatically, as opposed to the previous method where the user had to first select system settings.


manojVivek (approved):
Lgtm!

PR #3281, opened 2023-06-14
Remove usage of lerna bootstrap

yomete:
This PR removes the usage of lerna bootsrap and simply replaces with yarn install as documented [here](https://lerna.js.org/docs/legacy-package-management).


manojVivek (approved):
Looks great!

PR #3282, opened 2023-06-14
ui: Add loading indicator when downloading pprof

### LINKED ISSUE #3029: Only one pprof download at a time (and add loading indicator)
With particularly large amounts of profiling data it's easy to click the pprof download button multiple times, especially as it's unclear if a download is already in progress.

We should block a user from clicking the button twice if a download is already in progress, and show a loading indicator.

yomete:
Resolves #3029


manojVivek (approved):
Lgtm!

PR #3326, opened 2023-06-20
ui/shared/profile/IcicleGraphArrow: Create arrow-based IcicleGraph

metalmatze:
These are the frontend changes to read flame graphs from arrow records. 
The backend changes are in https://github.com/parca-dev/parca/pull/3069 which this PR depends on.


manojVivek (approved):
Lgtm overall!

ui/packages/shared/profile/src/ProfileIcicleGraph/IcicleGraphArrow/index.tsx
```diff
@@ -0,0 +1,220 @@
...
+    return <></>;
+  }
+
+  return (
+    <div onMouseLeave={() => dispatch(setHoveringNode(undefined))}>
+      <select
+        className="rounded-md border bg-gray-50 px-3 py-2 text-sm shadow-sm focus:border-indigo-500 focus:outline-none focus:ring-1 focus:ring-indigo-500 dark:bg-gray-900"
+        onChange={e => setSortBy(e.target.value)}
+      >
+        <option value="function_name">Function Name</option>
```
yomete, line 168: Shall we place the select dropdown next to the `Reset View` button and add a text in front. Like this:

"Reset View"  "Sort by: `<Select />`"

ui/packages/shared/profile/src/ProfileView/index.tsx
```diff
@@ -52,7 +53,8 @@ type NavigateFunction = (path: string, queryParams: any, options?: {replace?: bo
 
 export interface FlamegraphData {
   loading: boolean;
-  data?: Flamegraph;
```
yomete, line 55: We might need to cater for this removed prop for the grafana plugin as that's the reason for the typing error [here](https://github.com/parca-dev/parca/actions/runs/5457111282/jobs/9930779290?pr=3326).

yomete (approved):
lgtm overall! just a few things to work over that i mentioned.

ui/packages/shared/profile/src/ProfileView/index.tsx
metalmatze: aah, I see. Let's not rename it then, but only add the table one for now. I think there is a bigger refactoring we should do anyway.

metalmatze:
There are some longer-running CI checks still running, but I would like to get this in before we have another conflict. Happy to follow up with PRs if anything is found.

PR #3331, opened 2023-06-21
ui: Redundant query request fix

### LINKED ISSUE #2823: Multiple `Query` requests are made when node trimming is enabled.
Screenshot_from_2023-03-22_10-45-54

manojVivek:
Contains the following:
- [x] Fix for the redundant query request from the UI due to the updated `nodeThreshold`.
- [x] Enables node trimming by default.
- [x] Fix the `@parca/hooks` package version drift from the the consumed version. 
- [x] Fix the bug with the bigint `linearScale` scale function. 

Resolves #2823


brancz:
Works perfectly now! Thanks for taking care of this!

Looks like some tests failing otherwise lgtm

ui/packages/shared/utilities/src/bigint.ts
```diff
@@ -15,7 +15,7 @@
  * Divides two bigints and returns a number with two decimal places
  */
 export const divide = (a: bigint, b: bigint): number => {
-  return Number((a * 10000n) / b) / 10000;
+  return Number((a * 1000000n) / b) / 1000000;
```
yomete, line 18: out of curiosity, why did the value being multiplied by change here?

yomete (approved):
lgtm! just a question about a change in value below.

ui/packages/shared/utilities/src/bigint.ts
manojVivek: Since `bigint` divisions don't give any decimals, we premultiply it by 1M and then divide the result again by 1M to get the result with decimals. 

I increased the number from 10K to 1M to get more precision to the decimal part, which was needed for the `linearScale` function. 

Hope it answers the question.

PR #3334, opened 2023-06-21
Add meticulous.ai to UI

metalmatze:
This adds meticulous AI for development and preview environments to record


yomete:
I'm not sure this would work in our `app` directory as `_document.tsx` works only in a Next.js app. We might need to use JS to append the script conditionally.

Que3216:
If you want to control whether the script gets downloaded or not conditionally in javascript you can use our recorder-loader package. Your https://github.com/parca-dev/parca/blob/main/ui/packages/app/web/src/index.tsx#L21 would then become:

```tsx
import {createRoot} from 'react-dom/client';

import App from './App';
import reportWebVitals from './reportWebVitals';

async function startApp() {
    if (shouldEnableMeticulousRecorder(window.location))
        // Start the Meticulous recorder before you initialise your app.
        // Note: all errors are caught and logged, so no need to surround with try/catch
        await tryLoadAndStartRecorder({
          projectId: 'lgBUqjZEtPYwCUmwxdv0IEiMsaPwDJWo9AJHvbzX',
        });
    }
    
    // Initalise app after the Meticulous recorder is ready: Meticulous needs to be initialised first,
    // so it can capture the initial network requests
    const container = document.getElementById('root');
    const root = createRoot(container!);
    root.render(<App />);
    
    // If you want to start measuring performance in your app, pass a function
    // to log results (for example: reportWebVitals(console.log))
    // or send to an analytics endpoint. Learn more: https://bit.ly/CRA-vitals
    reportWebVitals();
}


startApp();
```

Or, if you want to record all sessions on all environments you can add the script tag to your https://github.com/parca-dev/parca/blob/main/ui/packages/app/web/public/index.html (or use different HTML files for different deploys, or something like https://github.com/jantimon/html-webpack-plugin, if you want the script to only be present on some environments).

ui/packages/app/web/src/index.tsx
```diff
@@ -11,16 +11,34 @@
...
 import {createRoot} from 'react-dom/client';
 
 import App from './App';
 import reportWebVitals from './reportWebVitals';
 
-const container = document.getElementById('root');
-const root = createRoot(container!);
-root.render(<App />);
+const shouldEnableMeticulousRecorder =
+  process.env.NODE_ENV === 'development' || process.env.VERCEL_ENV === 'preview';
```
maxbrunet, line 21: Should the `isDevMode` utility be used?

https://github.com/parca-dev/parca/blob/df4fafb8a6639867aae35f1790b2df4e0d279d00/ui/packages/shared/utilities/src/index.ts#L96-L98

It be nice to add `|| process.env.VERCEL_ENV === 'preview'` to it as well, it'd fix the path prefix in the preview :)

brancz: I think largely we want previews to be as similar to prod deployments as possible though, no? So changing is dev might cause this to blur further.

yomete: Agreed. Modifying the `isDevMode` utility would change how that behaves .

maxbrunet: It would not change much in behavior, the function is only used here:

https://github.com/parca-dev/parca/blob/830413a74b8d89f5aa9f35ba20fe3d4b4837b883/ui/packages/app/web/src/App.tsx#L76-L78

Maybe rename it to `isDevModeOrPreview`?

The idea would be to fix this kind of links in the preview:
https://parca-hn9um2wya-parca-dev.vercel.app/%7B%7B.PathPrefix%7D%7D/targets

Double check if it'd be the fix tho, something is off in the prefix handling for dev/preview, but I understand very little about React :smile:

yomete: Ahh I'm understanding you now. Had no idea that was happening, I'll try it out to see if it's a fix

yomete:
Targets page still not working on Preview, but i'll dedicate some time later today to look at.

PR #3363, opened 2023-06-26
ui: Fixes the bug where flamegraph goes blank for some instrumented profiles 

manojVivek:
Root cause: This happens when the `linearScales`'s domain range is so high compared to the device width range, making the `rate` a `0`, so no nodes in the flame graph are rendered. 

Fixed it by increasing the multiple to the same number of digits as `Number.MAX_SAFE_INTEGER`, so that the `rate` is not rounded off to `0`.


brancz (approved):
Nice find!

PR #3396, opened 2023-07-05
debuginfo: Add traces

kakkoyun:
- Extract tracer to its own package
- Add a TracingBucket wrapper
- Configure gRPC and HTTP clients propagate context
- Upgrade FrostDB
- Upgrade thanos-io/objstore

## Test Plan
```
$ cd script/local-tracing && docker compose up -d
$ ./bin/parca --otlp-address=127.0.0.1:4317
```
CleanShot 2023-07-06 at 09 40 10

CleanShot 2023-07-13 at 16 04 58

xref: https://github.com/thanos-io/objstore/pull/61
xref: https://github.com/polarsignals/frostdb/pull/469


pkg/parca/parca.go
```diff
@@ -357,10 +390,16 @@ func Run(ctx context.Context, logger log.Logger, reg *prometheus.Registry, flags
 
 	var debuginfodClient debuginfo.DebuginfodClient = debuginfo.NopDebuginfodClient{}
 	if len(flags.Debuginfod.UpstreamServers) > 0 {
-		httpDebugInfoClient, err := debuginfo.NewHTTPDebuginfodClient(logger, flags.Debuginfod.UpstreamServers, &http.Client{
-			Transport: promconfig.NewUserAgentRoundTripper(fmt.Sprintf("parca.dev/debuginfod-client/%s", version), http.DefaultTransport),
-			Timeout:   flags.Debuginfod.HTTPRequestTimeout,
-		})
+		httpDebugInfoClient, err := debuginfo.NewHTTPDebuginfodClient(logger, flags.Debuginfod.UpstreamServers,
```
kakkoyun, line 393: HTTP GET error in the screenshot comes from here.

pkg/parca/parca.go
```diff
@@ -357,10 +390,16 @@ func Run(ctx context.Context, logger log.Logger, reg *prometheus.Registry, flags
...
+			&http.Client{
+				Transport: otelhttp.NewTransport(
```
kakkoyun, line 395: HTTP GET error in the screenshot comes from here.

pkg/debuginfo/debuginfod.go
```diff
@@ -182,6 +184,10 @@ func (c *HTTPDebuginfodClient) request(ctx context.Context, u url.URL, buildID s
 	// this request will result in a binary object that contains the customary .*debug_* sections.
 	u.Path = path.Join(u.Path, "buildid", buildID, "debuginfo")
 
+	// Uses the default tracer provider and propagator that's set in tracer package,
+	// or from the span context passed in.
+	ctx = httptrace.WithClientTrace(ctx, otelhttptrace.NewClientTrace(ctx))
```
kakkoyun, line 189: This might be cleaner than the transport wrapper. This is what we used on the agent side. I wonder how this would interact with the transport wrapper.

brancz: Oh nice I didn't see this. Should we maybe then use the [`WithTracerProvider`](https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace#WithTracerProvider) so we don't use the global one?

pkg/objectstore/bucket.go
```diff
@@ -0,0 +1,260 @@
...
+	"github.com/thanos-io/objstore/providers/oci"
+	"github.com/thanos-io/objstore/providers/oss"
+	"github.com/thanos-io/objstore/providers/s3"
+	"github.com/thanos-io/objstore/providers/swift"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+	"gopkg.in/yaml.v2"
+)
+
+func NewBucket(tracer trace.Tracer, logger log.Logger, confContentYaml []byte, reg prometheus.Registerer, component string) (objstore.InstrumentedBucket, error) {
```
brancz, line 46: Since we're gonna want this in at least 3 repos, I'd suggest we propose this upstream.

For what it's worth I'm not a fan of `NewBucket` returning fully instrumented things, that makes assumptions that are not universally true. I think it's better for the caller to make the decision whether to instrument or not.

I would say here we just have the `NewTracingBucket` and defer the instrumentation decision to the caller.

pkg/debuginfo/debuginfod.go
```diff
@@ -184,10 +188,7 @@ func (c *HTTPDebuginfodClient) request(ctx context.Context, u url.URL, buildID s
 	// this request will result in a binary object that contains the customary .*debug_* sections.
 	u.Path = path.Join(u.Path, "buildid", buildID, "debuginfo")
 
-	// Uses the default tracer provider and propagator that's set in tracer package,
-	// or from the span context passed in.
-	ctx = httptrace.WithClientTrace(ctx, otelhttptrace.NewClientTrace(ctx))
-
+	ctx = httptrace.WithClientTrace(ctx, otelhttptrace.NewClientTrace(ctx, otelhttptrace.WithTracerProvider(c.tp)))
```
brancz, line 191: While I like this a lot I'm currently re-working this code pretty significantly, can we revert this here, and I promise I'll add this instrumentation! :D

kakkoyun: Removed those changes 👍

pkg/objectstore/bucket.go
kakkoyun: https://github.com/thanos-io/objstore/pull/61

PR #3415, opened 2023-07-06
pkg/query: TestGenerateFlamegraphArrowWithInlined

metalmatze:
Turns out we already inlined almost correctly. pprof just stores the lines in reverse order, just like the locations too. Flipping these around did the trick.

Comparing the arrow and table flame graphs the inlining seems fine now.


brancz (approved):
Nice!

PR #3433, opened 2023-07-11
pkg/query: Support unsymbolized profiles with arrow fg

metalmatze:
To support adding rows with just addresses, we'd have had to copy a lot of code. This was the perfect opportunity to finally introduce a better abstract - the `flamegraphBuilder`. 
There are some other pieces we may consider moving into this struct/builder in the future. I can also see having something like a `flamegraphBuilder.NewRecord()` to wrap some of the internals even more. Overall, these are minor improvements though. 

The UI needed some small tweaks to properly support function names being `null`.


brancz (approved):
lgtm but pleeeeeeease let's refactor those continue statements

metalmatze:
Yep, I'm sure they are going sooner than later too.

PR #3440, opened 2023-07-12
pkg/query: Correctly sum up total cumulative value

metalmatze:
We were counting inlined functions additionally.


brancz (approved):
I guess as long as this is consistent with our previous flamegraph implementation I'm happy with it, but I'll be honest I don't understand this change.

metalmatze:
It is now. 
The real root cause is folded away by GitHub, it's the lines above:
```
for k := len(location.Lines) - 1; k >= 0; k-- {
				line := location.Lines[k]
```

We were adding the cumulative value multiple times if the leaf is an inlined function.

PR #3448, opened 2023-07-13
pkg/query/flamegraph_arrow: Trim children if below trim threshold

metalmatze:
To get rid of entries in the flame graph that are smaller than a pixel (and thus can't be rendered) the code iterates over all rows and checks a row's children cumulative values against its own cumulative value. If the difference is so drastic we remove that child.

Not yet implemented is rewriting the arrow record after trimming. Given the little overhead we can always improve the rendering by trimming too small children. Based on some heuristic (like if more than 1% is trimmed) we may consider recreating the arrow record with just the necessary rows. This would mostly be a bandwidth optimization.

The differences with the added code won't be noticeable for now.
```
name      old time/op    new time/op    delta
Query-24    24.7ms ± 4%    25.7ms ± 8%    ~     (p=0.421 n=5+5)

name      old alloc/op   new alloc/op   delta
Query-24    14.1MB ± 0%    14.2MB ± 0%  +0.30%  (p=0.032 n=5+5)

name      old allocs/op  new allocs/op  delta
Query-24      119k ± 0%      120k ± 0%    ~     (p=0.095 n=5+5)
```

### DIFF
--- a/pkg/query/flamegraph_arrow.go
+++ b/pkg/query/flamegraph_arrow.go
@@ -338,17 +338,47 @@ func generateFlamegraphArrowRecord(ctx context.Context, mem memory.Allocator, tr
 	// Now we set/overwrite the cumulative value for the root row (which is always the 0 row in our flame graphs).
 	fb.builderCumulative.Set(0, cumulative)
 
+	// Set the children for the root row to the root rows we have tracked.
+	fb.children[0] = rootsRow
+
+	// Iterate over all rows.
+	// Read their cumulative values and compare that against the childrens cumulative values.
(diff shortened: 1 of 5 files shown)


brancz (commented):
I'm not sure the strategy is right here. The only thing this does is remove the children that shouldn't be rendered from the child list of parents. That means that the rows themselves stay intact, therefore still bloating the result size of the resulting arrow record. I think for trimming to be effective the only thing we can do is re-write the whole record skipping the trimmed rows.

metalmatze:
Superseded by https://github.com/parca-dev/parca/pull/3722

PR #3459, opened 2023-07-17
compress the flamegraph record with lz4

thorfour:


metalmatze (approved):
Yikes! Thanks for finding this! :100:

PR #3471, opened 2023-07-18
pkg/query: Arrow flamegraph keeps labels unless they differ

metalmatze:
This change makes the arrow flamegraph track the labels for each row in a `map[int]map[string]string`. The `int` row is used as the key and its value is the sample's `map[string]string`. 

If the functions/location should be merged we now compare the existing pprof labels. If there are none, we ignore this step. If either side has no pprof labels the side that has labels is chosen. If both the previous stack had pprof labels and this new one we want to merge has pprof labels we compare them.

For now, this doesn't implement pure set theory intersection. It should be actually closer to [Symmetric Difference](https://en.wikipedia.org/wiki/Symmetric_difference). Essentially we keep all label key-value pairs around that are unique. **Only if the key is the same, but their values change, we remove the entire key-value pair.**

### DIFF
--- a/pkg/query/flamegraph_arrow.go
+++ b/pkg/query/flamegraph_arrow.go
@@ -83,6 +83,8 @@ func GenerateFlamegraphArrow(ctx context.Context, mem memory.Allocator, tracer t
 	}, cumulative, nil
 }
 
+const labelsDiffer = "THESELABELSDIFFER"
+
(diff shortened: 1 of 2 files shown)


metalmatze:
This has been reworked and updated to actually compute the intersection of all labels of a row.

brancz (approved):

PR #3479, opened 2023-07-19
Add SortBy and GroupBy actions for Icicle Graphs

metalmatze:
This PR adds two additional UI components to the arrow icicle graph. 
image

One is a **SortBy** dropdown to either sort by the function name (the default and what we currently only had) or cumulative value. 
The other component is a dropdown that toggles **GroupBy** function name or pprof labels or no grouping at all. 

All state is **stored in the URL**. Sharing a profile keeps the group by and sort by order. 

### No grouping
image
### Function name (the default)
image
### (pprof) Labels
image
### Function and Labels
image

### Function name (the default)

This sorts alphabetically by function names: 
image
### Cumulative value

This sorts by cumulative value in descending order. This makes the biggest values show up to the left and the smallest to the right: 
image


ui/packages/shared/profile/src/ProfileIcicleGraph/index.tsx
```diff
@@ -57,6 +64,46 @@ const ProfileIcicleGraph = ({
...
+    if (storeGroupBy !== undefined) {
+      if (typeof storeGroupBy === 'string') {
+        return [storeGroupBy];
+      }
+      return storeGroupBy as string[];
+    }
+    return [FIELD_FUNCTION_NAME];
+  }, [storeGroupBy]);
+
+  console.log('groupBy', storeGroupBy, groupBy);
```
manojVivek, line 91: Looks like this log statement got slipped through.

ui/packages/shared/profile/src/ProfileIcicleGraph/index.tsx
```diff
@@ -94,18 +141,29 @@ const ProfileIcicleGraph = ({
...
+      <div className="flex w-full justify-end gap-2 pb-2">
+        <div className="flex w-full items-center justify-between space-x-2">
+          {table !== undefined && (
+            <>
+              <GroupByDropdown groupBy={groupBy} toggleGroupBy={toggleGroupBy} />
+              <SortBySelect compareMode={compareMode} sortBy={localSortBy} setSortBy={setSortBy} />
+            </>
+          )}
+          <div>
+            <label>&nbsp;</label>
```
manojVivek, line 153: This can be removed by adding a `gap-2` class to the parent element.

ui/packages/shared/profile/src/ProfileIcicleGraph/index.tsx
```diff
@@ -57,6 +64,46 @@ const ProfileIcicleGraph = ({
...
+    },
+    [setStoreGroupBy]
+  );
+  const toggleGroupBy = useCallback(
+    (key: string): void => {
+      groupBy.includes(key)
+        ? setGroupBy(groupBy.filter(v => v !== key)) // remove
+        : setGroupBy([...groupBy, key]); // add
+    },
+    [groupBy]
```
manojVivek, line 105: Is it intentional to not have the `setGroupBy` in the dependencies?

manojVivek (approved):
Great looking UI! 👏🏼 

Just a couple of minor nits and the linter comments, otherwise looks good from the UI perspective.

ui/packages/shared/profile/src/ProfileIcicleGraph/index.tsx
metalmatze: Oh, I thought I caught them all :smiley:

ui/packages/shared/profile/src/ProfileIcicleGraph/index.tsx
metalmatze: Seems like `gap-2` on its own isn't helpful but in combination with `inline-block` it pushes the `Reset View` button down. In fact, `inline-block` is all that's needed.

ui/packages/shared/profile/src/ProfileIcicleGraph/index.tsx
metalmatze: Yes, and further down as well. If I add those functions/callbacks to the effect dependencies the component goes into an endless re-render. Happy to brainstorm how to fix it. Already tried some `useCallback` and `useMemo` approaches yesterday.

metalmatze: Actually, adding `setGroupBy` seems to be fine. It's the `useEffect` further down that ends up looping.

ui/packages/shared/profile/src/ProfileIcicleGraph/index.tsx
```diff
@@ -94,18 +141,29 @@ const ProfileIcicleGraph = ({
...
+              variant="neutral"
+            >
+              Reset View
+            </Button>
+          </div>
+        </div>
+      </div>
     );
-  }, [setNewCurPath, curPath, setActionButtons]);
+  }, [setNewCurPath, curPath, setActionButtons, table, compareMode, localSortBy, groupBy]);
```
metalmatze, line 166: Adding `setSortBy` or `toggleGroupBy` makes the component re-render endlessly. Both of those callbacks are wrapped with `useCallback`.

brancz (approved):
backend/API changes lgtm

metalmatze:
Thanks a lot for fixing those re-renders @manojVivek :clap:

PR #3486, opened 2023-07-20
ui: Enabling navbar path prefix for preview environments

manojVivek:


manojVivek:
Merging to test this on demo.parca.dev/devel. Please leave your comments.

PR #3516, opened 2023-07-27
flamegraph_arrow: Inject artifical label node

metalmatze:
On the backend, we check every sample for its pprof labels.
If it has pprof labels we either inject a artifical node into the stack containing the pprof labels as JSON for the function name, or we merge the stack into the already existing same pprof label node. 

In the frontend, instead of showing `{"thread_name":"foo"}` we're going to parse the JSON and display it as `thread_name="foo"` in a Prometheus fashion. Given that each injected pprof label node is unique the drilling down into the icicle graph by function name now works flawlessly again. :+1: 

Screenshot from 2023-07-27 20-12-51

### DIFF
--- a/pkg/query/flamegraph_arrow.go
+++ b/pkg/query/flamegraph_arrow.go
@@ -86,6 +86,11 @@ func generateFlamegraphArrowRecord(ctx context.Context, mem memory.Allocator, tr
 	for _, f := range aggregate {
 		aggregateFields[f] = struct{}{}
 	}
+	// this is a helper as it's frequently accessed below
+	aggregateLabels := false
+	if _, found := aggregateFields[FlamegraphFieldLabels]; found {
+		aggregateLabels = true
+	}
 
 	totalRows := int64(0)
 	for _, r := range p.Samples {
@@ -264,26 +269,24 @@ func generateFlamegraphArrowRecord(ctx context.Context, mem memory.Allocator, tr
 				}
 			}
 
-			sampleLabelsKey := ""
 			if _, aggregateLabels := aggregateFields[FlamegraphFieldLabels]; aggregateLabels && len(sampleLabels) > 0 {
 				lsbytes = lsbytes[:0]
 				lsbytes = MarshalStringMap(lsbytes, sampleLabels)
-				sampleLabelsKey = bytesToString(lsbytes)
 
 				sampleLabelRow := row
-				if _, ok := rootsRow[sampleLabelsKey]; ok {
-					sampleLabelRow = rootsRow[sampleLabelsKey][0]
+				if _, ok := rootsRow[unsafeString(lsbytes)]; ok {
+					sampleLabelRow = rootsRow[unsafeString(lsbytes)][0]
 					compareRows = compareRows[:0] //  reset the compare rows
 					// We want to compare against this found label root's children.
-					rootRow := rootsRow[sampleLabelsKey][0]
+					rootRow := rootsRow[unsafeString(lsbytes)][0]
 					compareRows = append(compareRows, fb.children[rootRow]...)
 					fb.addRowValues(r, sampleLabelRow, i) // adds the cumulative and diff values to the existing row
 				} else {
-					err := fb.AppendLabelRow(r, sampleLabelRow, sampleLabelsKey, sampleLabels, i)
+					err := fb.AppendLabelRow(r, sampleLabelRow, unsafeString(lsbytes), sampleLabels, i)
 					if err != nil {
 						return nil, 0, 0, 0, fmt.Errorf("failed to inject label row: %w", err)
 					}
-					rootsRow[sampleLabelsKey] = []int{sampleLabelRow}
+					rootsRow[unsafeString(lsbytes)] = []int{sampleLabelRow}
 				}
 				fb.parent.Set(sampleLabelRow)
 				row = fb.builderCumulative.Len()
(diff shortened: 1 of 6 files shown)


pkg/query/flamegraph_arrow.go
```diff
@@ -277,9 +305,9 @@ func generateFlamegraphArrowRecord(ctx context.Context, mem memory.Allocator, tr
 				if !r.Lines.IsValid(j) || llOffsetEnd-llOffsetStart <= 0 {
 					if isRoot {
 						compareRows = compareRows[:0] //  reset the compare rows
-						compareRows = append(compareRows, rootsRow...)
+						compareRows = append(compareRows, rootsRow[sampleLabelsKey]...)
 						// append this row afterward to not compare to itself
-						fb.parent.Reset()
+						// fb.parent.Reset()
```
brancz, line 310: is this intentional?

pkg/query/flamegraph_arrow.go
```diff
@@ -173,9 +174,10 @@ func generateFlamegraphArrowRecord(ctx context.Context, mem memory.Allocator, tr
 	cumulative := int64(0)
 	// This keeps track of the max depth of our flame graph.
 	maxHeight := int32(0)
-	// This keeps track of the root rows.
+	// This keeps track of the root rows indexed by the labels string.
+	// If the stack trace has no labels, we use the empty string as the key.
 	// This will be the root row's children, which is always our row 0 in flame graphs.
-	rootsRow := []int{}
+	rootsRow := map[string][]int{}
```
brancz, line 180: is there any other way other than a map that we could do this? profiling data already suggests that a pretty large amount of time is spent in map accesses

pkg/query/flamegraph_arrow.go
```diff
@@ -261,6 +264,31 @@ func generateFlamegraphArrowRecord(ctx context.Context, mem memory.Allocator, tr
...
+					// We want to compare against this found label root's children.
+					rootRow := rootsRow[sampleLabelsKey][0]
+					compareRows = append(compareRows, fb.children[rootRow]...)
+					fb.addRowValues(r, sampleLabelRow, i) // adds the cumulative and diff values to the existing row
+				} else {
+					err := fb.AppendLabelRow(r, sampleLabelRow, sampleLabelsKey, sampleLabels, i)
+					if err != nil {
+						return nil, 0, 0, 0, fmt.Errorf("failed to inject label row: %w", err)
+					}
+					rootsRow[sampleLabelsKey] = []int{sampleLabelRow}
```
brancz, line 286: this is unsafe, when storing something the string is kept by the map, it's only safe to use for accesses

I would recommend to just use `bytesToString` directly inline wherever there are accesses to the map are done since the "conversion" is extremely cheap and it's harder to do something wrong.

pkg/query/flamegraph_arrow.go
```diff
@@ -306,7 +334,7 @@ func generateFlamegraphArrowRecord(ctx context.Context, mem memory.Allocator, tr
 
 					if isRoot {
 						// We aren't merging this root, so we'll keep track of it as a new one.
-						rootsRow = append(rootsRow, row)
+						rootsRow[sampleLabelsKey] = append(rootsRow[sampleLabelsKey], row)
```
brancz, line 337: same here, this is unsafe

pkg/query/flamegraph_arrow.go
```diff
@@ -357,9 +386,9 @@ func generateFlamegraphArrowRecord(ctx context.Context, mem memory.Allocator, tr
 						compareRows = compareRows[:0]
 					}
 
-					if isRoot {
+					if isRoot && sampleLabelsKey == "" {
 						// We aren't merging this root, so we'll keep track of it as a new one.
-						rootsRow = append(rootsRow, row)
+						rootsRow[sampleLabelsKey] = append(rootsRow[sampleLabelsKey], row)
```
brancz, line 391: this as well

pkg/query/flamegraph_arrow.go
```diff
@@ -576,6 +607,52 @@ func (fb *flamegraphBuilder) appendRow(
...
+		// We need to grow the children slice, so we'll do that here.
+		// We'll double the capacity of the slice.
+		newChildren := make([][]int, len(fb.children)*2)
+		copy(newChildren, fb.children)
+		fb.children = newChildren
+	}
+	// Add this label row to the root row's children.
+	fb.children[0] = append(fb.children[0], row)
+	//// Add the next row as child of this label row.
+	//fb.children[row] = append(fb.children[row], row+1)
```
brancz, line 623: leftover?

pkg/query/flamegraph_arrow.go
```diff
@@ -622,3 +699,10 @@ keys:
 
 	return intersection
 }
+
+func bytesToString(b []byte) string {
```
brancz, line 703: I would make sure it's clear from the caller's perspective that this is unsafe. I like to call these `unsafeString` for example.

pkg/query/flamegraph_arrow.go
metalmatze: We want this to be unique per pprof label JSON string. 
We want to be able to find the individual rows under a specific label string. 
I'm not really sure if we have some better primitives right now. We would need to investigate some other data structures altogether. 
Happy to do this as a follow-up with some more profiling/benchmarking data.

pkg/query/flamegraph_arrow.go
metalmatze: Nope, I missed this. Will remove it.

pkg/query/flamegraph_arrow.go
brancz: Sounds good as a follow-up and based on data.

pkg/query/flamegraph_arrow.go
brancz: I don't think this has been resolved?

pkg/query/flamegraph_arrow.go
```diff
@@ -576,11 +616,57 @@ func (fb *flamegraphBuilder) appendRow(
...
+	fb.builderMappingLimit.AppendNull()
+	fb.builderMappingOffset.AppendNull()
+	fb.builderMappingFile.AppendNull()
+	fb.builderMappingBuildID.AppendNull()
+	fb.builderLocationAddress.AppendNull()
+	fb.builderLocationFolded.AppendNull()
+	fb.builderLocationLine.AppendNull()
+	fb.builderFunctionStartLine.AppendNull()
+
+	_ = fb.builderFunctionName.AppendString(labelKey)
```
brancz, line 644: we should actually check and return this error

brancz (approved):

PR #3564, opened 2023-08-05
pkg/query: Add fast map of strings to json marshaling

brancz:
This has a ~82% improvement compared to the previously used method.

```
$ benchstat stdlib.txt ours.txt
name        old time/op  new time/op  delta
MarshalMap   412ns ± 1%    73ns ± 0%  -82.34%  (p=0.008 n=5+5)
```

Which results in an overall improvement of ~44% for real life profiling data that has labels:

```
$ benchstat old.txt new.txt
ArrowFlamegraph-10   289ms ± 1%   161ms ± 1%  -44.23%  (p=0.008 n=5+5)
```

Let's profile live stream where we explored this: https://www.youtube.com/watch?v=4hZtHsOWKKk (turns out we incorrectly measured and the improvement was actually significantly larger than initially thought)


metalmatze:
Very cool! I knew this was eventually going to happen, but didn't expect it so soon :smiley: 
Very happy that we are not using reflection anymore now!

PR #3572, opened 2023-08-07
feat(snap): add config options to send profiles to a remote store

jnsgruk:
Add a config option to the Parca snap to simplify the onboarding to Parca Cloud.

This change introduces a new snap config option named remote-store-bearer-token, and sets the default store address to grpc.polarsignals.com:443.

When a token is set, any other config about storage persistence or memory is ignored, and Parca's mode is set to `scraper-only`.

### DIFF
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -34,9 +34,17 @@ if [[ -z "$storage_persist" ]]; then
     snapctl set enable-persistence=false
 fi
 
-port="$(snapctl get port)"
-if [[ -z "$port" ]]; then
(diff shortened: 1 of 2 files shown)


jnsgruk:
While testing this I noticed that we were still using the deprecated `--port` option, so I've added a commit that migrates that to a new config option named `http-address`. Existing installs shouldn't notice a change.

brancz (approved):

jnsgruk:
@brancz can we get this merged so I can build this into the charm & relations? :)

PR #3590, opened 2023-08-09
*: Allow uploading debuginfo sources

brancz:
This will allow building a component to view source code with profiling data inline similar to pprof's "source" feature. First step to do that is to have the sources available, and this patch allows uploading the sources associated with a piece of debuginfo.


metalmatze (approved):
Nice!
I love the idea of where this is going! :rocket:

PR #3591, opened 2023-08-09
ui: UI Build failure fix

manojVivek:
- [x] Fixed the babel warning
- [x] Revert the recent `eslint` package upgrade

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -52,6 +52,11 @@
       "description": "Disable tailwindcss upgrades",
       "matchPackageNames": ["tailwindcss"],
       "enabled": false
+    },
+    {
+      "description": "Disable eslint upgrades",
+      "matchPackageNames": ["eslint"],
+      "enabled": false
     }
   ],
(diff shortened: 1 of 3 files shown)


metalmatze:
Why do we need to revert eslint? Do the plugins not yet work with that newer version? 
How can we make sure renovate doesn't instantly overwrite this revert?

manojVivek:
> Why do we need to revert eslint? Do the plugins not yet work with that newer version? 

It looks like the newer version of the ESLint and the `react-scripts` are incompatible. `react-scripts` is right now in an unmaintained state, last released a year ago, should be the reason for this. 

> How can we make sure renovate doesn't instantly overwrite this revert?

This works in my local, just waiting to see if this passes in the CI so that I can add the necessary renovatebot config to prevent upgrading this package.

metalmatze:
Ok, great! The JS/TS parts are passing. I'll merge this ignoring the running but unrelated CI checks.

PR #3593, opened 2023-08-09
ci: Scheduling renovate merges between midnight and 4am UTC

manojVivek:


metalmatze:
Great idea! Looking forward to renovate not interfering with our work too much anymore. :tada:

metalmatze (approved):

maxbrunet (commented):
Being all over the world, I do not think schedules are the best solution.

We should enable merge queue on `main` instead, it would solve problems beyond Renovate: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#managing-a-merge-queue

metalmatze:
Good point. I wonder if it gets rid of the congestion we sometimes have with concurrent CI runs though. 
How much work would enabling the merge queue be? 

Can't be have both? If so, we might be able to get this in to at least improve it for most Parca maintainers.

maxbrunet:
> How much work would enabling the merge queue be?

That's really just a few clicks, I would like one of you (@manojVivek or @metalmatze) to enable it, so you get familiar with the settings, since you are the ones who are going to benefit from it the most.

manojVivek:
Since we tried and disabled merge queues, can we give scheduling a try? 

Renovate PR merges are holding up a lot of jobs in the queue. 😕

renovate.json
```diff
@@ -1,6 +1,7 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": ["github>parca-dev/.github"],
+  "automergeSchedule": ["after 12am", "before 4am"],
```
maxbrunet, line 4: If doing that, I do not think there is a reason to open PRs if they are not going to be merged

```suggestion
  "schedule": ["after 12am", "before 4am"],
```

manojVivek: Yeah makes sense.

PR #3617, opened 2023-08-14
Add main merge_group to workflows

metalmatze:
Seems like these are needed so the checks are actually run with the merge_groups.


maxbrunet (approved):
Thank you for doing this! :heart:

PR #3624, opened 2023-08-15
pkg/query/flamegraph_arrow: Expose diff values, dont set to null

metalmatze:
So far, the `AppendNull` has been used for the first time a row showed up. Even overwriting (adding new diff values) to the row resulted in the value being returned as null, although the underlying value was there.

## Old 
Notice the blue functions in the first two rows:
image

## New
Notice that the functions in the first two rows now have the proper color:
Screenshot from 2023-08-15 13-12-45


brancz:
Very nice find! This probably fixes the diffing issues that I've been noticing lately with the new arrow based flamegraph.

PR #3632, opened 2023-08-16
pkg/query/flamegraph_arrow: Properly calculate height

metalmatze:
Previously the height was estimated, but didn't take inlined functions and injected label nodes into account.


metalmatze:
I've also updated Go to 1.21 to use the new `max` keyword. :slightly_smiling_face:

brancz (approved):

PR #3635, opened 2023-08-16
pkg/query/flamegraph_arrow: Add tracing

metalmatze:
This will be helpful for long queries.


pkg/query/flamegraph_arrow.go
```diff
@@ -249,10 +256,14 @@ func generateFlamegraphArrowRecord(ctx context.Context, mem memory.Allocator, tr
 		}
 	}
 
+	ctx, spanNewRecord := tracer.Start(ctx, "NewRecord")
```
thorfour, line 259: wouldn't it be cleaner to just defer end the span here instead of before both potential returns?

thorfour (approved):

brancz (approved):

pkg/query/flamegraph_arrow.go
metalmatze: I was thinking about that too. It would work since it's basically the last function call, yes. I felt like being explicit here, might be nicer.

PR #3651, opened 2023-08-19
pkg/query: Improve flamegraph performance through dictionary unification

brancz:
Previously we looked at each individual string to be inserted into the resulting flamegraph, which was very inefficient since strings occur many times over. That causes dictionary building to be inefficient, but also comparisons are done on strings rather than integers.

Changing this to use dictionary unification and transposing yielded a very good performance improvement:

```
$ benchstat old.txt new.txt
ArrowFlamegraph-10  5.95ms ± 0%  3.38ms ± 0%  -43.28%  (p=0.008 n=5+5)
```

### DIFF
--- a/pkg/query/flamegraph_arrow.go
+++ b/pkg/query/flamegraph_arrow.go
@@ -27,7 +27,6 @@ import (
 	"github.com/polarsignals/frostdb/pqarrow/builder"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
-	"golang.org/x/exp/maps"
 
 	queryv1alpha1 "github.com/parca-dev/parca/gen/proto/go/parca/query/v1alpha1"
 	"github.com/parca-dev/parca/pkg/profile"
@@ -117,6 +116,15 @@ func generateFlamegraphArrowRecord(ctx context.Context, mem memory.Allocator, tr
 
 	profileReader := profile.NewReader(p)
 	for _, r := range profileReader.RecordReaders {
+		if err := fb.ensureLabelColumns(r.LabelFields); err != nil {
+			return nil, 0, 0, 0, fmt.Errorf("ensure label columns: %w", err)
+		}
+
+		recordLabelIndex := make(map[string]int, len(r.LabelFields))
+		for i, f := range r.LabelFields {
+			recordLabelIndex[f.Name] = i
+		}
+
 		t, err := fb.newTranspositions(r)
 		if err != nil {
 			return nil, 0, 0, 0, fmt.Errorf("create transpositions: %w", err)
@@ -134,7 +142,6 @@ func generateFlamegraphArrowRecord(ctx context.Context, mem memory.Allocator, tr
 					if sampleLabels == nil {
 						sampleLabels = map[string]string{}
 					}
-
 					labelName := strings.TrimPrefix(r.LabelFields[j].Name, profile.ColumnPprofLabelsPrefix)
 					sampleLabels[labelName] = string(labelColumn.Dict.Value(labelColumn.Col.GetValueIndex(i)))
 				}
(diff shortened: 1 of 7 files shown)


brancz:
The latest commit (removing maps from tracking labels) improves it by another 15%:

```
$ benchstat old.txt new.txt
ArrowFlamegraph-10  3.39ms ± 2%  2.87ms ± 1%  -15.52%  (p=0.008 n=5+5)
```

pkg/query/flamegraph_arrow.go
```diff
@@ -270,9 +281,106 @@ func generateFlamegraphArrowRecord(ctx context.Context, mem memory.Allocator, tr
 	return record, fb.cumulative, fb.maxHeight + 1, 0, nil
 }
 
+type transpositions struct {
+	mappingIDIndicesData *array.Data
```
asubiotto, line 285: nit: I think it would be nicer to unify these `*Data` and `*Indices` fields into a struct to halve the number of fields here and e.g. `Release` calls.

asubiotto: Yes

pkg/query/flamegraph_arrow.go
```diff
@@ -556,168 +685,248 @@ func (fb *flamegraphBuilder) NewRecord() (arrow.Record, error) {
...
+
+	fb.builderLabelsOnly.Append(false)
+
+	// Mapping
+	if r.Mapping.IsValid(locationRow) {
+		fb.builderMappingStart.Append(r.MappingStart.Value(locationRow))
+		fb.builderMappingLimit.Append(r.MappingLimit.Value(locationRow))
+		fb.builderMappingOffset.Append(r.MappingOffset.Value(locationRow))
+		fb.builderMappingFileIndices.Append(t.mappingFileIndices.Value(r.MappingFile.GetValueIndex(locationRow)))
+		fb.builderMappingBuildIDIndices.Append(t.mappingIDIndices.Value(r.MappingBuildID.GetValueIndex(locationRow)))
```
metalmatze, line 844: Oh, very nice!

ui/packages/shared/profile/src/ProfileIcicleGraph/IcicleGraphArrow/utils.ts
```diff
@@ -26,16 +32,20 @@ export function nodeLabel(
   showBinaryName: boolean
 ): string {
   const functionName: string | null = table.getChild(FIELD_FUNCTION_NAME)?.get(row);
+  const labelsOnly: boolean | null = table.getChild(FIELD_LABELS_ONLY)?.get(row);
+  const labels: string | null = table.getChild(FIELD_LABELS)?.get(row);
+  console.log(labelsOnly, labels);
```
metalmatze, line 37: We probably want to remove this before merging.

ui/packages/shared/profile/src/ProfileIcicleGraph/IcicleGraphArrow/utils.ts
```diff
@@ -26,16 +32,20 @@ export function nodeLabel(
...
-    if (level === 1 && functionName.startsWith('{') && functionName.endsWith('}')) {
-      return Object.entries(JSON.parse(functionName))
-        .sort(([a], [b]) => a.localeCompare(b))
-        .map(([k, v]) => `${k}="${v as string}"`)
-        .join(', ');
-    }
     return functionName;
   }
 
+  if (level === 1 && labelsOnly !== null && labelsOnly && labels !== null && labels !== '') {
```
metalmatze, line 42: If we now have this `labelsOnly` column to read from, we can probably remove passing the `level` all the way down.

metalmatze (commented):
You even included the UI changes :tada:

PR #3659, opened 2023-08-21
Move table to using arrow

metalmatze:
This creates an arrow-based table that has full parity with the current table implementation. 
Overall, this is just the start of the table improvements. There are some low-hanging fruits I want to follow up with after this. 

We are now also showing `Flat Diff` and have the `Cumulative Diff` column be named that, rather than just `Diff`. 

image


pkg/query/table.go
```diff
@@ -0,0 +1,392 @@
...
+					var buildID []byte
+					if r.MappingBuildIDDict.IsValid(locationRow) {
+						buildID = r.MappingBuildIDDict.Value(r.MappingBuildID.GetValueIndex(locationRow))
+					}
+					addr := r.Address.Value(locationRow)
+
+					// Check if we've seen the address for the mapping before.
+					// If not, we add it as a new row and add the address to the mapping to keep track of it.
+					// If we have seen the address before, we merge the address with the existing row by summing the values.
+					if cr, ok := tb.addresses[unsafeString(buildID)][addr]; !ok {
```
brancz, line 122: where do we ensure that this doesn't panic?

brancz (approved):
Just a minor question, otherwise lgtm

pkg/query/table.go
metalmatze: Investigating this a bit. Seems like this works out fine. If the first check isn't successful the second one doesn't get checked.
```go
func TestNestedMapPanic(t *testing.T) {
    foo := map[int]map[int]struct{}{}

    if _, ok := foo[0][0]; ok {
        fmt.Println("found")
    } else {
        fmt.Println("not found")
    }
}
```

brancz (approved):

PR #3661, opened 2023-08-22
Improve flamegraph rendering performance

brancz:
With these changes we can render ~800k unique stacks within ~2s. There's still more that can be done, but I felt like this would be a good check in.

All these changes have already been verified to be working as expected by running them on Polar Signals Cloud for some time.


metalmatze (approved):
Super nice on the performance but also some real refactoring goodies in there. Awesome! :tada: :rocket:

PR #3667, opened 2023-08-23
pkg/query/flamegraph_arrow: Fix panic when buildID is empty

metalmatze:
Even if there's a mapping the buildID can still be null.
Found through panics while running with pprof scraping only.


brancz:
How is this possible? 🤔 This doesn't really make any sense. If a mapping exists then worst case the build id might be an empty string, but not null.

metalmatze:
I can try to reproduce some more with my debugger running trying to catch the panic to investigate some more.

brancz:
Could you also paste the exact panic you saw? Judging by this code it should be like I said, this should never be possible to be null if the mapping itself is valid: https://github.com/parca-dev/parca/blob/18567fe5daa44cf7a7f214124e01178a82b1e5a1/pkg/parcacol/arrow.go#L337-L350

PR #3670, opened 2023-08-24
ui: Read source UI improvements

manojVivek:
Fixes:
- [x] Eliminate redundant source requests for the same build id.
- [x] Clicking on a line number should highlight that line and persist it in the URL so that I can send the URL to someone, and it will scroll to that position
- [x] It seems that somehow the hover-bubble keeps re-rendering or something else that makes it feel laggy when having sources open
- [x] Horizontal scrollbar for the source viewer.
- [x] Fallbacks handled when the line number field is missing.


yomete (approved):
lgtm!

brancz (approved):
Amazing work!

PR #3673, opened 2023-08-24
Use checked allocator

thorfour:
Uses a checked allocator throughout unit tests. 

This exposed some allocation leaks that were fixed as well.

### DIFF
--- a/pkg/query/columnquery_test.go
+++ b/pkg/query/columnquery_test.go
@@ -210,8 +210,8 @@ func TestColumnQueryAPIQueryRange(t *testing.T) {
 		require.NoError(t, err)
 	}
 
-	mem := memory.NewCheckedAllocator(memory.DefaultAllocator)
-	defer mem.AssertSize(t, 0)
+	// TODO: this should used a checked allocator but there is a bug in Frostdb that causes it to leak allocations.
+	mem := memory.DefaultAllocator
 	api := NewColumnQueryAPI(
 		logger,
 		tracer,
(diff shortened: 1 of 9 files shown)


pkg/query/columnquery_test.go
```diff
@@ -208,6 +210,8 @@ func TestColumnQueryAPIQueryRange(t *testing.T) {
 		require.NoError(t, err)
 	}
 
+	// TODO: this should used a checked allocator but there is a bug in Frostdb that causes it to leak allocations.
```
brancz, line 213: Do we have an issue that we can link here?

pkg/query/flamegraph_arrow.go
```diff
@@ -942,19 +959,36 @@ func (fb *flamegraphBuilder) NewRecord() (arrow.Record, error) {
 		fb.builderCumulative.NewArray(),
 		fb.builderDiff.NewArray(),
 	}
+	defer func() {
+		for i, arr := range arrays {
+			switch i {
+			case 4, 5, 10, 11, 12:
+				// These arrays have already been defered release above where they were created.
```
brancz, line 966: this seems pretty brittle, maybe let's rather spell them out?

metalmatze (approved):
Nice, definitely needed! :100: 
`defer mem.AssertSize(t, 0)` is pretty cool.

brancz (approved):

pkg/query/flamegraph_arrow.go
thorfour: Moved all array creation to append to a single slice that gets cleaned up at the end.

PR #3683, opened 2023-08-25
ui: Docked Graph MetaInfo Panel

manojVivek:
Screenshots:
Screenshot 2023-08-25 at 1 58 10 PM
Screenshot 2023-08-25 at 1 57 41 PM


manojVivek:
> The code looks good to me, but when looking at the Vercel preview, selecting the option to "Dock Graph Metainfo" doesn't seem to dock the tooltip to the bottom of the screen? Is that the expected behavior?

@monicawoj You need to enable `Arrow Flamegraph` as well. 
Since we're going to phase out the non-arrow flamegraph I made this change only for the arrow version.

ui/packages/shared/profile/src/GraphTooltipArrow/Content.tsx
```diff
@@ -86,31 +79,15 @@ const GraphTooltipArrowContent = ({
...
-
-  const getTextForCumulative = (hoveringNodeCumulative: bigint): string => {
-    const filtered =
-      totalUnfiltered > total
-        ? ` / ${(100 * divide(hoveringNodeCumulative, total)).toFixed(2)}% of filtered`
-        : '';
-    return `${valueFormatter(hoveringNodeCumulative, unit, 2)}
-    (${(100 * divide(hoveringNodeCumulative, totalUnfiltered)).toFixed(2)}%${filtered})`;
-  };
+  const {name, locationAddress, cumulativeText, diffText, diff, row: rowNo} = graphTooltipData;
```
monicawoj, line 82: Super minor comment, but could we rename "rowNo" to "rowNumber". "No" can be interpreted in multiple ways.

monicawoj (approved):
looks fantastic! just one tiny comment

ui/packages/shared/profile/src/GraphTooltipArrow/Content.tsx
manojVivek: Done, addressed it.

PR #3685, opened 2023-08-25
pkg/query: Small arrow flamegraph cleanups

brancz:


metalmatze (approved):
heh, nice, no more label needed :sunglasses:

PR #3719, opened 2023-08-30
pkg/query: Use hash of labels instead of marshaling labels to json

brancz:
Profiling data showed that:

* building the sample label map
* string interactions involved with dealing with keys/values
* map accesses

Were quite expensive (10-15% CPU of flamegraph rendering).

xxhash3 is extremely fast (it's in a way replacing the json being built), and we'll no longer use strings for key/values anywhere, which is much faster to begin with. This does come with the veeeeeery small chance of a hash collision of label-sets. I'd happily see that happen though and then address it.


metalmatze (approved):
Given this is only on the first level of the flame graph the chances for collisions seem indeed _very low_.

PR #3728, opened 2023-08-31
Fix Clear Selection button in table

monicawoj:
With this fix, if the user hits "Clear selection" in the table and something has been previously entered in the "filter by function" input, the highlighting will revert to the filtered result instead of improperly clearing all highlighting (as had been the case before the fix).


manojVivek (approved):
Lgtm

PR #3722, opened 2023-08-30
pkg/query/flamegraph_arrow: Trim using breadth-first-search

metalmatze:
This adds trimming based on breadth-first-search using dictionary compaction. 

It passes the tests locally now. :relieved: :tada: 

Some TODOs are left either to fix within this PR or as follow-ups. 
- [ ] Remove the sorting of label rows from production code (required)
- [ ] Either update Arrow or use a fork for https://github.com/apache/arrow/pull/37459 (required)
- [x] Fix all the memory leaks with the CheckedAllocator (seems like there might have been leaks before)
- [ ] Potentially upstream the compactDictionary at some point later.

PR #3730, opened 2023-08-31
 pkg/query/flamegraph_arrow: Trim using breadth-first-search #3722 

thorfour:
Fork of https://github.com/parca-dev/parca/pull/3722 with memory leak fixes

This adds trimming based on breadth-first-search using dictionary compaction. 

It passes the tests locally now. :relieved: :tada: 

Some TODOs are left either to fix within this PR or as follow-ups. 
- [x] Remove the sorting of label rows from production code (required)
- [x] Either update Arrow or use a fork for https://github.com/apache/arrow/pull/37459 (required)
- [x] Fix all the memory leaks with the CheckedAllocator (seems like there might have been leaks before)
- [ ] Potentially upstream the compactDictionary at some point later.

### DIFF
--- a/pkg/query/flamegraph_arrow.go
+++ b/pkg/query/flamegraph_arrow.go
@@ -17,6 +17,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"sort"
 	"unsafe"
 
 	"github.com/apache/arrow/go/v14/arrow"
@@ -272,6 +273,17 @@ func generateFlamegraphArrowRecord(ctx context.Context, mem memory.Allocator, tr
 	// the last row can also have the most height.
 	fb.maxHeight = max(fb.maxHeight, fb.height)
 
+	// We need to set the first row's cumulative and diff values.
+	// We unify the dictionaries unifiers and indices into actual dictionaries.
+	// These are need for trimming and compaction later on.
+	if err := fb.prepareNewRecord(); err != nil {
+		return nil, 0, 0, 0, fmt.Errorf("failed to prepare the new record: %w", err)
+	}
+
+	if err := fb.trim(ctx, tracer, trimFraction); err != nil {
+		return nil, 0, 0, 0, fmt.Errorf("failed to trim flame graph: %w", err)
+	}
+
 	_, spanNewRecord := tracer.Start(ctx, "NewRecord")
 	defer spanNewRecord.End()
 
@@ -281,7 +293,7 @@ func generateFlamegraphArrowRecord(ctx context.Context, mem memory.Allocator, tr
 	}
 	spanNewRecord.SetAttributes(attribute.Int64("rows", record.NumRows()))
 
-	return record, fb.cumulative, fb.maxHeight + 1, 0, nil
+	return record, fb.cumulative, fb.maxHeight + 1, fb.trimmed, nil
 }
 
 func (fb *flamegraphBuilder) labelsEqual(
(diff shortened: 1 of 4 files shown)


pkg/query/flamegraph_arrow.go
```diff
@@ -832,99 +856,133 @@ func newFlamegraphBuilder(
...
+	// We can combine all of them into the final children root
+	// making our lives easier later on when trimming and then creating the arrow array.
+	fb.children[0] = nil
+
+	// TODO: This shouldn't be here. Either the new hashing is stable or it should be fixed in the tests.
+	rootsRowKeys := make([]uint64, 0, len(fb.rootsRow))
+	for k := range fb.rootsRow {
+		rootsRowKeys = append(rootsRowKeys, k)
+	}
+	sort.Slice(rootsRowKeys, func(i, j int) bool { return rootsRowKeys[i] < rootsRowKeys[j] })
```
metalmatze, line 884: This needs to be removed.
It works without this, but the tests won't be in a stable order anymore.

We need to fix the tests to sort the first level of rows for the group-by-pprof-labels sub test.

thorfour: Great thanks!

pkg/query/flamegraph_arrow_test.go
```diff
@@ -321,65 +320,73 @@ func TestGenerateFlamegraphArrow(t *testing.T) {
...
+			{MappingStart: 0, MappingLimit: 0, MappingOffset: 0, MappingFile: array.NullValueStr, MappingBuildID: array.NullValueStr, LocationAddress: 0, LocationFolded: false, LocationLine: 0, FunctionStartLine: 0, FunctionName: `(null)`, FunctionSystemName: array.NullValueStr, FunctionFilename: array.NullValueStr, Cumulative: 4, Labels: map[string]string{"goroutine": "2"}, Children: []uint32{5}, LabelsOnly: true}, // 2
+			{MappingStart: 0, MappingLimit: 0, MappingOffset: 0, MappingFile: array.NullValueStr, MappingBuildID: array.NullValueStr, LocationAddress: 0, LocationFolded: false, LocationLine: 0, FunctionStartLine: 0, FunctionName: `(null)`, FunctionSystemName: array.NullValueStr, FunctionFilename: array.NullValueStr, Cumulative: 3, Labels: map[string]string{"goroutine": "1"}, Children: []uint32{6}, LabelsOnly: true}, // 3
```
metalmatze, line 374: These 3 lines won't be in stable order anymore. We need to sort them upfront somehow and then use the same order in the levels below.

pkg/query/flamegraph_arrow.go
```diff
@@ -1207,6 +1247,262 @@ func (fb *flamegraphBuilder) addRowValues(r *profile.RecordReader, row, sampleRo
 	fb.builderDiff.Add(row, r.Diff.Value(sampleRow))
 }
 
+func (fb *flamegraphBuilder) trim(ctx context.Context, tracer trace.Tracer, threshold float32) error {
+	_, span := tracer.Start(ctx, "trim")
+	defer span.End()
+
+	releasers := make([]releasable, 0, 3)
```
metalmatze, line 1254: We should be able to calculate the correct amount upfront now. 
```suggestion
	releasers := make([]releasable, 0, 10 + 2 * len(fb.labels))
```
We can follow up with that in another PR. Shouldn't be bad.

pkg/query/flamegraph_arrow_test.go
```diff
@@ -187,8 +187,7 @@ func requireColumnChildren(t *testing.T, record arrow.Record, expected [][]uint3
 func TestGenerateFlamegraphArrow(t *testing.T) {
 	ctx := context.Background()
 	mem := memory.NewCheckedAllocator(memory.DefaultAllocator)
-	defer mem.AssertSize(t, 0)
-	var err error
+	// defer mem.AssertSize(t, 0)
```
metalmatze, line 190: Do we want this disabled here still?

pkg/query/flamegraph_arrow.go
thorfour: Done

pkg/query/flamegraph_arrow_test.go
thorfour: done

PR #3736, opened 2023-09-01
ui: Flamegraph - passing width from the parent instead of detecting from within

manojVivek:
Detecting the width form within the component is creating a flaky situation where the `ref.current` might not be populated based on whether the child component has rendered or not, that was making the width to `undefined` sometimes causing the flame graph not visible. 

So, this fix passed the width from the parent component so that we can be well sure that the ref is already populated.


monicawoj (approved):
looks great!

brancz (approved):

PR #3738, opened 2023-09-01
ui/shared/profile: Pass down arrow response object

metalmatze:
Instead of building the table from the response object right away, pass down the wrapper with additional information, such as trimming value.
Only in the final step, right before building the icicle graph, parse the record bytes into an actual arrow table.

`Trimmed 7 (0.267%) too small values.` shows up with both flame graph report types now.


manojVivek (approved):
Lgtm!

PR #3763, opened 2023-09-05
pkg/query: Select children map before iterating over stack

brancz:


pkg/query/flamegraph_arrow.go
```diff
@@ -113,6 +114,9 @@ func generateFlamegraphArrowRecord(ctx context.Context, mem memory.Allocator, tr
 	profileReader := profile.NewReader(p)
 	labelHasher := xxh3.New()
 	for _, r := range profileReader.RecordReaders {
+		fb.cumulative += math.Int64.Sum(r.Value)
+		fb.diff += math.Int64.Sum(r.Diff)
```
metalmatze, line 118: Sweet. That was a TODO I had left in my head too, since working on the arrow flame graph. :100:

PR #3764, opened 2023-09-05
ui: Make arrow flamegraph the default

metalmatze:
We're confident in the arrow flame graph implementation now.
The same is true for trimming. 

Make the arrow flame graph the default. Always enabled trimming. 

Remove the obsolete user preferences.


brancz (approved):
🥳

yomete (approved):
lgtm!

PR #3771, opened 2023-09-06
Various small arrow flamegraph perf improvements

brancz:
Surprising how much more perf improvements we keep finding:

```
$ benchstat old.txt new.txt
ArrowFlamegraph-10  1.35ms ± 1%  1.15ms ± 1%  -14.25%  (p=0.000 n=10+10)
```

### DIFF
--- a/pkg/query/flamegraph_arrow.go
+++ b/pkg/query/flamegraph_arrow.go
@@ -1521,6 +1521,14 @@ func compactDictionary(mem memory.Allocator, arr *array.Dictionary) (*array.Dict
 	case *array.String:
 		stringBuilder := array.NewStringBuilder(mem)
 		stringBuilder.Reserve(newLen)
+		numBytes := 0
+		for i, count := range keepValues {
+			if count == 0 {
+				continue
+			}
+			numBytes += len(dict.Value(i))
+		}
+		stringBuilder.ReserveData(numBytes)
 		for i, count := range keepValues {
 			if count == 0 {
 				continue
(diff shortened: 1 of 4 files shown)


pkg/query/flamegraph_arrow.go
```diff
@@ -691,6 +698,7 @@ func newFlamegraphBuilder(
 
 		// ensuring that we always have space to set the first row below
 		children:       make([]map[uint64]int, maxInt64(rows, 1)),
+		childrenList:   make([][]int, maxInt64(rows, 1)),
```
metalmatze, line 701:
```suggestion
		childrenList:   make([][]int, max(rows, 1)),
```

pkg/query/flamegraph_arrow.go
```diff
@@ -1196,6 +1214,48 @@ func (fb *flamegraphBuilder) trim(ctx context.Context, tracer trace.Tracer, thre
...
+	trimmedMappingFileIndices.Reserve(row)
+	trimmedMappingBuildIDIndices.Reserve(row)
+	trimmedLocationAddress.Reserve(row)
+	trimmedLocationLine.Reserve(row)
+	trimmedFunctionStartLine.Reserve(row)
+	trimmedFunctionNameIndices.Reserve(row)
+	trimmedFunctionSystemNameIndices.Reserve(row)
+	trimmedFunctionFilenameIndices.Reserve(row)
+	trimmedCumulative.Reserve(row)
+	trimmedDiff.Reserve(row)
```
metalmatze, line 1250: Nice! :rocket:

pkg/query/flamegraph_arrow.go
brancz: We already have a max function defined that expects `a` and `b` to be of type `int`. We can refactor to use the 1.21 built-ins but here I would stick with what we used in the line above.

metalmatze (approved):

PR #3772, opened 2023-09-06
pkg/query: Use optimal index uint type for dictionaries

metalmatze:
Benchmarks have shown no negative CPU impact due to the typecasting. Instead, memory allocation was reduced, as expected.

If the record crosses 16MiB we attach some stats about the record to the tracing span.

Screenshot from 2023-09-06 12-06-39

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -38,7 +38,6 @@ repos:
       rev: master
       hooks:
         - id: go-mod-tidy
-        - id: go-fumpt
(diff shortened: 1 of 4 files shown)


brancz (approved):
I'm expecting profiling is going to show the type casts as hot paths, but I'm happy with waiting and seeing it in data before potentially prematurely optimizing it.

metalmatze:
Yes, I benchmarked the type casts locally, but nothing showed up in CPU time. 
Curious to see the difference in prod and with profiling.

PR #3774, opened 2023-09-06
pkg/query/flamegraph_arrow: Remove mapping start,limit,offset

metalmatze:
Given those are unused in the UI and make up almost 20% of bytes in the arrow record, it's time to finally remove them.

```
name                old time/op    new time/op    delta
ArrowFlamegraph-24    1.88ms ± 6%    1.56ms ±15%  -16.62%  (p=0.000 n=10+10)

name                old alloc/op   new alloc/op   delta
ArrowFlamegraph-24    2.49MB ± 0%    2.00MB ± 1%  -19.84%  (p=0.000 n=10+10)

name                old allocs/op  new allocs/op  delta
ArrowFlamegraph-24     7.31k ± 0%     7.11k ± 1%   -2.72%  (p=0.000 n=10+10)
```


brancz (approved):
Awesome!

PR #3786, opened 2023-09-08
pkg/query: Check individual mapping fields for null

metalmatze:
It seems like mappingStart was valid (not null) but the build ID index was.

### DIFF
--- a/pkg/query/flamegraph_arrow.go
+++ b/pkg/query/flamegraph_arrow.go
@@ -1010,12 +1010,14 @@ func (fb *flamegraphBuilder) appendRow(
 
 	fb.builderLabelsOnly.Append(false)
 
-	// Mapping
-	if r.MappingStart.IsValid(locationRow) {
+	if r.MappingFileIndices.IsValid(locationRow) {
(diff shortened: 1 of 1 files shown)


brancz:
Can we try to understand why this happens? As far as I can tell all codepaths that generate these inputs write a non-null value for all mapping fields if writing a mapping start.

brancz:
As in, this is the code that produces this: https://github.com/parca-dev/parca/blob/00471e8f7986c689f770e2507fede9e0d943323f/pkg/parcacol/arrow.go#L338-L354

metalmatze:
Really odd indeed. Reading around the code you posted I briefly assumed that we might be hitting this part: 
https://github.com/apache/arrow/blob/50015f084846faaa5bf867346d045499e3f68316/go/arrow/array/dictionary.go#L1300-L1303

However, we are calling `Append([]byte(loc.Mapping.BuildId))`. `loc.Mapping` cannot be `nil` as we check that right above and even if `loc.Mapping.BuildId == ""` the dictionary won't appendNull, as `"" != nil`. :thinking:

brancz:
Is there maybe something special about the conversion of an empty string?

PR #3787, opened 2023-09-08
Icicle graph legend

monicawoj:
- Button to toggle visibility of icicle graph color stack legend
- Ellipsis added to legend items in case of overflow. On hover, the full item name is shown
- Align legend items in columns (instead of centered)
- Update URL state to reflect the visibility of the legend (shown or not shown)


yomete (approved):
lgtm!

PR #3795, opened 2023-09-11
pkg/query: Allow filtering by stack prefix

brancz:


gen/proto/swagger/parca/query/v1alpha1/query.swagger.json
```diff
@@ -767,6 +767,10 @@
         "sourceReference": {
           "$ref": "#/definitions/v1alpha1SourceReference",
           "title": "source information about the source requested, required if source report is requested"
+        },
+        "stackPrefix": {
+          "$ref": "#/definitions/v1alpha1StackPrefix",
+          "title": "stack prefix that all samples must have to be considered as a candidiate for the query"
```
javierhonduco, line 773: typo: 
```suggestion
          "title": "stack prefix that all samples must have to be considered as a candidate for the query"
```

javierhonduco (approved):
LGTM! IS this to filter out "interpreter:"? 😏

brancz:
For now this is to decide whether to keep a stack or filter out the whole sample, but the code for the interpreter filter is going to be very similar to this!

proto/parca/query/v1alpha1/query.proto
```diff
@@ -252,6 +252,36 @@ message QueryRequest {
 
   // source information about the source requested, required if source report is requested
   optional SourceReference source_reference = 9;
+
+  // stack prefix that all samples must have to be considered as a candidiate for the query
```
metalmatze, line 256:
```suggestion
  // stack prefix that all samples must have to be considered as a candidate for the query
```

metalmatze (commented):
Super cool! 

I would like to see how usable this already is in the UI. 
It may be worth adding some dynamic parts to the UI (in a separate branch).

brancz:
I was trying to solve lots of things all at once and I don’t think the solution actually solves any of them well. Closing for now.

PR #3805, opened 2023-09-13
pkg/query: Set different metadata on aggregated row to null

metalmatze:
Whenever we merge/aggregate functions we want to make sure that all meta data columns match with their content. 
For example, if the underlying filename is different, we have to set that column's value to `null`. 
This is to avoid showing misleading information.


metalmatze:
FrostDB is updated in this PR and the replace in the go.mod is only to get https://github.com/apache/arrow/pull/37695 in. Once that merges we can simply remove that replace too.

pkg/query/flamegraph_arrow.go
```diff
@@ -528,6 +528,60 @@ func (fb *flamegraphBuilder) mergeSymbolizedRows(
 			fb.intersectLabels(r, t, recordLabelIndex, sampleIndex, cr)
 		}
 
+		// Compare the existing row's metadata values with the one we're merging.
+		// If these values differ we need to set the row's metadata column to null.
+		{
+			a := t.mappingFile.indices.Value(int(r.MappingFileIndices.Value(locationIndex)))
```
brancz, line 534: all of these need to be wrapped in `if fb.builderX.IsValid`, and then an additional internal check that checks if the one being merged is null, as if that's the case it should also be set null. The reason both have to be checked is that setting something null after the fact can have the effect that the equality match might actually pass even if `.SetNull` was called, as `.SetNull` only updates the validity bitmap, not the value array.

metalmatze: Good catch! Updating.

pkg/query/flamegraph_arrow.go
```diff
@@ -528,6 +528,104 @@ func (fb *flamegraphBuilder) mergeSymbolizedRows(
 			fb.intersectLabels(r, t, recordLabelIndex, sampleIndex, cr)
 		}
 
+		// Compare the existing row's metadata values with the one we're merging.
+		// If these values differ we need to set the row's metadata column to null.
+		{
+			if !fb.builderMappingFileIndices.IsNull(cr) {
```
brancz, line 534: I realize this seems drastic, but it has been shown that the negation here is a perf problem, so use `IsValid` rather than `IsNull`, they actually do different validity bitmap checks and therefore we won't loose on perf.

brancz (approved):
One tiny last thing, then lgtm

pkg/query/flamegraph_arrow.go
metalmatze: Right. I can add those later on. However, right now the underlying builders don't have those methods and I instead opted for `!IsNull`. Making this consistent with `IsValid` should be a follow-up.

brancz: Ah got it, ok lgtm then let's move forward

PR #3806, opened 2023-09-13
Modify wording of error messages

yomete:
Removes the `Error!` bit and now only shows the error message itself. Also added implementation of the `onError` callback in the `ui/packages/shared/profile/src/ProfileIcicleGraph/index.tsx` component.


brancz (commented):
I think this is better, but I think we should probably even remove the "An error occurred" and just print what the error says. Also the grpc error should have a status, I think we should render depending on the type of error.

manojVivek (approved):

manojVivek (approved):
lgtm

PR #3818, opened 2023-09-15
Fix sharing profile bug

yomete:
This fixes a bug where clicking on the "Copy Link" button in the profile sharing modal refreshes the page and navigates to the homepage (`/`). The use of the `form` tag seemed to be the issue here.

This also fixes a bug I noticed where hovering on any part of the icicle graph causes the app to crash instantly, with the error seen in the screenshot below. Looks like the check `mappingBuildID === null` wasn't enough, changing it to `!mappingBuildID` worked.
image


monicawoj (approved):
looks great!

PR #3855, opened 2023-09-22
Right click context menu for Icicle Graph

monicawoj:
This PR creates a “right click” context menu for users to more intuitively interact with the Icicle Graph tooltip values/execute other actions as opposed to having to hold shift to lock the tooltip in place.

Designs: https://www.figma.com/file/sMMnQwWpPovvn5JM3vhwL7/Parca?type=design&node-id=540-303451&mode=design&t=sDdXr17zJUh9v7ld-4


brancz:
I like this a lot, one thing that was nice about the previous approach was that I could visually see at least a bit whas I was going to copy, can we add parts of that into the menu? (like the first X characters of the string I'm going to copy)

monicawoj:
> I like this a lot, one thing that was nice about the previous approach was that I could visually see at least a bit whas I was going to copy, can we add parts of that into the menu? (like the first X characters of the string I'm going to copy)

Yes, definitely!

monicawoj:
> Loving where this is going! Curious about your thoughts .. I feel especially with the bottom-docker tooltip, I think it's rather surprising that it disappears when I do a right click. The last teeny tiny nit I have is I would like to be able to see more of the copied value, and maybe format it the way we do for the profile type dropdown? As in "Function Name" as the main test and then slightly grey and smaller underneath the value.

Those both sound good to me. I think in terms of the tooltip and the context menu being open, like you said, if the tooltip is docked there is not really a need to close it on right click. As for the non-docked tooltip, I would close this when the context menu is opened as it essentially contains the same information and occupies a similar location on the screen. 

In terms of the text formatting - I like that suggestion, will incorporate that!

manojVivek (approved):
Lgtm!

Issue #3880, opened 2023-09-26
Add percentage to table


brancz:
While we have the table "visualization" for profiling data, it currently only shows the static value. It would be very useful to understand at glance how much in relation to the whole that is, so show the percentage as well. I think we could just show that inline with the value.

<img width="1829" alt="Screenshot 2023-09-26 at 13 26 32" src="https://github.com/parca-dev/parca/assets/4546722/f877b4b4-8a17-4bd2-a84a-86c4bc318dba">

metalmatze:
This has been implemented.

PR #3780, opened 2023-09-07
ui/profile/table: Add visibility dropdown for columns

metalmatze:
- ui/profile/Table: Add dropdown to change column visibility
- ui/component/Table: Make table header sticky

image

PR #3881, opened 2023-09-26
ui: Add percentage to flat and cumulative columns in table

brancz:
This is a start of #3880, but we should separate it out to separate columns instead as the line breaks are not very good this way.

@metalmatze had some WIP work that allowed dynamically adding columns, I think that would be the right thing to do here.

Screenshot_2023-09-26_at_14 02 32

### DIFF
--- a/ui/packages/shared/profile/src/ProfileView/index.tsx
+++ b/ui/packages/shared/profile/src/ProfileView/index.tsx
@@ -295,6 +295,8 @@ export const ProfileView = ({
       case 'table': {
         return topTableData != null ? (
           <Table
+            total={total}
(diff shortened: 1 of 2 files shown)


metalmatze:
It would be great if we can merge #3780 first.

brancz:
Agreed, like I said in the description I’d want this to use the column selector anyway.

monicawoj:
As the work was rather related, I've gone ahead and added these percentage columns directly in #3780 - please take a look there whenever you get a chance.

brancz:
Superseded by https://github.com/parca-dev/parca/pull/3780

PR #3908, opened 2023-10-12
Metrics Graph - click info panel

monicawoj:
Implements an info panel to inform users of what different types of clicks can do in the Metrics Graph

Design: https://www.figma.com/file/sMMnQwWpPovvn5JM3vhwL7/Parca?type=design&node-id=492-245232&mode=design&t=sNbDSDXIRgg4Om5G-4


yomete (approved):
Looks like we need to target dark mode styling, other than that lgtm!
image

monicawoj:
> Looks like we need to target dark mode styling, other than that lgtm! image

Ah yes, good catch! Will modify that!

PR #3913, opened 2023-10-17
Metrics graph axis labels

### LINKED ISSUE #3894: Add axis labels to the metrics graph
Axis labels would help provide context on what exactly the user is seeing in the metrics graph

For the x-axis the label should be "Time"

For the y-axis, when it's not a delta type:
- the label it's whatever the sample unit is

For the y-axis, when it is a delta type:
- and it's cpu sample count, then the y-axis is actually the "CPU Cores"
- and it's memory space bytes, then the y-axis is bytes per second

monicawoj:
Resolves #3894


brancz (approved):
Flawless 💯

PR #3914, opened 2023-10-17
ui: Context menu to use darkMode state from Parca Context to support downstream integrations

manojVivek:
…


monicawoj (approved):
looks great! 🚀

PR #3918, opened 2023-10-18
.github/workflows: Remove unnecessary files

metalmatze:
Given we explicitly set up Go and Node.js we might be fine deleting of all these things.


brancz (approved):

metalmatze:
Cherry-picked this one into https://github.com/parca-dev/parca/pull/3918 to combine it with updating the `podman` dependency. We need both in the same PR.

PR #3920, opened 2023-10-19
Add profile type to url for shareability

monicawoj:


brancz:
Profile type is already part of the URL as it's part of the query. Is there a specific bug that we're trying to fix here?

monicawoj:
> Profile type is already part of the URL as it's part of the query. Is there a specific bug that we're trying to fix here?

Yes you're right. I can rework this PR so that we don't need a separate key for profile type itself. The problem we're trying to solve here is that although the profile type is in the URL, it is not being recognized on page load (in other words, the profile type selector is not populating with the profile type value from the query)

brancz:
I see! Yes, we should definitely fix that. Do you have a repro for this bug? I can't trigger it. When I select a different profile type and hit search it changes.

monicawoj:
Closing as the issue actually lies in pprof.me, not Parca

PR #3986, opened 2023-10-26
config: Validate duplicate job names

nabokihms:
This commit fixes the problem when jobs with the same names are [compacted to a single target](https://github.com/parca-dev/parca/blob/912596987067593ddb69f7fe12dc6b43eae9d8d7/pkg/scrape/manager.go#L110-L114), leading to unexpected behavior: a user will only see the last target with this job name.

It is a breaking change for users who have duplicates in their configs, but for new users, it will be more transparent.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -56,7 +56,7 @@ type ObjectStorage struct {
 func (c *Config) Validate() error {
 	if err := validation.ValidateStruct(c,
 		validation.Field(&c.ObjectStorage, validation.Required, ObjectStorageValid),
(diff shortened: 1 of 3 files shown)


brancz (approved):
Nice find!

brancz:
Can you try running the benchmarks? They seem to fail consistently on CI.

nabokihms:
@brancz sorry, it was my mistake. I accidentally made the scrape config required.

brancz:
Thank you so much for the contribution!

manojVivek:
@all-contributors Please add @nabokihms for code.

PR #4025, opened 2023-10-31
Create the arrow record directly

thorfour:
Due to historical reasons we were writing the series into a parquet file to sort the rows, to then extract into an arrow record. 

This removes the intermediary sorting step and writes the series directly into arrow records now that FrostDB supports unsorted insertions.

### DIFF
--- a/pkg/parcacol/ingest.go
+++ b/pkg/parcacol/ingest.go
@@ -24,13 +24,10 @@ import (
 
 	"github.com/apache/arrow/go/v14/arrow"
 	"github.com/go-kit/log"
-	"github.com/go-kit/log/level"
 	"github.com/klauspost/compress/gzip"
-	"github.com/parquet-go/parquet-go"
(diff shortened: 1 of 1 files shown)


metalmatze (approved):
Very nice!

asubiotto:
But frostdb still assumes data is sorted in order to reduce compaction memory usage, won't this break that?

thorfour:
> But frostdb still assumes data is sorted in order to reduce compaction memory usage, won't this break that?

Did you change that recently?

asubiotto:
> > But frostdb still assumes data is sorted in order to reduce compaction memory usage, won't this break that?
> 
> Did you change that recently?

Yes, in https://github.com/polarsignals/frostdb/pull/559 and saw significant memory savings. There are other improvements including removing the sorting writer (https://github.com/polarsignals/frostdb/pull/570) that was killing us with labels compaction. More improvements summarized in https://github.com/polarsignals/polarsignals/issues/2850.

PR #4056, opened 2023-11-06
area/ui: Add a `viewComponent` prop in ParcaContext

yomete:
This will allow us to show a `create a view from this query` button similar to the one below. 
image

I also added some `id`s to elements in the `ProfileSelector` as that can be useful for autofocusing into them.


manojVivek (approved):
Lgtm!

PR #4067, opened 2023-11-08
fix y axis ticks for small intervals

### LINKED ISSUE #3964: Y-axis labels sometimes appear strange sometimes
This is likely due to how many decimals are displayed in the D3-created scale

Screenshot 2023-10-24 at 19 03 07

monicawoj:
I believe this resolves #3964 - the problem was likely occuring due to small intervals between y-axis tick marks that were rounding to a single digit.


yomete (approved):
lgtm! I did notice though (after Meticulous reported it) that this PR might have introduced a space before numbers and their units. e.g. `95ns` is now `95 ns`.

or was that intentional?

monicawoj:
> lgtm! I did notice though (after Meticulous reported it) that this PR might have introduced a space before numbers and their units. e.g. `95ns` is now `95 ns`.
> 
> or was that intentional?

Yes, that was intentional. I thought it looked better / is more readable that way. WDYT?

brancz:
After looking at screenshots I think I like it as well. If users end up not liking it we can still change it later.

yomete:
> > lgtm! I did notice though (after Meticulous reported it) that this PR might have introduced a space before numbers and their units. e.g. `95ns` is now `95 ns`.
> > or was that intentional?
> 
> Yes, that was intentional. I thought it looked better / is more readable that way. WDYT?

No qualms here, just wanted to let you know if it wasn't intentional.

PR #4069, opened 2023-11-08
Add additional snap configuration options

morphis:
This adds additional options to be able to configure the `--path-prefix` and `--debuginfod-upstream-servers` arguments of the `parca` binary.


kakkoyun (approved):
LGTM ❤️ 💚 💙 💛 💜

PR #4070, opened 2023-11-08
fix(snap): correct debuginfod upstream servers option

morphis:
This corrects #4069 where I missed to add some late changes to correct the option to specify custom debuginfod servers.


kakkoyun (approved):
Oppsie. Also my bad, I didn't catch them

PR #4073, opened 2023-11-09
ui: Moved compareMode state to ProfileViewContext

manojVivek:


manojVivek:
Merging to use this on downstream, please add any comments.

PR #4138, opened 2023-11-21
ui: Highlight similar stacks in Icicle graph

yomete:
This adds a new preference item  "Highlight similar stacks when hovering over a node" to the preferences modal. This allows for seeing all stacks with the same function name at a glance.

image
image


manojVivek (approved):
Lgtm

monicawoj (approved):
🚀

brancz:
I think the way the user wanted it was a bit different.

yomete:
> I think the way the user wanted it was a bit different. They wanted it to be automatic, perhaps in their preferences, that _whenever_ you hover over a stack it highlights the others, not that you have to find one first, the point is that it helps discover similar frames, it's a bit backward if people first have to find the one they want to highlight.

Thanks for the clarification, must have missed the part about automatic discovery. I'll rework the PR so it behaves like that.

ui/packages/shared/profile/src/ProfileIcicleGraph/IcicleGraphArrow/IcicleGraphNodes.tsx
```diff
@@ -197,6 +213,23 @@ export const IcicleNode = React.memo(function IcicleNodeNoMemo({
...
+  );
+
+  const shouldBeHighlightedIfSimilarStacks = useMemo(() => {
+    return (
+      highlightSimilarStacksPreference &&
+      functionName != null &&
+      functionName === hoveringName &&
+      row !== hoveringRow
+    );
+  }, [highlightSimilarStacksPreference, functionName, row, hoveringName, hoveringRow]);
```
manojVivek, line 230: I guess, the performance impact is because of this: hovering around nodes is going to trigger this memo and cause a re-render on all the nodes in the flamegraph. 

Maybe pre-computing the nodes to highlight might help?

manojVivek: Debouncing in a way that this computation runs only when the mouse rests on a node for a second (or half) could help smoothen it, too.

yomete: thanks for the suggestions, i'll give them a go.

yomete:
Superseded by [4180](https://github.com/parca-dev/parca/pull/4180)

PR #4139, opened 2023-11-21
Time Range picker redesign

### LINKED ISSUE #4024: Make relative time picker linear
We've had user feedback saying that the two-column time picker is confusing as a they read in the order of an academic paper, which is unexpected.

Screenshot 2023-10-31 at 16 18 56

Maybe we should try a single column approach?

monicawoj:
Resolves #4024

Implements a new time range picker design to improve UX of quicker choosing or adjusting time range of query. Additionally, in this PR, a new query is run every time the time range changes instead of having to explicitly click “search”


ui/packages/shared/components/src/DateTimeRangePicker/RelativeDatePicker/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+        .findIndex(totalMinutes => {
+          return totalMinutes === currentTotalMinutes;
+        }) - 1;
+
+    return closestPresetIndex;
+  };
+
+  const currentPresetIndex = getClosestPresetIndex();
+
+  console.log(currentPresetIndex, presetRanges[currentPresetIndex]);
```
yomete, line 120: stray `console.log` here :)

yomete (approved):
lgtm! 🚢🇮🇹

ui/packages/shared/components/src/DateTimeRangePicker/RelativeDatePicker/index.tsx
monicawoj: Will remove that, thanks!

brancz:
Loooooooove this!!

A few minor things:

1) I clicked on the absolute thing while it was 15m, and then on relative again, and it was 6h. That was unexpected.
2) I typed a different value and hit enter, but that didn't do anything, I think hitting enter should run the query.

brancz:
Not sure what's up, but now only the first time hitting enter in the input box works, after that it only triggers when I either use the buttons, or click out of the field. I think hitting enter should already deselect the input field.

PR #4140, opened 2023-11-21
fix timeout checko

### LINKED ISSUE #3436: Problem with Scrape Timeout and Scrape Interval
Hi and sorry for opening another issue, but I think there is sth. wrong with the scrape timeout and scrape interval.

In the [documentation](https://www.parca.dev/docs/ingestion#alternative-approaches) there is an example where the scrape timeout and interval is set:

```yaml
scrape_configs:
  - job_name: "default"
    scrape_interval: "60s"
    scrape_timeout: "45s"
    static_configs:
      - targets: [ '127.0.0.1:7070' ]
```

In the example the scrape interval is larger then the scrape timeout, but while the configuration is loaded there is a check that the timeout is larger then the interval: https://github.com/parca-dev/parca/blob/09c8abda89eaa5dede779b05e458cc23309326ce/pkg/config/config.go#L262-L264

Should this really be the case or is this just a mistake in the documentation. Coming from Prometheus I think the scrape timeout must always be smaller then the scrape interval?

thorfour:
Fixes #3436

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -258,7 +258,7 @@ func (c *ScrapeConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	// Validate the scrape and timeout internal configuration. When /debug/pprof/profile scraping
 	// is enabled we need to make sure there is enough time to complete the scrape.
 	if c.ScrapeTimeout == 0 {
-		c.ScrapeTimeout = c.ScrapeInterval + model.Duration(3*time.Second)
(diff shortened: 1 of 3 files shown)


brancz:
Timeout being larger than the interval is correct.

brancz:
Otherwise, the request would timeout before the next interval starts, and a requests duration is the scrape interval.

thorfour:
> Otherwise, the request would timeout before the next interval starts, and a requests duration is the scrape interval.

Huh that's confusing

PR #4165, opened 2023-11-28
Fix bug when loading Metrics graph

yomete:
Fixes a bug where the ProfileSelector component is trying to set the query expression before it has been successfully fetched from the URL. Adding an early return to the useEffect, and adding `queryExpression` to the dependency array fixes this.


monicawoj (approved):
nice debugging! 💪

PR #4166, opened 2023-11-28
pkg/query: Scale merge profiles by period for diffing merges

metalmatze:
- pkg/parcacol: Aggregate period and duration by timestamp
- pkg/query: Multiply the base and compare profiles for diffs

Something is still slightly off with the multiplication of values and diff values.

### DIFF
--- a/pkg/query/columnquery.go
+++ b/pkg/query/columnquery.go
@@ -819,229 +819,104 @@ func ComputeDiff(ctx context.Context, tracer trace.Tracer, base, compare profile
 	_, span := tracer.Start(ctx, "ComputeDiff")
 	defer span.End()
 
-	// If both durations are > 0 then it's basically comparing delta profiles.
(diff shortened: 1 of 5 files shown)


pkg/query/columnquery.go
```diff
@@ -861,14 +922,140 @@ func ComputeDiff(ctx context.Context, tracer trace.Tracer, base, compare profile
...
+		if valueSum.IsValid(i) && period.IsValid(i) {
+			vs := valueSum.Value(i)
+			vc := valueCount.Value(i)
+			p := period.Value(i)
+
+			normValue := factor * (vs / vc)
+			normPeriod := float64(totalPeriod) / float64(p)
+
+			v := float64(normValue) * normPeriod
+			b.Append(int64(v))
```
metalmatze, line 1038: This is the most important part we should check and discuss. 
I'll add some more tests before merging, which should help too.

Issue #4169, opened 2023-11-29
Separate Normalizer and Ingester Components


Cluas:
In the current system, the Normalizer and Ingester components are tightly coupled, making it difficult to extend and maintain. To improve the system's maintainability and scalability, we need to separate these two components.

The goal of separating the Normalizer and Ingester is to make them independent modules, each with well-defined responsibilities and interfaces. This will allow us to make individual modifications, tests, and deployments more easily.

PR #4170, opened 2023-11-29
refactor: separate Normalizer and Ingester

Cluas:
This PR addresses the problem described in the issue by separating the two components.
#4169


Cluas:
By making this modification, we can use a different storage for the profile store, which will become easy to do. We just need to implement the Ingester ourselves, without the need for coupling with the Normalizer.

pkg/parca/parca.go
```diff
@@ -108,7 +111,8 @@ type Flags struct {
 	Debuginfo  FlagsDebuginfo  `embed:"" prefix:"debuginfo-"`
 	Debuginfod FlagsDebuginfod `embed:"" prefix:"debuginfod-"`
 
-	Metastore string `default:"badger" help:"Which metastore implementation to use" enum:"badger"`
+	Metastore    string `default:"badger" help:"Which metastore implementation to use" enum:"badger"`
+	Profilestore string `default:"frostdb" help:"Which profilestore implementation to use" enum:"frostdb"`
```
brancz, line 115: I would say let's skip a flag until we actually decide on something as part of the Parca project itself.

go.mod
```diff
@@ -59,6 +59,12 @@ require (
 	gopkg.in/yaml.v3 v3.0.1
 )
 
+require (
```
brancz, line 62: let's merge this with the below block

brancz (approved):
Nice! lgtm let's just remove the flag until we actually have other options.

go.mod
Cluas: Done

thorfour (approved):
LGTM!

PR #4173, opened 2023-11-29
fix relative time picker minus click

monicawoj:
Clicking minus too many times (when the value was very small) caused an error to occur. This PR resolves this issue by disabling the minus button in the relative time picker when the current value is below the minimum default values (in other words, below 1m).


yomete (approved):
lgtm!

manojVivek (approved):

PR #4180, opened 2023-11-30
ui: Highlight similar stacks when hovering on a node

yomete:
This adds a new preference item  "Highlight similar stacks when hovering over a node" to the preferences modal. This allows for seeing all stacks with the same function name at a glance.

image
image


manojVivek (approved):
Lgtm!

PR #4245, opened 2024-01-02
ui: Added support for now keyword in time range selector

### LINKED ISSUE #4212: Feature request: support `now` keyword in the time range
I'd like to be able to just put `now` as the end of the time range to see the profiles from a specific moment in time (eg. when a spike started) util `now`. When loading the data, `now` would simply be replaced with the actual current timestamp. This would allow time ranges such as "from 11:05:00 until now".

This would be similar to Grafana, although Grafana allows a lot of sophisticated expressions that I personally almost never use. If there's a way to share a library that evaluates/resolves these expressions into time ranges, I guess this would be nice, although probably much more work:
https://grafana.com/docs/grafana/latest/dashboards/use-dashboards/#semi-relative-time-range

Personally I'd be perfectly happy with just `now` as an "alias" for the current time.

Thanks

manojVivek:
Resolves #4212

### DIFF
--- a/ui/packages/shared/components/src/DateTimePicker/index.tsx
+++ b/ui/packages/shared/components/src/DateTimePicker/index.tsx
@@ -27,6 +27,11 @@ const DateTimePicker = ({selected, onChange}: Props): JSX.Element => (
     showTimeInput
     dateFormat="yyyy-MM-dd HH:mm:ss"
     className="h-[38px] w-full rounded-md border border-gray-200 p-2 text-center text-sm dark:border-gray-600 dark:bg-gray-900"
+    onChangeRaw={event => {
+      if (event.target.value === 'now') {
+        onChange(new Date());
+      }
+    }}
   />
 );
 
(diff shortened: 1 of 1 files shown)


yomete (approved):
lgtm!

brancz:
I don't think the behavior requested in #4212 is quite what was implemented here.

> Personally I'd be perfectly happy with just now as an "alias" for the current time.

I don't think the expected behavior is to literally replace the "now" with the timestamp, but rather keep "now" and transparently use the current timestamp when running the query. We would have to keep both the "meta value" like "now" in the URL as well as the "real" timestamp used to run the last query so we have all the state persisted in the URL.

I do think this also requires us to replace the built-in datetime picker with a custom one where one would be able to pick "now" (and potentially other meta values like "15min ago" in the future but I would skip other things for now).

manojVivek:
> I don't think the expected behavior is to literally replace the "now" with the timestamp, but rather keep "now" and transparently use the current timestamp when running the query. We would have to keep both the "meta value" like "now" in the URL as well as the "real" timestamp used to run the last query so we have all the state persisted in the URL.

Yes, this what Yomi and I finalised on Discord as well, will change the implementation.

manojVivek:
Superseded by #4248

Issue #4241, opened 2023-12-13
Figure the reasoning for the magic numbers used in the metrics range selection and derive it from some global parameter instead of hard-coding it. 


manojVivek:
Reference PR: https://github.com/parca-dev/parca/pull/4240

PR #4246, opened 2024-01-02
ui: Metrics graph highlighted point bug fix

manojVivek:


yomete (approved):
lgtm!

ui/packages/shared/profile/src/MetricsGraph/index.tsx
```diff
@@ -201,11 +201,13 @@ export const RawMetricsGraph = ({
...
-  const getClosest = (): HighlightedSeries | null => {
+  const highlighted = useMemo(() => {
+    // Return the closest point as the highlighted point
+
     const closestPointPerSeries = series.map(function (s) {
       const distances = s.values.map(d => {
-        const x = xScale(d[0]);
-        const y = yScale(d[1]);
+        const x = xScale(d[0]) + margin / 2;
+        const y = yScale(d[1]) - margin / 3;
```
yomete, line 210: just curious here, why did adding a margin and dividing by `2/3` fix the bug?

brancz: curious about this as well, does it maybe have to do with the 2.5 factor in the `xScale` setup?

manojVivek: Yes, these are because of the additional padding that got introduced from the axis labels. 

But still don't have a concrete explanation on why these fractions work. We need to figure that out as part of https://github.com/parca-dev/parca/issues/4241.

PR #4247, opened 2024-01-03
ui: Unify date time format in Metrics graph

yomete:
Changes in this PR:

- Ensures the x-axis date-time format uses the American format (month before the day).
- Make use of the 24H clock across board.
- When zooming in on the metrics graph, show the hour and seconds in the 24H format, as opposed to just the hour only.


manojVivek (approved):
Lgtm!

PR #4248, opened 2024-01-04
ui: Added support for now keyword in time range selector

### LINKED ISSUE #4212: Feature request: support `now` keyword in the time range

manojVivek:
Resolves #4212

Screenshot:
Screenshot 2024-01-04 at 2 02 56 PM


yomete (approved):
lgtm!

PR #4250, opened 2024-01-08
area/ui: Fix datetime picker bug when zooming on graph using mouse drag and drop

yomete:
When zooming in the metrics graph using the drag and drop interaction, everything is updated (graph and icicle) but not the time selector in UI.

image

After some investigation, I discovered the selector is updated after the first zoom but fails to do so for any other subsequent zooms. 

This was due to the `from` and `to` local states in the `ui/packages/shared/components/src/DateTimeRangePicker/AbsoluteDatePicker/index.tsx` file not being updated when the `range` prop changes. Adding a `useEffect` with a dependency of `range` to update the `from` and `to` seems to fix the bug.

_To recreate the original bug, zoom in on the metrics graph using the drag and drop interaction, then do another zoom and you'll notice the date and time isn't being updated._


manojVivek (approved):
Lgtm!

PR #4257, opened 2024-01-11
Fix Dockerfile for parca server

orgads:
If `/data` doesn't exist, `nobody` cannot create it so it fails all the time.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,6 +41,7 @@ LABEL \
     org.opencontainers.image.description="Continuous profiling for analysis of CPU and memory usage, down to the line number and throughout time. Saving infrastructure cost, improving performance, and increasing reliability." \
     org.opencontainers.image.licenses="Apache-2.0"
 
+RUN mkdir /data && chown nobody /data
 USER nobody
 
 COPY --chown=0:0 --from=builder /bin/grpc_health_probe /


metalmatze:
In which scenario does this fail? Please elaborate. Thanks!

orgads:
I'm just running the server, and on each mapping it receives, there's an error.

Server side:
```
level=error ... grpc.error="mark debuginfo upload as uploading via gRPC: write debuginfo metadata to object storage: mkdir /data: permission denied" grpc.time_ms=0.183
```

Client side:
```
level=error ... msg="upload mapping" err="failed to ensure debug information uploaded: initiate upload: rpc error: code = Unknown desc = mark debuginfo upload as uploading via gRPC: write debuginfo metadata to object storage: mkdir /data: permission denied" buildid=9042a6de34394719252c50aed79cdbabdff902a9 filepath=/proc/13207/root/usr/lib/x86_64-linux-gnu/libldap.so.2.0.200
```

metalmatze:
Excellent, thanks for those commands! 
I was able to reproduce it. This should be fixed.

orgads:
The failing test doesn't look related.

orgads:
Rebased

kakkoyun (approved):

PR #4288, opened 2024-01-15
Encode query params when navigation is triggered in useURLState

### LINKED ISSUE #4158: Special characters in search bar get lost when changing UI settings
I typically use the UI to search using regex matchers as in `pod=~"querier.+"`. When this state is present in the search bar and I change sorting or grouping options for selected profiles, the `.+` gets replaced by `. ` and I lose my search results.

yomete:
We currently have a bug where special characters get lost (replaced by an empty space) in the query browser, only after using the `Group` and `Sort` actions. See #4158.

This is happening because (when a navigation is triggered using the `useURLState` hook) the query expression param is not being encoded before being sent back to the URL. This PR fixes that and also for additional cases that were not listed in the original issue i.e. when comparing two profiles and closing one of the compared profiles.

How to reproduce:
Add the following to your `parca.yaml` file to add the label value

```
   labels:
      pod: "querier.+"
```
Using the query browser, query for the label value above (querier.+) and then try to sort or group, and you should get the error below, and notice in the query browser that the plus sign has been replaced by an empty space.

image

Resolves #4158

### DIFF
--- a/ui/packages/shared/components/src/Input/index.tsx
+++ b/ui/packages/shared/components/src/Input/index.tsx
@@ -59,7 +59,7 @@ const Input = ({
       <input
         {...props}
         className={cx(
-          'relative rounded-md border border-gray-200 bg-white p-2 dark:border-gray-600 dark:bg-gray-900 focus:outline-none focus:ring-1 focus:z-10',
+          'relative rounded-md border border-gray-200 bg-white p-2 dark:border-gray-600 dark:bg-gray-900 focus:outline-none focus:ring-1 focus:z-2',
           {
             [className]: className.length > 0,
             '!pr-8': hasAction,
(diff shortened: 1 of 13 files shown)


ui/packages/shared/components/src/Input/index.tsx
```diff
@@ -74,7 +74,7 @@ const Input = ({
       {hasAction ? (
         <div
           className={cx(
-            '!absolute inset-y-[1px] right-[1px] rounded-bl-none rounded-tl-none rounded-md w-8 z-10'
+            '!absolute inset-y-[1px] right-[1px] rounded-bl-none rounded-tl-none rounded-md w-8 z-4'
```
yomete, line 77: This fixes a visual bug where the filter by function action button is superimposed on top of the metrics graph tooltip.

image

brancz (approved):

PR #4293, opened 2024-01-29
area/ui: Replace loading spinners with skeleton placeholders

### LINKED ISSUE #3851: Replace loading spinners with skeleton placeholders
This can be better for UX as users get to see all other parts of the layout like buttons as data is being fetched and provides a better loading experience.

yomete:
This PR replaces the loading spinners with skeleton placeholders for:
- the metrics graph
- the icicle graph
- the top table
- the source view

The skeleton placeholder can be seen in action by using the Vercel preview link.

I also fixed the visual bug where the action buttons for the icicle graph were smushed together when there were multiple visualizations side by side. Fixed this by replacing some of the buttons with texts with icons (only when two visualizations are side by side) which shows a tooltip when you hover on them.
**Before**
image
**After**
image

Resolves #3851


ui/packages/app/web/build/keep.go
yomete, line 1: We've been getting an error when publishing the packages to npm, an example can be seen here https://github.com/parca-dev/parca/actions/runs/7612558919/job/20730449671

image

But after running `yarn build` locally, this file was auto-deleted in the process. I'm not really sure what the file does so I'm leaving this change as is.

metalmatze: Sadly we need this file to successfully use the Parca code in other projects. 
Is there some way we can get lerna to explicitly ignore this particular file? 
Here's the related PR that introduced this file: https://github.com/parca-dev/parca/pull/4105/files

yomete: I see. Thanks for the clarification. I'll look into how we can do that.

manojVivek:
Would it be possible to adjust the height and the padding between the bars to match the actual Flamegraph closely? 

Screenshot 2024-01-30 at 11 37 03 AM 1

manojVivek (approved):
Just a feedback on the overall approach:
Based on my previous experiences with the skeleton loader, it's better if the skeleton loaders are handled by the actual components themselves (by taking in a `loading` prop) instead of creating new components with the matching design.

1. It helps to reduce the jumps in the UI when the app switches between the loaders and the actual content. 
2. It is easier to maintain the loader designs when they are handled by the actually component, otherwise any design change/additions to actual components have to be replicated with the skeleton components as well, becomes a pain in the longer run. 

But this is almost like a rewrite of this whole PR, so I guess we can proceed with this implementation but keep this in mind when we plan any work on the loaders in the future so we can transition towards this during that.

yomete:
> Would it be possible to adjust the height and the padding between the bars to match the actual Flamegraph closely?
> 
> Screenshot 2024-01-30 at 11 37 03 AM 1
> Same with the font of the headers in the top table loader: Screenshot 2024-01-30 at 11 39 27 AM

Thanks for spotting that, I need to do some more responsive work for the skeletons so it's consistent across different screens.

PR #4294, opened 2024-01-29
pkg/query: Return invalid argument if merging non-delta profiles with time range

metalmatze:
This shouldn't be allowed. In fact, our UI will never do such queries, it's just that the backend is still fine with this.


metalmatze:
Some tests fail with `rpc error: code = InvalidArgument desc = merge queries with different timestamps are only supported for delta profiles`. 
We should update these tests. I'll take a look.

metalmatze:
However, we don't want to enforce it at the API level. Closing.

Issue #4295, opened 2024-01-30
`pkg/query`: Data races


metalmatze:
[CI](https://github.com/parca-dev/parca/actions/runs/7612675791/job/20730791979#step:7:429) runs fail to run `make test ENABLE_RACE=yes` due to race conditions. 
Our CI doesn't always catch this, but I can consistently reproduce it on a beefier machine locally. 
`go test -race -tags assert -v ./pkg/query` scopes to test the specific package. 

It seems to already have been on https://github.com/parca-dev/parca/releases/tag/v0.20.0; however, not on https://github.com/parca-dev/parca/releases/tag/v0.19.0. 

`git bisect` said it was this commit that broke things: https://github.com/parca-dev/parca/commit/44bc42e47fabfa319ae39ea19098633c4f362978 
That is quite odd as this commit only ran `buf generate`.

I need more time to investigate when this started happening and then find out why.
If you have an idea, feel free to share.

PR #4296, opened 2024-02-01
Persist `keep.go` file in the web `build` directory after building

yomete:
This PR ensures that the `keep.go` file that was introduced in this [PR](https://github.com/parca-dev/parca/pull/4105/files) is persisted in the `packages/app/web/build/keep.go` after building the packages with Lerna. 

This solves the recurring issue of our npm packages not being published.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -9,6 +9,7 @@
     "type-check": "tsc --noEmit",
     "test": "jest --coverage --config jest.config.cjs",
     "build": "./scripts/run-yarn-build.sh",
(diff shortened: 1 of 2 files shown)


manojVivek (approved):
Lgtm!

metalmatze (approved):
Thanks for taking care of this. 
It's a bit sad this causes so much trouble...

metalmatze:
The Go tests are failing due to the data race #4295. If everything is green we can merge.

PR #4299, opened 2024-02-07
Build arrow records without converting to parquet for sorting

brancz:


pkg/parcacol/sample.go
```diff
@@ -224,100 +225,46 @@ func SeriesToArrowRecord(
...
-
-	records := []arrow.Record{}
-	recordSize := uint(0)
-	for {
-		n, err := rows.ReadRows(rowBuf)
-		if err != nil && err != io.EOF {
-			return nil, err
+	sortingColumns := []arrowutils.SortingColumn{}
+	arrowFields := as.Fields()
+	for _, col := range schema.SortingColumns() {
```
asubiotto, line 230: Is there a way to avoid constructing this sorting schema on every record in the happy case? This might not be worth optimizing at this stage especially since this is already likely a huge improvement.

metalmatze (approved):
Fantastic! :partying_face: I've been looking forward to this day! :tada:

pkg/parcacol/sample.go
```diff
@@ -224,100 +225,46 @@ func SeriesToArrowRecord(
...
-						} else {
-							if err := bldr.Field(i).(*array.BinaryDictionaryBuilder).AppendString(r[i].String()); err != nil {
-								return nil, err
-							}
-						}
-					default:
-						panic(fmt.Sprintf("unknown column %v", as.Field(i).Name))
-					}
+		if colDef.Dynamic {
+			for i, c := range arrowFields {
```
asubiotto, line 242: Kind of annoying to have to iterate over all fields for each dynamic column. What if we created a map with the field prefix so this can instead be an O(1) lookup? Again, not sure if worth optimizing.

asubiotto (commented):
I can't tell you how happy I am to see this

metalmatze:
Anything specific this is blocked on?

brancz:
A benchmark suggested that this was significantly slower than the previous approach.

asubiotto:
Closing in favor of #4740 which builds the record column-by-column and does show an improvement

PR #4304, opened 2024-02-14
*: Display CPU cores and CPU seconds for samples count profiles

brancz:
This patch changes the UX of interacting with CPU profiling data that only reports samples count. It ensures that samples count is multiplied by their period (the time between samples), which allows comparing two profiles that were taken at different frequencies.

As a side effect while changing the queries, this also fixed an issue where the average cpu time was previously summed up over a step period, which was incorrect, as the total cpu time over the step needs to be divided by the total duration over the step.


pkg/parcacol/querier.go
```diff
@@ -471,40 +504,31 @@ func (q *Querier) queryRangeDelta(ctx context.Context, filterExpr logicalplan.Ex
...
-			duration := durationSum / valueCount
-
-			// If we have a CPU samples value type we make sure we always do the next calculation with cpu nanoseconds.
-			// If we already have CPU nanoseconds we don't need to multiply by the period.
-			valuePerSecondSum := valueSum
-			if sampleTypeUnit != "nanoseconds" {
-				valuePerSecondSum = valueSum * period
-			}
-
-			valuePerSecond := float64(valuePerSecondSum) / float64(duration)
```
metalmatze, line 500: 👏

PR #4313, opened 2024-02-20
Improve filter profile data

thorfour:
Mark fields as null instead of rebuilding record during filtering.

```bash
thor@thors-MacBook-Pro ~/.../parca/pkg/query % benchstat before.txt after.txt
name           old time/op    new time/op    delta
FilterData-10    5.00ms ± 2%    0.38ms ± 1%  -92.37%  (p=0.000 n=10+10)

name           old alloc/op   new alloc/op   delta
FilterData-10    6.28MB ± 0%    0.00MB ± 1%  -99.94%  (p=0.000 n=10+10)

name           old allocs/op  new allocs/op  delta
FilterData-10     3.91k ± 1%     0.07k ± 2%  -98.19%  (p=0.000 n=10+10)
```

### DIFF
--- a/pkg/parcacol/arrow.go
+++ b/pkg/parcacol/arrow.go
@@ -129,6 +129,10 @@ func (c *ArrowToProfileConverter) Convert(
 			lOffsetEnd := locationOffsets[i+1]
 			stacktrace := make([]*profile.Location, 0, lOffsetEnd-lOffsetStart)
 			for j := int(lOffsetStart); j < int(lOffsetEnd); j++ {
+				if locations.ListValues().IsNull(j) { // Ignore null locations; they have been filtered out.
+					continue
+				}
+
 				llOffsetStart := lineOffsets[j]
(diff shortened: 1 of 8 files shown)


thorfour:
Updating it to slice the records instead of marking full rows as null

```
name           old time/op    new time/op    delta
FilterData-10    5.00ms ± 2%    0.39ms ± 2%  -92.13%  (p=0.000 n=10+10)

name           old alloc/op   new alloc/op   delta
FilterData-10    6.28MB ± 0%    0.09MB ± 0%  -98.55%  (p=0.000 n=10+9)

name           old allocs/op  new allocs/op  delta
FilterData-10     3.91k ± 1%     0.18k ± 1%  -95.45%  (p=0.000 n=10+9)
```

It's slightly worse, but removes some of the footguns from just marking whole rows as null

metalmatze (approved):
Very nice! 
Improving ergonomics seems the right way forward. Overall this performance so much better, so I'm fine paying that small price. :+1:

Issue #4314, opened 2024-02-20
Profile has no samples


thorfour:
The latest on a self-scraping Parca frequently runs into a `Profile has no samples` error. You typically need only to run for about 5m to reproduce. 

![image](https://github.com/parca-dev/parca/assets/8681572/593b92da-1133-4c31-9959-fc2ddb63cc9f)

git bisect results:

```
thor@thors-MacBook-Pro ~/.../github.com/parca-dev/parca % git bisect bad
  36efece69b215cca8b23e1c1f5890a1df867c1f6 is the first bad commit
  commit 36efece69b215cca8b23e1c1f5890a1df867c1f6
  Author: Frederic Branczyk <fbranczyk@gmail.com>
  Date:   Wed Feb 14 11:17:18 2024 +0100

      *: Display CPU cores and CPU seconds for samples count profiles
```

dreamerlzl:
Using the latest release with md5 4e8627542475a22e564cccb7fcdd8b2b, I still meet the same issue occasionally

![Screenshot 2024-03-05 at 13 53 46](https://github.com/parca-dev/parca/assets/28644777/5f8bd1c5-573f-4b6c-8a75-05ed0dee662e)

brancz:
Is this v0.21.0?

dreamerlzl:
> Is this v0.21.0? The fix is not in that release, it's currently only on `main`.

My bad for not checking the release commits, sorry for bothering.

brancz:
No worries, I was just double checking, I'll make sure to cut a patch release!

PR #4317, opened 2024-02-21
Use sample duration for merge query instead of step

brancz:
Fixes https://github.com/parca-dev/parca/issues/4314


metalmatze (approved):
Makes sense.

manojVivek (approved):

PR #4360, opened 2024-03-06
ui: Importing the cjs version of react-syntax-highlighter

manojVivek:


manojVivek:
Merging to test this, please add any comments.

PR #4476, opened 2024-04-02
ui: Fixed the bug that was resetting the profile selection state on page load

manojVivek:
This `useEffect` was causing the state reset of the selected profile on page load. 

Also, this effect seems unnecessary as it was just causing a new search effect, we could have added this as a workaround to some edge case, but the app works fine without this. So we can re-visit if any problem surfaces.

### DIFF
--- a/ui/packages/shared/profile/src/ProfileSelector/index.tsx
+++ b/ui/packages/shared/profile/src/ProfileSelector/index.tsx
@@ -110,18 +110,6 @@ const ProfileSelector = ({
   const [queryExpressionString, setQueryExpressionString] = useState(querySelection.expression);
 
   useEffect(() => {
-    if (querySelection.expression === undefined) {
-      return;
-    }
-
-    setIsDataLoading(true);
-    setQueryExpression();
-    setIsDataLoading(false);
-
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [timeRangeSelection]);
-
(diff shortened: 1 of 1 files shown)


ui/packages/shared/profile/src/ProfileSelector/index.tsx
```diff
@@ -109,18 +109,6 @@ const ProfileSelector = ({
...
-    if (querySelection.expression === undefined) {
-      return;
-    }
-
-    setIsDataLoading(true);
-    setQueryExpression();
-    setIsDataLoading(false);
-
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [timeRangeSelection]);
```
brancz, line 122: this seems like we had an intention behind this, are we sure we're not breaking something else with this?

manojVivek: I just updated the PR description about this. Basically, this effect feels like a hack as it watches on the `timeRangeSelection` and resets the query state, which is only necessary as any `timeRangeSelection` would do reset the state by itself. 

For now, the app works fine without this on all the common use cases I tried, but we can watch out for any other edge case issue and see how to properly handle that.

yomete (approved):
yeah this lgtm too. I could have sworn that useEffect helped to fix a bug, but like frederic said, we probably solved with some other change.

PR #4525, opened 2024-04-17
Query valuePerSecond for delta profiles

metalmatze:
For delta profiles (cpu with nanoseconds) we query the value per second, which basically is the `sum(value) / sum(duration)` per stack trace.

We already show this value in the metrics graphs. Now, the icicle graph is going to show it too. More importantly, we're going to use this for diffs.

### DIFF
--- a/pkg/parcacol/querier.go
+++ b/pkg/parcacol/querier.go
@@ -403,35 +403,17 @@ func (q *Querier) queryRangeDelta(
 		)
 	}
 
-	var perSecondExpr logicalplan.Expr
-	if isDelta(resultType) {
-		perSecondExpr = logicalplan.Div(
-			logicalplan.Convert(totalSum, arrow.PrimitiveTypes.Float64),
-			logicalplan.Convert(
-				logicalplan.If(
-					logicalplan.IsNull(timestampUnique),
-					logicalplan.Literal(step.Nanoseconds()),
-					durationMin,
-				),
-				arrow.PrimitiveTypes.Float64,
+	perSecondExpr := logicalplan.Div(
+		logicalplan.Convert(totalSum, arrow.PrimitiveTypes.Float64),
+		logicalplan.Convert(
+			logicalplan.If(
+				logicalplan.IsNull(timestampUnique),
+				logicalplan.Literal(step.Nanoseconds()),
+				durationMin,
 			),
-		).Alias(ValuePerSecond)
-	} else {
-		perSecondExpr = logicalplan.Div(
-			logicalplan.Convert(totalSum, arrow.PrimitiveTypes.Float64),
-			logicalplan.Div(
-				logicalplan.Convert(
-					logicalplan.If(
-						logicalplan.IsNull(timestampUnique),
-						logicalplan.Literal(step.Nanoseconds()),
-						durationMin,
-					),
-					arrow.PrimitiveTypes.Float64,
-				),
-				logicalplan.Literal(float64(time.Second.Nanoseconds())),
-			),
-		).Alias(ValuePerSecond)
-	}
+			arrow.PrimitiveTypes.Float64,
+		),
+	).Alias(ValuePerSecond)
 
 	err := q.engine.ScanTable(q.tableName).
 		Filter(filterExpr).
@@ -883,7 +865,7 @@ func (q *Querier) SymbolizeArrowRecord(
 	ctx context.Context,
 	records []arrow.Record,
 	valueColumnName string,
-	meta profile.Meta,
+	queryParts QueryParts,
 ) ([]arrow.Record, error) {
 	res := make([]arrow.Record, len(records))
 
@@ -909,7 +891,7 @@ func (q *Querier) SymbolizeArrowRecord(
 		valueColumn := r.Column(indices[0]).(*array.Int64)
 
 		var valuePerSecondColumn arrow.Array
-		if isDelta(meta.PeriodType) {
+		if queryParts.Delta {
 			indices = schema.FieldIndices(ValuePerSecond)
 			if len(indices) != 1 {
 				return nil, ErrMissingColumn{Column: ValuePerSecond, Columns: len(indices)}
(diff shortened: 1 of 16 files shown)


metalmatze:
I just realized we need to check how these changes will work with pprof.me. I'm not entirely sure right now how we are going to compute the per-second-values there.

pkg/parcacol/querier.go
```diff
@@ -1118,22 +1201,44 @@ func (q *Querier) selectMerge(
...
 		resultType = queryParts.Meta.PeriodType
 	}
 
+	firstProject := append(aggrCols, valueCol)
+	finalProject := append(aggrCols, totalSum)
+	aggrFunctions := []*logicalplan.AggregationFunction{
+		totalSum,
+	}
+
+	if resultType.Type == "cpu" && resultType.Unit == "nanoseconds" {
```
thorfour, line 1217: constants

thorfour (approved):
Few small nits otherwise LGTM!

pkg/parcacol/querier.go
```diff
@@ -897,6 +908,24 @@ func (q *Querier) SymbolizeArrowRecord(
 		}
 		valueColumn := r.Column(indices[0]).(*array.Int64)
 
+		var valuePerSecondColumn arrow.Array
+		if isDelta(meta.PeriodType) {
```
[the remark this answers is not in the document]
metalmatze: Do you mean, that instead of checking it here like this, pass in a `isDelta bool` into the SymbolizeArrow function, like we pass the valueColumnName?

pkg/parcacol/querier.go
```diff
@@ -1008,26 +1055,55 @@ func (q *Querier) findSingle(ctx context.Context, query string, t time.Time) ([]
...
+	var valueCol logicalplan.Expr = logicalplan.Col(profile.ColumnValue)
+
+	firstProject := append(aggrCols, valueCol)
+	finalProject := append(aggrCols, totalSum)
+	aggrFunctions := []*logicalplan.AggregationFunction{
+		logicalplan.Sum(logicalplan.Col(profile.ColumnValue)),
+	}
+
+	if isDelta(queryParts.Meta.PeriodType) {
+		// Only for cpu and nanoseconds do we first project the ColumnDuration.
```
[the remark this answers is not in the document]
metalmatze: I guess, generally speaking, any profile type that has durations is applicable. Do we have any profile types that have duration > 0 with bytes?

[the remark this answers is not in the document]
brancz: yes, allocation profiles that are deltas can be produced from the go runtime if `?seconds=X` is appended

pkg/parcacol/querier.go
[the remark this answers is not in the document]
brancz: I don't know off the top of my head where `SymbolizeArrowRecord` is called, but yes we can better determine whether we're looking at a delta profile using the query instead of the PeriodType.

pkg/parcacol/querier.go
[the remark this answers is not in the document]
metalmatze: Ah, right. Forgot about that.

ui/packages/shared/profile/src/GraphTooltipArrow/Content.tsx
```diff
@@ -61,7 +61,17 @@ const GraphTooltipArrowContent = ({
...
+    name,
+    locationAddress,
+    cumulativeText,
+    cumulativePerSecondText,
+    diffText,
+    diff,
+    row: rowNumber,
+  } = graphTooltipData;
+
+  const delta = unit === 'nanoseconds';
```
brancz, line 74: I think this is the same thing we had in the backend

ui/packages/shared/profile/src/ProfileIcicleGraph/IcicleGraphArrow/index.tsx
```diff
@@ -229,6 +231,7 @@ export const IcicleGraphArrow = memo(function IcicleGraphArrow({
               sortBy={sortBy}
               darkMode={isDarkMode}
               compareMode={compareMode}
+              delta={sampleUnit === 'nanoseconds'}
```
brancz, line 234: ditto

brancz (commented):
apologies didn't see these earlier

metalmatze:
Good point @brancz. 
I've pushed the UI changes to a new branch and PR that we can merge once we land this one: https://github.com/parca-dev/parca/pull/4558
This PR is big enough.

pkg/parcacol/querier.go
```diff
@@ -393,35 +403,17 @@ func (q *Querier) queryRangeDelta(
...
-				logicalplan.If(
-					logicalplan.IsNull(timestampUnique),
-					logicalplan.Literal(step.Nanoseconds()),
-					durationMin,
-				),
-				arrow.PrimitiveTypes.Float64,
-			),
-		).Alias(ValuePerSecond)
-	} else {
-		perSecondExpr = logicalplan.Div(
```
brancz, line 410: why was this branch removed?

pkg/parcacol/querier.go
```diff
@@ -1117,37 +1174,58 @@ func (q *Querier) selectMerge(
 	}
 
 	var valueCol logicalplan.Expr = logicalplan.Col(profile.ColumnValue)
-	if resultType.Type == "samples" && resultType.Unit == "count" {
-		valueCol = logicalplan.Mul(logicalplan.Col(profile.ColumnValue), logicalplan.Col(profile.ColumnPeriod)).Alias(profile.ColumnValue)
-		resultType = queryParts.Meta.PeriodType
```
brancz, line 1122: we needed this

metalmatze: Yes, it's there. Just formatted.

metalmatze: Ah, wait. You're talking about the `resultType = queryParts.Meta.PeriodType` specifically?

pkg/parcacol/querier.go
metalmatze: This entire block of code is executed within the queryRangeDelta function.
It only ever executes when we already checked for a delta profile:
```go
if queryParts.Delta {
	return q.queryRangeDelta(
		ctx,
		filterExpr,
		step,
		queryParts.Meta,
	)
}
```

Looking at this now, I think we need the check though as we can have delta profiles that aren't cpu/nanoseconds. So I'll bring this back.

PR #4531, opened 2024-04-18
pkg/parca: Add forwarder mode

brancz:
This allows the Parca server to be used as a collector within an infrastructure to point all agents or other push-based producers at a single point, to only deal with authentication for example once.

### DIFF
--- a/pkg/parca/parca.go
+++ b/pkg/parca/parca.go
@@ -83,12 +83,13 @@ import (
 const (
 	symbolizationInterval = 10 * time.Second
 	flagModeScraperOnly   = "scraper-only"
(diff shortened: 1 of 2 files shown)


pkg/parca/parca.go
```diff
@@ -83,6 +83,7 @@ import (
 const (
 	symbolizationInterval = 10 * time.Second
 	flagModeScraperOnly   = "scraper-only"
+	flagModeForwarder     = "forwarder"
```
metalmatze, line 86: I think you need to add this to the enum in line 92.

metalmatze (commented):
Very cool idea!

thorfour (approved):

PR #4554, opened 2024-04-25
*: filter profiling data using binary name

yomete:
This PR adds functionality to filter the icicle graph data by a binary.  This was achieved in two parts:

1. by adding a `frameFilter` parameter to the query request which is an array of strings so we can filter by one or many binaries.

2. by implementing a new metadata call to fetch all mapping files (binary) attached to a query as well as other pprof labels. The new metadata call is implemented using a new `REPORT_TYPE`, `REPORT_TYPE_PROFILE_METADATA` on the query request. This was needed because if we filter the icicle graph data using the binary, after filtering we won't know the "total" set of binaries anymore, only those that we filtered by. But with this new call, we can always display all the binaries, irrespective of what it is being currently filtered by.

Additional things included:
- We no longer add runtime functions as a binary.
- The Runtime Filter dropdown (which was essentially a binary filter) has been removed in favor of this new filtering by binary implementation.

image


proto/parca/query/v1alpha1/query.proto
```diff
@@ -257,6 +257,9 @@ message QueryRequest {
 
   // which runtime frames to filter out, often interpreter frames like python or ruby are not super useful by default
   optional RuntimeFilter runtime_filter = 10;
+
+  // which binary to filter by
+  optional string binary_filter = 11;
```
brancz, line 262: this should be a repeated field, and I think we can future proof here a bit, and make this a repeated "filter" and a filter can then be of varying kinds, and the first one would be the binary filter, that way we can then define it to be multiple binaries (which in the Go code we would turn into a set and in the actual filter don't do a byte-match but instead check if the binary is in the set)

pkg/parcacol/querier.go
```diff
@@ -1434,3 +1434,57 @@ func (q *Querier) selectMerge(
...
+		Filter(filterExpr).
+		Project(logicalplan.Col("stacktrace")).
+		Execute(ctx, func(ctx context.Context, r arrow.Record) error {
+			r.Retain()
+
+			locations := r.Column(0).(*array.List)
+
+			values := locations.ListValues().(*array.Dictionary)
+			valueDict := values.Dictionary().(*array.Binary)
+			for i := 0; i < locations.Len(); i++ {
```
brancz, line 1471: thanks to this being dictionary encoded, we can just iterate over the dictionary and already get all unique values

pkg/parcacol/querier.go
```diff
@@ -1434,3 +1434,57 @@ func (q *Querier) selectMerge(
...
+				if locations.IsNull(i) {
+					continue
+				}
+				start, end := locations.ValueOffsets(i)
+				for j := int(start); j < int(end); j++ {
+					encodedLocation := valueDict.Value(values.GetValueIndex(j))
+					symInfo, _ := profile.DecodeSymbolizationInfo(encodedLocation)
+					records = append(records, symInfo.Mapping.File)
```
brancz, line 1479: since we can have duplicates, I'd recommend using a set here instead of a slice

pkg/query/columnquery.go
```diff
@@ -271,13 +272,41 @@ func (q *ColumnQueryAPI) Query(ctx context.Context, req *pb.QueryRequest) (*pb.Q
 		}
 	}()
 
+	if req.GetReportType() == pb.QueryRequest_REPORT_TYPE_PROFILE_METADATA {
+		mappingFiles, err := q.GetMappingFiles(ctx, req.GetMerge())
+
+		labels, labels_err := q.Labels(ctx, &pb.LabelsRequest{
```
brancz, line 278: the labels we want here are not quite the same, we need to replicate what `.Labels` does but for the pprof profile label columns.

pkg/parcacol/querier.go
```diff
@@ -1457,6 +1457,110 @@ func isSamplesCount(st profile.ValueType) bool {
...
+	if err != nil {
+		return nil, err
+	}
+
+	res := make([]string, 0, len(records))
+	for r := range records {
+		res = append(res, r)
+	}
+
+	return res, nil
```
brancz, line 1511: I quite like the idea of sorting the labels, let's do the same with the binaries?

pkg/query/columnquery.go
```diff
@@ -367,8 +378,7 @@ func filterRecord(
 	pool memory.Allocator,
 	rec arrow.Record,
 	filterQueryBytes []byte,
-	binariesToExclude [][]byte,
-	showInterpretedOnly bool,
+	frameFilter []string,
```
brancz, line 381: I'd make this more specific, make this `binaryFrameFilter` and make it a set, we don't want to do the `bytes.Contains` anymore, that was just because we were guessing the binary names, but now we can do exact matches, for which a set-inclusion test is much better.

proto/parca/query/v1alpha1/query.proto
```diff
@@ -256,10 +259,37 @@ message QueryRequest {
...
+
+// StackFilter is a filter for a single or multiple binary names
+message StackFilter {
+  // filter is the list of binary names to filter by
+  repeated string filter = 1;
+}
+
+// FrameFilter is a filter for a single or multiple functions
+message FrameFilter {
+  // filter is the list of function names to filter by
```
brancz, line 279: We need the `FrameFilter` itself to be a `oneof` as well, as we already got the feature request to also be able to filter out certain functions. So we want the first `oneof` of the `FrameFilter` to be a `BinaryFrameFilter` which has a list of strings `include_binaries` (I think we should make the API an allow-list rather than a deny list, otherwise something like "show me only interpreted frames" is going to be a really odd request).

pkg/query/columnquery.go
yomete: can we still do a set inclusion for an exact match since the binary name we'll be getting from the frontend is typically something like `libpython3.11.so.1.0` but the mapping file itself is more like `/usr/lib/libpython3.11.so.1.0` or `kubelet` vs `/home/kubernetes/bin/kubelet`?

proto/parca/query/v1alpha1/query.proto
```diff
@@ -256,10 +259,46 @@ message QueryRequest {
...
+
+    // frame_filter is a filter for filtering by frames
+    FrameFilter frame_filter = 2;
+  }
+}
+
+// StackFilter is a filter for filtering by stacks
+message StackFilter {
+  // filter is the list of stacks to filter by
+  repeated string filter = 1;
```
brancz, line 286: This API should be the same kind of one of, with the only one being FunctionNameStackFilter

ui/packages/shared/profile/src/ProfileIcicleGraph/IcicleGraphArrow/index.tsx
```diff
@@ -174,8 +174,23 @@ export const IcicleGraphArrow = memo(function IcicleGraphArrow({
   };
 
   const hideBinary = (binaryToRemove: string): void => {
-    const newMappingsList = mappingsList.filter(mapping => mapping !== binaryToRemove);
+    // second/subsequent time filtering out a binary i.e. a binary has already been filtered out
+    // and we want to filter out another binary, we need to discover what the first one was
+    // and then add more as needed
+    if (binaryFrameFilter !== undefined && binaryFrameFilter.length > 0) {
+      const binariesToExclude = mappingsList.filter(
```
brancz, line 181: why filter the original list again? `binaryFrameFilter` is already the "current" list of binaries to show if it's non-empty, so just filter that, no?

pkg/compactdictionary/compactdictionary.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/apache/arrow/go/v16/arrow/memory"
+)
+
+type Releasable interface {
+	Release()
+}
+
+// compactDictionary copies only the needed values from the old dictionary to the new dictionary.
+// Once all needed values are copied, it updates the indices referencing those values in their new place.
+func CompactDictionary(mem memory.Allocator, arr *array.Dictionary) (*array.Dictionary, error) {
```
brancz, line 31: no need to move it to its own package, we could already access it within the same package

yomete: Hmm so I did this because Go complains about importing the "github.com/parca-dev/parca/pkg/query" package (the `compactDictionary` function was originally in the flamegraph_arrow.go file which is essentially a `query` package) in the querier.go file with an error of `import cycle not allowed`.

Open to ideas on how to circumvent this.

brancz: Oh my bad, I thought the querier.go was in the `query` package. Ignore.

pkg/compactdictionary/compactdictionary.go
```diff
@@ -41,15 +40,36 @@ func CompactDictionary(mem memory.Allocator, arr *array.Dictionary) (*array.Dict
...
+	case *array.Int32:
+		indices = arr.Indices().(*array.Int32)
+
+		for i := 0; i < indices.(*array.Int32).Len(); i++ {
+			if arr.IsValid(i) {
+				if keepValues[indices.(*array.Int32).Value(i)] == 0 {
+					// keep track of how many values we need to keep to reserve the space upfront
+					newLen++
+				}
+				keepValues[indices.(*array.Int32).Value(i)]++
```
brancz, line 56: you're missing this line for the `Uint32` case

pkg/compactdictionary/compactdictionary.go
```diff
@@ -41,15 +40,36 @@ func CompactDictionary(mem memory.Allocator, arr *array.Dictionary) (*array.Dict
 
 	newLen := 0
 	keepValues := make([]int, arr.Dictionary().Len())
-	for i := 0; i < indices.Len(); i++ {
+
+	var indices interface{}
+
+	switch arr.Indices().(type) {
```
brancz, line 46: you can do something much simpler here:
```
switch arr := arr.Indices().(type) {
```
then `arr` is already typed to the `case` and you no longer need to type assert anything

pkg/compactdictionary/compactdictionary.go
```diff
@@ -111,15 +131,26 @@ func CompactDictionary(mem memory.Allocator, arr *array.Dictionary) (*array.Dict
 	} else {
 		indexBuilder = array.NewUint64Builder(mem)
 	}
-	indexBuilder.Reserve(indices.Len())
+
+	switch arr.Indices().(type) {
```
brancz, line 135: same as the first type switch

pkg/compactdictionary/compactdictionary.go
```diff
@@ -41,15 +40,36 @@ func CompactDictionary(mem memory.Allocator, arr *array.Dictionary) (*array.Dict
 
 	newLen := 0
 	keepValues := make([]int, arr.Dictionary().Len())
-	for i := 0; i < indices.Len(); i++ {
+
+	var indices interface{}
```
brancz, line 44: we don't need this at all, it complicated the code further down as well

pkg/compactdictionary/compactdictionary.go
```diff
@@ -41,34 +40,27 @@ func CompactDictionary(mem memory.Allocator, arr *array.Dictionary) (*array.Dict
...
-
-		for i := 0; i < indices.(*array.Int32).Len(); i++ {
+		for i := 0; i < arr.Len(); i++ {
 			if arr.IsValid(i) {
-				if keepValues[indices.(*array.Int32).Value(i)] == 0 {
+				if keepValues[indices.Value(i)] == 0 {
 					// keep track of how many values we need to keep to reserve the space upfront
 					newLen++
 				}
-				keepValues[indices.(*array.Int32).Value(i)]++
```
brancz: this is very important and is missing in both switch cases

yomete: I have no idea why but the linter/precommit keeps removing this line

pkg/parcacol/querier.go
```diff
@@ -1457,6 +1458,121 @@ func isSamplesCount(st profile.ValueType) bool {
...
+			compactedDict, err := compactDictionary.CompactDictionary(memory.DefaultAllocator, values)
+			if err != nil {
+				fmt.Println("failed to compact dictionary", err)
+				return err
+			}
+			defer compactedDict.Release()
+
+			newValues := compactedDict.Dictionary().(*array.Binary)
+
+			for i := 0; i < valueDict.Len(); i++ {
```
brancz, line 1504: this needs to now use `newValues`

pkg/parcacol/querier.go
```diff
@@ -1457,6 +1458,121 @@ func isSamplesCount(st profile.ValueType) bool {
...
+			}
+
+			for i := 0; i < stringCol.Len(); i++ {
+				// This should usually not happen, but better safe than sorry.
+				if stringCol.IsNull(i) {
+					continue
+				}
+
+				val := stringCol.Value(i)
+				seen[strings.TrimPrefix(val, "labels.")] = struct{}{}
```
brancz, line 1557: this needs to be `pprof_labels`

pkg/parcacol/querier.go
yomete: ughhh thanks for spotting that

pkg/query/columnquery.go
```diff
@@ -905,3 +917,32 @@ func sliceRecord(r arrow.Record, indices []int64) []arrow.Record {
 	slices = append(slices, r.NewSlice(cur.Start, cur.End))
 	return slices
 }
+
+func (q *ColumnQueryAPI) GetMappingFiles(
```
brancz, line 921: no need to export these

proto/parca/query/v1alpha1/query.proto
```diff
@@ -256,10 +259,55 @@ message QueryRequest {
...
 
   // which runtime frames to filter out, often interpreter frames like python or ruby are not super useful by default
-  optional RuntimeFilter runtime_filter = 10;
+  optional RuntimeFilter runtime_filter = 10 [deprecated = true];
 
   // invert_call_stack inverts the call stacks in the flamegraph
   optional bool invert_call_stack = 11;
+
+  // a varying set of filters to apply to the query request
+  Filters filters = 12;
```
brancz, line 268: Oh apologies for missing this earlier, this should be repeated, it should be possible to set multiple filters.

proto/parca/query/v1alpha1/query.proto
```diff
@@ -256,10 +259,55 @@ message QueryRequest {
...
+  oneof filter {
+    // function_name_stack_filter is the list of function names to filter by
+    FunctionNameStackFilter function_name_stack_filter = 1;
+  }
+}
+
+// FunctionNameStackFilter is a filter for filtering by function names
+message FunctionNameStackFilter {
+  // filter is the list of function names to filter by
+  repeated string functions_to_filter = 1;
```
brancz, line 295: this shouldn't be repeated, the top level filter should be repeated (what I commented in the other comment)

proto/parca/query/v1alpha1/query.proto
```diff
@@ -256,10 +259,55 @@ message QueryRequest {
...
 
   // invert_call_stack inverts the call stacks in the flamegraph
   optional bool invert_call_stack = 11;
+
+  // a varying set of filters to apply to the query request
+  Filters filters = 12;
+}
+
+// Filters is a set of filters to apply to the query request
+message Filters {
```
brancz, line 272: This should be singular

brancz (approved):

PR #4558, opened 2024-04-26
ui/packages/shared/profile: Pass profileType into all components

metalmatze:
Rather than only passing the sampleUnit into components, we want to pass the profileTypes into components. It also includes whether a profile is a delta profile.

### DIFF
--- a/ui/packages/shared/profile/src/ProfileView/ProfileViewContext.tsx
+++ b/ui/packages/shared/profile/src/ProfileView/ProfileViewContext.tsx
@@ -17,13 +17,11 @@ import {ProfileSource} from '../ProfileSource';
 
 interface Props {
   profileSource?: ProfileSource;
-  sampleUnit: string;
   compareMode: boolean;
 }
 
 export const defaultValue: Props = {
   profileSource: undefined,
-  sampleUnit: 'bytes',
   compareMode: false,
 };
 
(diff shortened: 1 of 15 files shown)


brancz (approved):

ui/packages/shared/profile/src/ProfileView/ProfileViewContext.tsx
```diff
@@ -17,13 +17,11 @@ import {ProfileSource} from '../ProfileSource';
 
 interface Props {
   profileSource?: ProfileSource;
-  sampleUnit: string;
```
manojVivek, line 20: Any specific reason to not replace this field with `profileType`? 

The reason we introduced this context is to avoid the unnecessary prop drilling for each one of the common data points. `profileType` sounds to fit that use case if I'm not mistaken.

metalmatze: We're passing `profileSource`, which already containers the `profileType` via `profileSource?.ProfileType()`.
If this wasn't the case, I'd agree that we should pass it instead.

manojVivek: Ah, ok didn't realize that.

manojVivek (approved):
Lgtm!

PR #4566, opened 2024-04-30
Pnpm migration

manojVivek:


brancz (approved):
Very nice, I was a bit worried the earley parser was going to cause trouble, but looks like it was pretty smooth. Nice work!

yomete (approved):
nice work!

metalmatze (commented):
amazing! :rocket:

.github/workflows/build-test.yml
```diff
@@ -61,13 +61,17 @@ jobs:
       with:
         go-version-file: .go-version
 
+    - uses: pnpm/action-setup@a3252b78c470c02df07e9d59298aecedc3ccdd6d # v3.0.0
+      with:
+        version: 8.15.6
```
maxbrunet, line 66: Could you set the version from the env and add a comment for Renovate like we do for other tools?

```suggestion
        version: ${{ env.PNPM_VERSION }}
```

```yaml
env:
  # renovate: datasource=npm depName=pnpm versioning=npm
  PNPM_VERSION: 8.15.6
```

maxbrunet (commented):
LGTM 🚀 

just one comment

.github/workflows/build-test.yml
manojVivek: Thanks for catching this, updated other workflow files as well.

PR #4578, opened 2024-05-02
pkg/query: relative comparisons

metalmatze:
When comparing profiles, people usually want to compare relatively, meaning they care about the relative percentages that changed between the profiles, not about the exact times a stack has been seen.
This change ensures that we scale both profiles to the same total cumulative value for the root and then multiply each diff value by the ratio between both roots.


pkg/query/columnquery.go
```diff
@@ -740,16 +747,32 @@ func ComputeDiff(ctx context.Context, tracer trace.Tracer, base, compare profile
...
+				cumulativePerSecondRatio = 1
+			}
+
 			cols := make([]arrow.Array, len(columns))
 			copy(cols, columns)
-			diff := multiplyInt64By(mem, columns[len(columns)-4].(*array.Int64), -1)
+			diff := multiplyInt64By(mem, columns[len(columns)-4].(*array.Int64), -1*cumulativeRatio)
 			defer diff.Release()
-			diffPerSecond := multiplyFloat64By(mem, columns[len(columns)-3].(*array.Float64), -1)
+			diffPerSecond := multiplyFloat64By(mem, columns[len(columns)-3].(*array.Float64), -1*cumulativePerSecondRatio)
```
metalmatze, line 775: Right now, only this side of the comparison is changed. However, if this side has the higher total cumulative value, we need to "scale" it down. Meaning that we sometimes calculate new values with `1 * 0.95` which results in `0` for the int64 columns. 
In that case, we probably should instead "scale" the other side up, right? Basically making it a calculation of `1 * 1.05` which would still be `1` once casted back to `int64`.

pkg/query/columnquery.go
```diff
@@ -732,9 +733,35 @@ func ComputeDiff(ctx context.Context, tracer trace.Tracer, base, compare profile
...
+	totalsBase := make([]totals, len(compare.Samples))
+	for i, s := range base.Samples {
+		totalsBase[i].value = math.Int64.Sum(s.Column(len(s.Columns()) - 4).(*array.Int64))
+		totalsBase[i].valuePerSecond = math.Float64.Sum(s.Column(len(s.Columns()) - 3).(*array.Float64))
+	}
+
 	records := make([]arrow.Record, 0, len(compare.Samples)+len(base.Samples))
+	for i, r := range compare.Samples {
+		var cumulativeRatio, cumulativePerSecondRatio float64
+		if totalsBase[i].value > totalsCompare[i].value {
```
brancz, line 754: I don't understand these. We need to scale up or down based on the ratio, why would we only scale up if the base is larger?

metalmatze: It's all about the percentage between the two profiles. Therefore, I opted to always scale up the side with the lower cumulative total value. It would avoid essentially dividing and instead always multiplying the smaller side. 
During development I've seen quite some profiles (granted it was single profiles) where values were simply scale to `0` because the ratio was something like `0.1` or less.

brancz: So always upscaling makes sense, but there are several problems here:

doing this on a per-record basis is wrong, we need to just sum up all values/per-second-values for all records of base/compare
the else statements don't make sense, just have a single ratio for base and compare that everything is multiplied by (if you want to optimize not multiplying by 1 since it's a no-op then handle that at the place where we perform the multiplication)

The way the code is right now, even if maybe correct, I can't follow it at all.

metalmatze: You're right. I'll refactor to sum up the total for all records of each base and compare.

pkg/query/columnquery.go
```diff
@@ -732,9 +733,35 @@ func ComputeDiff(ctx context.Context, tracer trace.Tracer, base, compare profile
...
+	for _, s := range base.Samples {
+		totalBaseValue += math.Int64.Sum(s.Column(len(s.Columns()) - 4).(*array.Int64))
+		totalBaseValuePerSecond += math.Float64.Sum(s.Column(len(s.Columns()) - 3).(*array.Float64))
+	}
 
+	records := make([]arrow.Record, 0, len(compare.Samples)+len(base.Samples))
 	for _, r := range compare.Samples {
+		var cumulativeRatio, cumulativePerSecondRatio float64
+		if totalBaseValue > totalCompareValue {
+			cumulativeRatio = float64(totalBaseValue) / float64(totalCompareValue)
```
brancz, line 755: can we just move all these calculations before the for loops? then we'll just have a compareRatio, and a baseRatio that we multiply the respective sides by

pkg/query/columnquery.go
```diff
@@ -732,8 +733,41 @@ func ComputeDiff(ctx context.Context, tracer trace.Tracer, base, compare profile
...
+		cumulativeBaseRatio             = 1.0
+		cumulativeBasePerSecondRatio    = 1.0
+	)
+	// We figure out which side needs to be scaled based on each profile's total cumulative value.
+	if totalCompareValue > totalBaseValue {
+		cumulativeBaseRatio = float64(totalCompareValue) / float64(totalBaseValue)
+	} else {
+		cumulativeCompareRatio = float64(totalBaseValue) / float64(totalCompareValue)
+	}
+	if totalCompareValuePerSecond > totalBaseValuePerSecond {
```
brancz, line 764: Having two if statements here doesn't make sense to me, that would mean we can have a situation where we're scaling cumulative based on one side and per second on the other. We should always scale the same side.

brancz (commented):
This largely looks good except for one small comment, but I'm actually more wondering about the UX of this, as we have plenty of evidence that suggests that people want to do both absolute and relative comparisons, depending on the situation, so we should allow users to do both, but with this patch we'd only allow them to do relative comparisons.

metalmatze:
## Absolute 

This shows the absolute comparison between the two profiles. 
It is what we have had all along. It'll be the default for all non-delta profiles (goroutine, memory).
absolute

## Relative 

This shows the relative comparison of the two profiles. The smaller (in terms of overall cumulative value) profile gets scaled up (all of its values are multiplied by the factor the bigger profile is bigger) to have both profiles' total cumulative values be the same. Therefore, compare the percentage between all functions of both profiles. 

relative

yomete (approved):
lgtm!

brancz (approved):

PR #4579, opened 2024-05-02
Add otel profiling support

brancz:
This patch implements the otel-profiling format to be accepted by Parca. Generation is a bit wonky as I needed it to not clash with existing `common` and `resource` types, but that will be gone once the official Go library that contains the generated Go code is updated.

As this is partly meant to inform otel-profiling format feedback, to reduce our own bias, use, and behavior of the format, instead of implementing support for otel-profiling in Parca-Agent first, this was instead primarily tested for compatibility by trying to ingest data from the otel-profiling-agent.

One thing to note about that is that the otel agent always normalizes addresses, therefore this currently hardcodes that mode, ideally this would actually be indicated by the otel profiling format (I'm going to bring this up in today's otel profiling meeting).


ui/packages/app/grafana-datasource-plugin/.prettierrc.js
metalmatze, line 1: All of these files were deleted. `rm -r ui/packages/app/grafana-datasource-plugin/`

PR #4592, opened 2024-05-07
pkg/parcacol: Correctly handle delta non cpu/ns profiles

metalmatze:
Since it's possible to have delta profiles that aren't cpu/nanoseconds we still need to query them correctly.
Additionally, setting the resultType is fixed again too.

### DIFF
--- a/pkg/parcacol/querier.go
+++ b/pkg/parcacol/querier.go
@@ -412,17 +412,35 @@ func (q *Querier) queryRangeDelta(
 		)
 	}
 
-	perSecondExpr := logicalplan.Div(
-		logicalplan.Convert(totalSum, arrow.PrimitiveTypes.Float64),
-		logicalplan.Convert(
-			logicalplan.If(
-				logicalplan.IsNull(timestampUnique),
-				logicalplan.Literal(step.Nanoseconds()),
-				durationMin,
+	var perSecondExpr logicalplan.Expr
+	if isNanoseconds(resultType) {
+		perSecondExpr = logicalplan.Div(
+			logicalplan.Convert(totalSum, arrow.PrimitiveTypes.Float64),
+			logicalplan.Convert(
+				logicalplan.If(
+					logicalplan.IsNull(timestampUnique),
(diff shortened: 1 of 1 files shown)


pkg/parcacol/querier.go
```diff
@@ -393,7 +393,7 @@ func (q *Querier) queryRangeDelta(
 		logicalplan.Col(profile.ColumnDuration),
 	}
 
-	if isSamplesCount(m.SampleType) {
+	if isSamplesCount(resultType) {
```
brancz, line 396: I would still use `m.SampleType`. It's no different, and using `resultType` only happens to be the same at this point in time, as `m.SampleType` is the "default".

pkg/parcacol/querier.go
```diff
@@ -1427,10 +1448,15 @@ func (q *Querier) selectMerge(
 	}
 
 	queryParts.Meta.Timestamp = start
+	queryParts.Meta.SampleType = resultType
```
brancz, line 1451: I'm not a fan of overriding a sub-struct of queryParts like this, we don't use anything else of queryParts

pkg/parcacol/querier.go
```diff
@@ -412,17 +412,35 @@ func (q *Querier) queryRangeDelta(
...
 
-	perSecondExpr := logicalplan.Div(
-		logicalplan.Convert(totalSum, arrow.PrimitiveTypes.Float64),
-		logicalplan.Convert(
-			logicalplan.If(
-				logicalplan.IsNull(timestampUnique),
-				logicalplan.Literal(step.Nanoseconds()),
-				durationMin,
+	var perSecondExpr logicalplan.Expr
+	if isCPUNanoseconds(resultType) {
```
brancz, line 416: just `isNanoseconds` should be sufficient here

pkg/parcacol/querier.go
metalmatze: We return queryParts because those are now passed into SymbolizeArrowRecord to check for delta profiles.

Issue #2818, opened 2023-03-21
pull pprof: allow infrequent profiling


redbaron:
currently there seem to be no way to have infrequent profiling, because  scape interval is added as `?seconds` param, making profiling continuous.

It would be nice if it was possible to specify custom `seconds` option. For a `scrape_interval: 5m` job I tried to create pprof_config with `/debug/pprof/profile?seconds=3s` path , but  it had 2 problems:
- `?` got URLencoded, so it wasn't recognized by pprof server handler
- all path from default  parca `pprof_config` profile config are still present, including `/debug/pprof/profile?seconds=300`

brancz:
For cpu profiling I would recommend to always profile using parca agent, it puts way less overhead on the workload so you can always do it. 

That said I think this use case does make sense. How about we allow specifying a duration in the scrape config and if specified we use that rather than the scrape interval? Eg. here https://github.com/parca-dev/parca/blob/fc8435ec173dbde62ce48cf4a84d9c7b08090421/parca.yaml#L23

redbaron:
> For cpu profiling I would recommend to always profile using parca agent

Thats what I tried first, but it gave me odd symbol names, unlike native pprof. "shape" of the flamegraph was very similar though. Also agent doesn't seem to be offering off-cpu profiling. Having said that, even with pprof continuous profiling enabled CPU usage graph on a 1 minute resolution chart didn't have any noticeable change in my case.

> What do you think?

this and disabling builtin config when `pprof_config` is provided  would do the trick

brancz:
It’s normal that the agent data looks slightly different. It’s actually good, because 1) the runtime hides some things in the “native” pprof profilers, and 2) it uses sigprof which potentially changes program flow. That said, if it looks suspicious we’d love to see it to be able to fix any bugs should there be any.

> this and disabling builtin config when pprof_config is provided would do the trick

Agreed! Do you want to give it a try and implement it?

PR #4623, opened 2024-05-14
pkg/config: support seconds in pprof configuration

alperkokmen:
this updates the definition of `PprofProfilingConfig` such that a new integer field `Seconds` is added which can be set for profiles, in an attempt to resolve https://github.com/parca-dev/parca/issues/2818.

additionally, it changes the behavior where setting any profiling configurations will end up overriding the default so merge will now become a replace.

seconds is accepted as a query parameter and with this change, when provided, it will be used as scraping duration rather than the scraping interval value. this allows having scraping where process being profiled isn't under constant profiling. this only applies to configurations that have delta enabled.

https://pkg.go.dev/net/http/pprof

### DIFF
--- a/pkg/scrape/target.go
+++ b/pkg/scrape/target.go
@@ -388,7 +388,11 @@ func targetsFromGroup(tg *targetgroup.Group, cfg *config.ScrapeConfig, targets [
 				}
 
 				if pcfg, found := cfg.ProfilingConfig.PprofConfig[profType]; found && pcfg.Delta {
-					params.Add("seconds", strconv.Itoa(int(time.Duration(cfg.ScrapeInterval)/time.Second)))
+					seconds := int(time.Duration(cfg.ScrapeInterval) / time.Second)
+					if pcfg.Seconds > 0 {
+						seconds = pcfg.Seconds
+					}
+					params.Add("seconds", strconv.Itoa(seconds))
 				}
 
 				targets = append(targets, NewTarget(lset, origLabels, params, keepSets[i]))
--- a/pkg/scrape/target.go
+++ b/pkg/scrape/target.go
@@ -388,6 +388,7 @@ func targetsFromGroup(tg *targetgroup.Group, cfg *config.ScrapeConfig, targets [
 				}
 
 				if pcfg, found := cfg.ProfilingConfig.PprofConfig[profType]; found && pcfg.Delta {
+					// If Seconds is NOT set on the pprof configuration explicitly, we default to ScrapeInterval.
 					seconds := int(time.Duration(cfg.ScrapeInterval) / time.Second)
 					if pcfg.Seconds > 0 {
 						seconds = pcfg.Seconds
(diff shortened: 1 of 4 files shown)


alperkokmen:
@brancz i have taken a stab at implementing the proposed solution in #2818. please take a look.

alperkokmen:
@brancz, i just wanted to bump this up to get some eyes on it; could you please take a look for add folks who would be able to review? thank you.

brancz:
Apologies for missing this, this looks great!

pkg/scrape/target_test.go
```diff
@@ -0,0 +1,149 @@
+package scrape
```
brancz, line 1: apologies, our linting will insist to add a license header here

pkg/scrape/target.go
```diff
@@ -388,7 +388,11 @@ func targetsFromGroup(tg *targetgroup.Group, cfg *config.ScrapeConfig, targets [
 				}
 
 				if pcfg, found := cfg.ProfilingConfig.PprofConfig[profType]; found && pcfg.Delta {
-					params.Add("seconds", strconv.Itoa(int(time.Duration(cfg.ScrapeInterval)/time.Second)))
+					seconds := int(time.Duration(cfg.ScrapeInterval) / time.Second)
```
brancz, line 391: do you mind just leaving a few code comments here, I had to read it twice to understand it right, just mention that we default to the scrape interval but if the seconds are explicitly set, we use that

brancz:
I'm really sorry, this needs a rebase, but otherwise lgtm! And sorry again for this slipping through the cracks, so thank you for pinging me again. In the future also feel free to ping me on the Parca Discord server if you want to make sure to get a faster review! :)

pkg/scrape/target.go
alperkokmen: no worries – i just included its patch in my rebased branch.

pkg/scrape/target_test.go
alperkokmen: no worries – i just included this in my rebased branch.

alperkokmen:
thanks for taking a look, @brancz. i rebased and addressed the feedback. PTAL when you get a chance.

brancz (approved):

pkg/scrape/target_test.go
```diff
@@ -0,0 +1,163 @@
...
+// limitations under the License.
+
+package scrape
+
+import (
+	"sort"
+	"testing"
+	"time"
+
+	"github.com/parca-dev/parca/pkg/config"
```
brancz, line 21: the linter is complaining that this import needs to be separated from the "external" imports

alperkokmen: fair. let me go over the pre-commit hooks you got here. i should've caught these earlier.

alperkokmen:
@brancz i think i need one last approval before checks are executed?

brancz:
Thank you for the awesome contribution!

PR #4650, opened 2024-05-20
EXPERIMENTAL: Add support for Iceberg storage

thorfour:
This adds a new hidden flag that enables [Apache Iceberg](https://iceberg.apache.org/) storage backend.

This should improve performance of historical queries of Parca.


pkg/parca/parca.go
```diff
@@ -856,3 +887,51 @@ func getDiscoveryConfigs(cfgs []*config.ScrapeConfig) map[string]discovery.Confi
...
+		return "", fmt.Errorf("failed to marshal content of bucket configuration: %w", err)
+	}
+
+	switch strings.ToUpper(string(bucketConf.Type)) {
+	case string(client.GCS):
+		var cfg Config
+		if err := yaml.Unmarshal(config, &cfg); err != nil {
+			return "", err
+		}
+		return filepath.Join("gs://", cfg.Bucket, bucketConf.Prefix), nil
```
brancz, line 912: how does authentication for these work?

brancz (commented):
lgtm

pkg/parca/parca.go
thorfour: It's just using the thanos objstore library. So the same way that authentication currently works in Parca

PR #4665, opened 2024-05-24
pkg/query: Explicitly show Flat and FlatPerSecond values on icicle graph

metalmatze:
It has become clear that it's not always easy for users to understand what exactly consumes resources, aka the flat value, compared to the cumulative value of a function and all of its children. 

This PR surfaces the flat value and flat value in cores/s for delta profiles in the UI for icicle graphs. 

image

With this change, it will be clearer that, for this example, the `scanobject` function used a total of 2.64s, including its children. The function itself, without its children, only used 700ms though (which can be seen on the right-hand side of the icicle graph where nothing is below the function itself).

### DIFF
--- a/pkg/query/flamegraph_arrow.go
+++ b/pkg/query/flamegraph_arrow.go
@@ -694,11 +694,14 @@ func (fb *flamegraphBuilder) mergeUnsymbolizedRows(
 			fb.intersectLabels(r, t, recordLabelIndex, sampleIndex, cr)
 		}
 
-		fb.builderCumulative.Add(cr, r.Value.Value(sampleIndex))
-		fb.builderCumulativePerSecond.Add(cr, r.ValuePerSecond.Value(sampleIndex))
+		value := r.Value.Value(sampleIndex)
+		valuePerSecond := r.ValuePerSecond.Value(sampleIndex)
+
+		fb.builderCumulative.Add(cr, value)
+		fb.builderCumulativePerSecond.Add(cr, valuePerSecond)
 		if leaf {
-			fb.builderFlat.Add(cr, r.Value.Value(sampleIndex))
-			fb.builderFlatPerSecond.Add(cr, r.ValuePerSecond.Value(sampleIndex))
+			fb.builderFlat.Add(cr, value)
+			fb.builderFlatPerSecond.Add(cr, valuePerSecond)
 		}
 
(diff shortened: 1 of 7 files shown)


pkg/query/flamegraph_arrow.go
```diff
@@ -1033,7 +1065,9 @@ func (fb *flamegraphBuilder) prepareNewRecord() error {
 // It adds the children to the children column and the labels intersection to the labels column.
 // Finally, it assembles all columns from the builders into an arrow record.
 func (fb *flamegraphBuilder) NewRecord() (arrow.Record, error) {
-	cleanupArrs := make([]arrow.Array, 0, 18+(2*len(fb.builderLabelFields)))
+	const numCols = 17
```
brancz, line 1068: why not leave 18, but instead make it 17+1? 🤔

pkg/query/flamegraph_arrow.go
```diff
@@ -1786,6 +1870,13 @@ func isLocationRoot(beg, end, i int64, list *array.List) bool {
 	return false
 }
 
+func isLocationLineLeaf(locations, lines *array.List, locationOffset, location, lineOffset, line int64) bool {
+	if locations.ListValues().IsValid(int(location)) && lines.ListValues().IsValid(int(line)) {
+		return location == locationOffset && line == lineOffset
```
brancz, line 1875: this is incorrect, it needs to be the "last valid in the list", it's very possible that the location at `locationOffset` was filtered out, which would mean that nothing will ever be the leaf with the current logic

metalmatze: Actually, we have the same thing in the table and fixed it there. I'll update and use this line from the table: 
https://github.com/parca-dev/parca/blob/7ff555c749a8bac2ef9a13b64923245bc4d4964c/pkg/query/table.go#L158

brancz: ah perfect, yes

pkg/query/flamegraph_arrow.go
metalmatze: Because the 17 is used a couple of times below to initialize `arrays` and `cleanupArrays` . 
To be honest, I'm not totally sure why we initialized this array with 18 compared to all other instances.

PR #4678, opened 2024-05-28
Call View Panel

manojVivek:
Screencast:

https://github.com/parca-dev/parca/assets/1283424/659c3b5a-ed32-4bad-8103-2a930d0459a5


brancz (commented):
I think we should do a few things differently:

1) we should not go further down the path where the function name is the only identifying factor of a line (going further down this path means we won't be able to do `group by`s)
2) we can get the caller callee relationships easier than implemented here, we can just keep the "previous row number" when we iterate through the stack and then at every step adjust the callees of the previous row
3) we should build the relationships directly as row numbers instead of function names

pkg/query/table.go
```diff
@@ -187,8 +192,8 @@ type tableBuilder struct {
 	cumulative int64
 	addresses  map[string]map[uint64]int
 	functions  map[string]int
-	callers    map[string]map[string]bool
-	callees    map[string]map[string]bool
+	callers    map[int]map[int]bool
```
brancz, line 195: no need for `bool` here, just make this a set by using `map[int]struct{}`

pkg/query/table.go
```diff
@@ -262,46 +267,33 @@ func newTableBuilder(mem memory.Allocator) *tableBuilder {
 	return tb
 }
 
-func (tb *tableBuilder) populateCallerAndCalleeData() {
-	reverseFunctionsMap := map[int]string{}
-	for k, v := range tb.functions {
-		reverseFunctionsMap[v] = k
+func getKeys(m map[int]bool) []int64 {
```
brancz, line 270: since Go 1.20 or something there's the `maps` package that has a generic `Keys` function, so yu can just use `maps.Keys(...)`, no need to define these anymore per type

pkg/query/table.go
```diff
@@ -412,81 +404,17 @@ func (tb *tableBuilder) appendRow(
 				tb.builderFlatDiff.Append(0)
 			}
 		case TableFieldCallers:
-			if lineRow != -1 {
-				key := string(r.LineFunctionNameDict.Value(int(r.LineFunctionNameIndices.Value(lineRow))))
-				caller := getCaller(r, sampleRow, locationRow, lineRow)
-				tb.addCaller(key, caller)
-			}
+			tb.addCaller(previousTableRow, currentTableRow)
```
brancz, line 407: love how elegant these are now

pkg/query/table_test.go
```diff
@@ -154,27 +155,39 @@ func TestTableCallView(t *testing.T) {
 		}
 	}
 
+	fmt.Println("nodeIndex unwind failed", nodeIndex)
```
brancz, line 158: debugging leftovers?

pkg/query/table_test.go
```diff
@@ -154,27 +155,39 @@ func TestTableCallView(t *testing.T) {
...
 	callerValues := callersColumn.ListValues().(*array.Int64)
 	calleeValues := calleesColumn.ListValues().(*array.Int64)
 
 	beg, end := callersColumn.ValueOffsets(nodeIndex)
-	require.Equal(t, 0, int(end-beg))
 
-	beg, end = callersColumn.ValueOffsets(child1Index)
-	require.Equal(t, 1, int(end-beg))
-	require.Equal(t, "unwind failed", functionNameColumnDict.Value(functionNameColumn.GetValueIndex(int(callerValues.Value(int(beg))))))
+	fmt.Println("callerValues ", callerValues.IsValid(1))
```
brancz, line 167: dito?

pkg/query/table.go
```diff
@@ -242,13 +271,44 @@ func newTableBuilder(mem memory.Allocator) *tableBuilder {
...
+		builderCallers:            rb.Field(schema.FieldIndices(TableFieldCallers)[0]).(*builder.ListBuilder),
+		builderCallees:            rb.Field(schema.FieldIndices(TableFieldCallees)[0]).(*builder.ListBuilder),
 	}
 
 	return tb
 }
 
+func (tb *tableBuilder) populateCallerAndCalleeData() {
+	for i := range tb.builderFunctionName.Len() {
+		if len(tb.callers) > i {
```
brancz, line 283: I think we need a comment explaining these if statements, if I understand correctly, I think we need it to protect against the case where we didn't expand the callers to the row number (let's say the last row in the flamegraph table doesn't have callers or callees)?

brancz (approved):

pkg/query/table.go
manojVivek: Done, added comments on these conditions.

brancz (approved):

PR #4692, opened 2024-05-30
ui: Refactor the `ColorStackLegend` component to be outside the icicle graph component

yomete:
This ensures that the legend will always be persistent (i.e. the metadata call isn't called again) when clicking through binaries.

### DIFF
--- a/ui/packages/shared/profile/src/ProfileIcicleGraph/IcicleGraphArrow/index.tsx
+++ b/ui/packages/shared/profile/src/ProfileIcicleGraph/IcicleGraphArrow/index.tsx
@@ -13,7 +13,7 @@
 
 import React, {memo, useCallback, useEffect, useMemo, useRef, useState} from 'react';
 
-import {Table, tableFromIPC} from 'apache-arrow';
+import {Dictionary, Table, Vector, tableFromIPC} from 'apache-arrow';
 import {useContextMenu} from 'react-contexify';
 
 import {FlamegraphArrow} from '@parca/client';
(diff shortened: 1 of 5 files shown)


ui/packages/shared/profile/src/ProfileIcicleGraph/IcicleGraphArrow/index.tsx
```diff
@@ -106,50 +126,55 @@ export const IcicleGraphArrow = memo(function IcicleGraphArrow({
 
   const currentSearchString = (selectQueryParam('search_string') as string) ?? '';
   const {compareMode} = useProfileViewContext();
-  const isColorStackLegendEnabled = selectQueryParam('color_stack_legend') === 'true';
+  // const isColorStackLegendEnabled = selectQueryParam('color_stack_legend') === 'true';
```
brancz, line 129: perhaps we're ready to remove this setting?

ui/packages/shared/profile/src/ProfileIcicleGraph/index.tsx
```diff
@@ -354,11 +345,68 @@ const ProfileIcicleGraph = function ProfileIcicleGraphNonMemo({
     return <ErrorContent errorMessage={capitalizeOnlyFirstLetter(error.message)} />;
   }
 
-  if (graph === undefined && arrow === undefined)
-    return <div className="mx-auto text-center">No data...</div>;
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const icicleGraph = useMemo(() => {
+    if (isLoading) {
```
brancz, line 350: we need to also block on `mappingsLoading` here, no?

ui/packages/shared/profile/src/ProfileIcicleGraph/IcicleGraphArrow/index.tsx
yomete: mean't to delete this, must have missed this.

ui/packages/shared/profile/src/ProfileIcicleGraph/index.tsx
yomete: not really, `loading` in the `useEffect` a couple of lines above this code block is already a combination of `flamegraphLoading && profilemetadataLoading` from the parent component's parent component.

```
useEffect(() => {
    if (!loading && (arrow !== undefined || graph !== undefined)) {
      setIsLoading(false);
    } else {
      setIsLoading(true);
    }
  }, [loading, arrow, graph]);
```

brancz:
Nevermind, discussed on discord:

1) loading already checks both https://github.com/parca-dev/parca/blob/8fc2a9919e1afcd7583b08d7757c68c770249bb4/ui/packages/shared/profile/src/ProfileViewWithData.tsx#L192
2) which is set using the combined loading and isloading https://github.com/parca-dev/parca/blob/8fc2a9919e1afcd7583b08d7757c68c770249bb4/ui/packages/shared/profile/src/ProfileIcicleGraph/index.tsx#L330-L336

PR #4712, opened 2024-06-04
ui: Fix the crash with the error `Rendered fewer hooks than expected`.

yomete:
By moving the memoized `icicleGraph` component to be above the early return statement.


manojVivek (approved):
Lgtm!

brancz (approved):

PR #4713, opened 2024-06-04
pkg/query: Only fetch mappings from `profileB` when in compare mode

yomete:


pkg/query/columnquery.go
```diff
@@ -289,22 +289,14 @@ func (q *ColumnQueryAPI) Query(ctx context.Context, req *pb.QueryRequest) (*pb.Q
 			}
 
 		case pb.QueryRequest_MODE_DIFF:
-			mappingFiles_a, labels_a, err := getMappingFilesAndLabels(ctx, q.querier, req.GetDiff().A.GetMerge().GetQuery(), req.GetDiff().A.GetMerge().Start.AsTime(), req.GetDiff().A.GetMerge().End.AsTime())
+			mappingFiles, labels, err := getMappingFilesAndLabels(ctx, q.querier, req.GetDiff().B.GetMerge().GetQuery(), req.GetDiff().B.GetMerge().Start.AsTime(), req.GetDiff().B.GetMerge().End.AsTime())
```
brancz, line 292: this probably deserves a comment why we're only doing `B`

yomete: fair enough, will add one.

brancz (approved):

Issue #4661, opened 2024-05-23
Implement new FunctionFilter API to filter by function


yomete:
As a follow up to #4554, let's ensure to use the new FunctionFilter API to filter by functions in the UI.

PR #4714, opened 2024-06-04
*: Use new Filter API to filter by function

yomete:
This PR uses the `StackFilter` from the Filter API to implement the filter by function functionality. 

I also added a deprecated value to the `filterQuery` field in the proto definition.

Resolves [#4661](https://github.com/parca-dev/parca/issues/4661)


pkg/query/columnquery.go
```diff
@@ -317,6 +317,16 @@ func (q *ColumnQueryAPI) Query(ctx context.Context, req *pb.QueryRequest) (*pb.Q
 		}, nil
 	}
 
+	var functionToFilterBy string
+
+	for _, filter := range req.GetFilter() {
+		if stackFilter := filter.GetStackFilter(); stackFilter != nil {
+			if functionNameFilter := stackFilter.GetFunctionNameStackFilter(); functionNameFilter != nil {
+				functionToFilterBy = functionNameFilter.GetFunctionToFilter()
```
brancz, line 325: this deserves a comment I think, since theoretically the API already allows passing multiple stack filters, but for now we just make the last one actually take effect, which we want to change in the future but not yet

brancz (approved):
Nice work!

pkg/query/columnquery.go
yomete: makes sense, will add the comment.

PR #4720, opened 2024-06-05
ui: Call view Bug fixes

manojVivek:
- [x] Caller/callee label overlap fix
- [x] Prevent selection while double-clicking row
- [x] Fixed time units

### DIFF
--- a/ui/packages/shared/profile/src/GraphTooltipArrow/useGraphTooltip/index.ts
+++ b/ui/packages/shared/profile/src/GraphTooltipArrow/useGraphTooltip/index.ts
@@ -34,6 +34,7 @@ import {
 interface Props {
   table: Table<any>;
   profileType?: ProfileType;
+  unit?: string;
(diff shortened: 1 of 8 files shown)


ui/packages/shared/profile/src/GraphTooltipArrow/useGraphTooltip/index.ts
```diff
@@ -65,6 +67,7 @@ export const useGraphTooltip = ({
   }
 
   const locationAddress: bigint = table.getChild(FIELD_LOCATION_ADDRESS)?.get(row) ?? 0n;
+  unit = unit ?? profileType.periodUnit;
```
brancz, line 70: if unset then `profileType.sampleUnit` would be the correct default

brancz (approved):
tiny comment otherwise lgtm

brancz (approved):

PR #4725, opened 2024-06-06
ui: Make handling of selections more robust

brancz:
This overhauls how we do selections. Since we always perform "merge" selections now this could be significantly simplified and now everything uses the parser as opposed to saving the profile name and labels separately in the URL.


manojVivek (approved):
Lgtm!

yomete (approved):
lgtm!

PR #4776, opened 2024-06-18
ui: Call view bug fixes

manojVivek:
Includes:
- [x] Switched to an improved label position method that works with large caller lists
- [x] Safari absolute position bug fix
- [x] Handled the positioning when the function name is too long
- [x] Preventing double clicks from triggering click events


yomete (approved):
lgtm

PR #4777, opened 2024-06-18
ui: Combine relative and absolute range picker

### LINKED ISSUE #4748: Combine relative and absolute range picker
Just to go back in time to understand why we redesigned the range picker in the first place, the original issue can be seen [here](https://github.com/parca-dev/parca/issues/4024).

`We've had user feedback saying that the two-column time picker is confusing as they read in the order of an academic paper, which is unexpected.`

One thing that came out from the discussion in the issue linked above (and was implemented) is that we wanted to not have the preset ranges as different buttons but instead be able to change them easily using the + and - buttons.

And now to today's problem, the feedback from users is the text below the range picker to switch between the relative and absolute isn't clear enough/easy to miss.

Image

So we're fine with how the relative and absolute range picker works, we just want to improve how to switch between either of them.

If we want to do a combination of both range pickers (relative and absolute), I imagine we can implement it in such a way that:

1. We remove any action buttons (that might increase/decrease the range of the query) so that it simply just displays the range as a text e.g. `Last 30 mins` or `2024-06-11 11:28:23 to 2024-06-11 11:43:23`. 
2. Users can then click on that range text to open up a modal which then shows both the relative and absolute range pickers as they are right now.
3. They can then decide to apply a range using either the relative or absolute picker.
4. The range of the query displayed is determined by what the user last interacted with e.g. on the first load of the profiler, we use the relative range e.g `last 15m` just as it is now, if a user then switches to an absolute range, it is displayed as `2024-06-11 11:28:23 **to** 2024-06-11 11:43:23`.

This may feel like an extra click to change the range, but it's the same no of clicks as clicking the "Use absolute start time instead" text.

yomete:
This PR combines the relative and absolute range pickers to be in a tab switcher so as to make it easier to switch between both range pickers.

Resolves #4748


ui/packages/shared/components/src/DateTimeRangePicker/DateTimeRangePickerText/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+}
+
+const DateTimeRangePickerText = ({
+  range,
+  onClick,
+  isActive,
+  onRangeSelection,
+}: DateTimeRangePickerTextProps): JSX.Element => {
+  const isRelativeRange = range.from.isRelative();
+  const dateString = `${formatDateStringForUI(range.from)} → ${formatDateStringForUI(range.to)}`;
```
manojVivek, line 33: Maybe we can try something like https://github.com/parca-dev/parca/blob/c56e0c29debe2bfaf9373802db4f136aba1f7872/ui/packages/shared/components/src/DateTimeRangePicker/utils.ts#L115-L125 here so that the text is a bit concise.

ui/packages/shared/components/src/DateTimeRangePicker/RelativeDatePicker/index.tsx
```diff
@@ -84,14 +86,110 @@ const parseInput = (input: string): {value: number; unit: string} | null => {
...
+    {value: 3, unit: UNITS.HOUR, text: 'Last 3 hours'},
+    {value: 6, unit: UNITS.HOUR, text: 'Last 6 hours'},
+    {value: 12, unit: UNITS.HOUR, text: 'Last 12 hours'},
+    {value: 1, unit: UNITS.DAY, text: 'Last 1 day'},
+  ];
+
+  return (
+    <div className="flex flex-col gap-3 items-center text-sm p-4">
+      {presetRanges.map(({value, unit, text}) => (
+        <Button
```
manojVivek, line 166: These buttons are rendered as `primary` variants, should we go with `neutral` as default and only set `primary` for the selected range? 

Having all the buttons as `primary` looks a bit too busy. 
Screenshot 2024-06-19 at 10 19 23 AM

yomete: sgtm!

ui/packages/shared/components/src/DateTimeRangePicker/DateTimeRangePickerText/index.tsx
yomete: yeah this is better, thanks for the suggestion

brancz (approved):

manojVivek (approved):
Just a minor comment to fix the `width` changes when switching around the tabs, otherwise lgtm!

PR #4784, opened 2024-06-19
ui: Change the compare button to be within the navbar

yomete:
The dashboard now looks like this:
image

image


yomete:
@manojVivek frederic and myself were discussing that before we start the larger redesign work (based on v0 mockups), one thing we should do first is to move the compare button from the `ProfileSelector` component to be in the navbar.
Screenshot 2024-06-19 at 15 26 54

this is a quick exploration of the implementation logic I had in mind, what do you think of simply making the `compareProfile` function exportable and then using it in the `Navbar` component. And then in the cloud, we can simply import the function again and then do the same thing. Or do you think this is another case for `ParcaContext`?

brancz (approved):

manojVivek (approved):
Lgtm!

PR #4810, opened 2024-06-25
ui: Auto compare profiles

yomete:
This ensures that the Compare button successfully opens the comparison mode for profiles by automatically comparing CPU profiles in the last 15 mins.

### DIFF
--- a/ui/packages/app/web/src/components/ui/Navbar.tsx
+++ b/ui/packages/app/web/src/components/ui/Navbar.tsx
@@ -23,7 +23,7 @@ import {Link, LinkProps, useLocation, useNavigate} from 'react-router-dom';
 import {Button} from '@parca/components';
 import {Parca, ParcaSmall} from '@parca/icons';
 import {selectDarkMode, useAppSelector} from '@parca/store';
-import {convertToQueryParams} from '@parca/utilities';
+import {convertToQueryParams, parseParams} from '@parca/utilities';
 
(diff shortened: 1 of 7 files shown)


yomete:
@manojVivek would appreciate a review on this

metalmatze (approved):
Perfect! 
This now works as expected with and without data.

ui/packages/app/web/src/components/ui/Navbar.tsx
```diff
@@ -164,7 +157,15 @@ const Navbar = () => {
                                   'rounded-none hover:no-underline border-b-2 focus:ring-0 focus:outline-none focus:ring-offset-0 h-full whitespace-nowrap font-medium'
                                 )}
                                 variant="link"
-                                onClick={() => compareProfile(navigateTo)}
+                                onClick={() =>
+                                  navigateTo(
```
manojVivek, line 161: Is it not necessary to include the existing query params as well? 
So that the current state is retained if available, or does it work that already?

manojVivek (approved):
Just a question to confirm a behaviour, looks neat otherwise!

ui/packages/app/web/src/components/ui/Navbar.tsx
yomete: it totally is necessary. I've added that bit now

PR #4814, opened 2024-06-27
area/ui: fix bug in relative range picker

yomete:
This fixes a bug where typing something like `1w` in the relative range picker and then hitting `enter` resets the input to `15m`. This was happening because we were not handling cases for where users are typing in week ranges. This has been added now.


manojVivek (approved):
Lgtm!

PR #4833, opened 2024-07-01
area/ui: Fix how the env var `DEV` is checked

yomete:
This ensures that we can use the Parca JS packages upstream without errors.


manojVivek (approved):
Lgtm!

PR #4834, opened 2024-07-01
Introduce Apache Arrow-based ingestion protocol

brancz:
The protocol works by

1) Sending Apache Arrow records directly.
2) Reducing the amount of traffic that has to be sent by first only sending the sample data, and then if necessary, only send the stacktrace data that is unknown to the server.

This reduces the amount of work that has to be done to ingest data, and reduces the amount of data that needs to be transferred by a client. For a start this implements no caching, that part is left for future work, but the protocol already supports it, at present the server will just always ask the client for the full stacktrace information.


proto/parca/profilestore/v1alpha1/profilestore.proto
```diff
@@ -17,6 +17,36 @@ service ProfileStoreService {
...
+  bytes record = 1;
+}
+
+// WriteResponse may be empty if the server doesn't need any further
+// information, or contain an arrow record that contains the stacktrace IDs
+// that are unknown and therefore requested by the client from the server.
+message WriteResponse {
+  // When record is non-empty it contains the bytes of an arrow record that
+  // contains a column containing the stacktraces that are unknown.
+  bytes record = 1;
```
metalmatze, line 49: It would be nice to make the returning record's schema part of the protocol, at least adding a comment on what to expect roughly.

metalmatze (commented):
Super excited for this!

proto/parca/profilestore/v1alpha1/profilestore.proto
brancz: The record has metadata and the protocol expects the `parca_write_schema_version` to be set to the version, which determines the schema. See the constants in `pkg/normalizer/arrow.go`.

metalmatze (approved):

PR #4847, opened 2024-07-04
ui: Configuring step count as a derivative of screen size such that there are 10px gap per point

manojVivek:


metalmatze (approved):
Nice! 
I'm very curious if this is going to make a difference when using it for longer.

yomete (approved):
lgtm!

PR #4862, opened 2024-07-08
ComputeDiff: Release created arrays

thorfour:
This prevents an underlying memory leak from occurring


metalmatze (approved):
Thank you for fixing this! :100: :pray:

PR #4863, opened 2024-07-08
ui/packages/profile: Always sent the profile type along for Labels and Values

metalmatze:
This greatly reduces the amount of data (row groups) FrostDB has to scan through when running the queries.


manojVivek (approved):
Lgtm!

PR #4864, opened 2024-07-09
ui: URL State hook revamp

manojVivek:
### Why?
- Eliminates the need for `navigateTo` arg to every hook
- Eliminates application logic spill into the hook
- Eliminates redundancy of default value assignment
- Eliminates redux store dependency
- Follows the `useState` like API structure with for generic type support

Includes:
- [x] Replicate existing hook's functionality
- [x] Support default values and in turn prevent URL polluting
- [x] Port non-query selector params
- [x] Remove the old hook and rename the new one to it

Future Work:
- [ ] Port query selector params
- [ ] Support for param config in the context


ui/packages/shared/profile/src/ProfileIcicleGraph/index.tsx
```diff
@@ -156,16 +138,15 @@ const ShowHideLegendButton = ({
...
   });
   const {compareMode} = useProfileViewContext();
 
-  const [storeGroupBy = [FIELD_FUNCTION_NAME], setStoreGroupBy] = useURLState({
-    param: 'group_by',
-    navigateTo,
+  const [groupBy, setStoreGroupBy] = useURLState<string[]>('group_by', {
+    defaultValue: [FIELD_FUNCTION_NAME],
+    alwaysReturnArray: true,
   });
```
yomete, line 150: what's the difference between setting a default value like this i.e. using the `defaultValue` prop,
```suggestion
  const [groupBy, setStoreGroupBy] = useURLState<string[]>('group_by', {
    defaultValue: [FIELD_FUNCTION_NAME],
    alwaysReturnArray: true,
  });
```

and also doing it like the one a bit further in this file, like this:
```suggestion
  const [compareAbsolute = compareAbsoluteDefault, setCompareAbsolute] =
    useURLState('compare_absolute');
```

yomete (approved):
lgtm! just one inquisitive question.

ui/packages/shared/profile/src/ProfileIcicleGraph/index.tsx
manojVivek: No technical difference, just an explicit way to set the default value right now. 

But in the future, I'm thinking we should move away from this distributed (passing it while accessing the hook) default value usage to a more centralized usage by passing the default values in the `URLStateProvider`. So that the consuming hooks don't have to worry about the default state and also when a param is in the default state, we don't actually have to push it to the URL. 
This will also help keep the URL to be a bit concise by only having non-default values.

PR #4878, opened 2024-07-18
ui: Moved sum-by to the profile selector

manojVivek:
Screenshot:

Screenshot 2024-07-18 at 4 21 52 PM


yomete (approved):
lgtm! only need to adjust the background color of the sum by input field in dark mode to match other input fields.

image

manojVivek:
> lgtm! only need to adjust the background color of the sum by input field in dark mode to match other input fields.
> 
> image

Good catch, pushed a commit to fix it.

PR #4895, opened 2024-07-25
area/ui: Parca design refresh

yomete:
<s>**Note: PR #4959 should be looked at, reviewed, and merged in before taking a look at this PR. i.e. #4959 takes precedence over this.**</s>

- [x] Implement new design for navbar
- [x] Implement new design for profile explorer i.ie getting rid of whitespace 
- [x] [Implement a "Simple" and "Advanced" toggle for the query browser](https://github.com/parca-dev/parca/issues/4898)
- [x] [Implement new minimalistic visualization toolbar](https://github.com/parca-dev/parca/issues/4899)
- [x] [Implement settings page](https://github.com/parca-dev/parca/issues/4921)

Follow up:
[Implement new design for compare page](https://github.com/parca-dev/parca/issues/4922)


brancz:
Three small things:

1) I believe the original design had a horizontal line after the metrics graph, I feel like this was nice to separate the two sections.
2) The iciclegraph still has a box around it, if anything, then I think the main visualization should be the thing that uses the most pixels possible.
3) Perhaps this can be integrated in 1, but the original design also had a button for hiding the metrics graph. I feel like it would be nice to have this.

manojVivek:
May not be directly from this PR, but looks like the `min-w` on the sum-by input got lost, making it difficult to see the values easily. 

Screenshot 2024-08-19 at 2 52 04 PM

yomete:
> May not be directly from this PR, but looks like the `min-w` on the sum-by input got lost, making it difficult to see the values easily.
> 
> Screenshot 2024-08-19 at 2 52 04 PM

thanks for pointing out, will take a look at that.

yomete:
> Three small things:
> 
> 1. I believe the original design had a horizontal line after the metrics graph, I feel like this was nice to separate the two sections.
> 2. The iciclegraph still has a box around it, if anything, then I think the main visualization should be the thing that uses the most pixels possible.
> 3. Perhaps this can be integrated in 1, but the original design also had a button for hiding the metrics graph. I feel like it would be nice to have this.

these are all good points. I'll make sure to implement them

brancz (approved):

PR #4931, opened 2024-07-31
area/ui: Add a simple querying mode

### LINKED ISSUE #4898: Implement a "Simple" and "Advanced" toggle for the query browser (Parca design refresh)

yomete:
This PR implements a new "Simple" mode for querying for label names and values in Parca where users will be able to use a select dropdown to choose the label name, operator and label value. The existing freeform querying remains as the "Advanced" mode.

image

To improve the UX of when there are multiple filters, only the first three are shown with the rest behind a "Show more" button.
image

Resolves #4898


manojVivek (approved):
Lgtm!

PR #4937, opened 2024-08-01
pkg/config: Set defaults for existing pprof configs again

### LINKED ISSUE #4883: Kubernetes Pods not visible under targets section
Hello,

To obtain memory profiles from my Go application, I enabled the K8s SD scrape configuration in the `parca.yaml` file using the configuration below. Despite enabling the scrape configs, no pods were listed under the 'kubernetes-pods' section. However, I was still able to see some memory profiles in the UI, even though no targets appeared there.

```
- job_name: kubernetes-pods
    scrape_interval: "60s"
    profiling_config:
        pprof_config:
          process_cpu:
            enabled: false
        process_cpu:
          enabled: false
    kubernetes_sd_configs:
      - role: pod
    relabel_configs:
      - action: labelmap
        regex: __meta_kubernetes_pod_label_app_kubernetes_io_(.+)
        replacement: "app_kubernetes_io_$1"
      - source_labels: [__meta_kubernetes_namespace]
        action: replace
        target_label: namespace
      - source_labels: [__meta_kubernetes_pod_name]
        action: replace
        target_label: pod
      - source_labels: [__meta_kubernetes_pod_container_name]
        action: replace
        target_label: container
 ```

However I started seeing targets listed under the 'kubernetes-pods' section in UI after modifying the scrape configuration to only target pods with a label 'parca.io/scrape: "true"'. There is only one application in this cluster with this label. I suspect that the UI is not listing the targets when the number of targets it need to scrape (pod + url combination) is too large. 

```
 - job_name: kubernetes-pods
      scrape_interval: "60s"
      profiling_config:
        pprof_config:
          process_cpu:
            enabled: false
      kubernetes_sd_configs:
        - role: pod
      relabel_configs:
        - source_labels: [__meta_kubernetes_pod_annotation_parca_io_scrape__]
          regex: true
          action: keep
        - action: labelmap
          regex: __meta_kubernetes_pod_label_app_kubernetes_io_(.+)
          replacement: "app_kubernetes_io_$1"
        - source_labels: [__meta_kubernetes_namespace]
          action: replace
          target_label: namespace
        - source_labels: [__meta_kubernetes_pod_name]
          action: replace
          target_label: pod
        - source_labels: [__meta_kubernetes_pod_container_name]
          action: replace
          target_label: container
```

For more details, please refer to this [thread](https://discord.com/channels/877547706334199818/1262727836847050802).

### PR DESCRIPTION
**Pull request by metalmatze**
Somehow this code was removed. Because of that, if enabled or path aren't specified in the config they simply were left untouched and therefore weren't enabled or with the wrong path, / instead of /debug/pprof/...

What the in-memory config looked like after unmarshalling before this fix:
config-before
What the in-memory config looks like after unmarshalling with this fix:
config-after

closes #4883

### CODE REVIEW

**Review by brancz — APPROVED** [review]

**Comment by brancz** [comment]
@alperkokmen you might want to test whether everything still works as expected after this merges. I suspect it should be fine, but just wanted to call it out since #4623 originally removed these defaults.

**Comment by metalmatze** [comment]
Yes, I did test them locally with our scraper configuration. This is exactly what broke it, and now it's working again. I'll do some more testing once we have a container image.

File: pkg/config/config.go [hunk 1]
```diff
@@ -205,6 +205,16 @@ func (c *ScrapeConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		unmarshalled.ProfilingConfig = defaults.ProfilingConfig
 	} else if unmarshalled.ProfilingConfig.PprofConfig == nil {
 		unmarshalled.ProfilingConfig.PprofConfig = defaults.ProfilingConfig.PprofConfig
+	} else {
+		// Merge unmarshalled config with defaults
+		for pt := range unmarshalled.ProfilingConfig.PprofConfig {
```
metalmatze, line 210: Before this section was removed the code was like this. 

```suggestion
		for pt := range defaults.ProfilingConfig.PprofConfig {
			// nothing set yet so simply use the default
			if unmarshalled.ProfilingConfig.PprofConfig[pt] == nil {
				unmarshalled.ProfilingConfig.PprofConfig[pt] = pc
				continue
			}
```

Maybe we want that again, but that means that we always add all defaults and then rather disable the ones we don't want?! I suppose that was the original behavior, though.

PR #5006, opened 2024-08-22
area/ui: Support typing multiple units in relative range picker

### LINKED ISSUE #4986: area/ui: The relative rangepicker does not support typing in multiple units
The relative rangepicker should support typing in MM:ss/HH:mm e.g. `15m6s`

yomete:
Fixes #4986

### DIFF
--- a/ui/packages/shared/components/src/DateTimeRangePicker/utils.test.ts
+++ b/ui/packages/shared/components/src/DateTimeRangePicker/utils.test.ts
@@ -1,0 +1,107 @@
+// Copyright 2022 The Parca Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {describe, expect, it} from 'vitest';
+
+import {UNITS, formatRange, getRelativeTimeRangeBetweenDates, parseInput} from './utils';
+
+describe('parseInput', () => {
+  it('should parse single unit inputs correctly', () => {
+    expect(parseInput('30s')).toEqual({value: 30, unit: UNITS.SECOND});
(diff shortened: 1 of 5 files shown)


ui/packages/shared/components/src/DateTimeRangePicker/utils.test.ts
```diff
@@ -0,0 +1,107 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {describe, expect, it} from 'vitest';
+
+import {UNITS, formatRange, getRelativeTimeRangeBetweenDates, parseInput} from './utils';
+
+describe('parseInput', () => {
+  it('should parse single unit inputs correctly', () => {
+    expect(parseInput('30s')).toEqual({value: 30, unit: UNITS.SECOND});
```
metalmatze, line 20: What's going to happen for something like 90s vs 1m30s? Are both parsed as 90s? 
```suggestion
    expect(parseInput('90s')).toEqual({value: 90, unit: UNITS.SECOND});
```

metalmatze (commented):
Looks really good! 
I wonder if our life could be easier if this abstraction could simple work on seconds or milliseconds (which is what JS uses)  rather than floating point numbers? Maybe I'm missing some context and what I'm saying doesn't make sense.

ui/packages/shared/components/src/DateTimeRangePicker/utils.test.ts
yomete: Good questions. 

To answer the test question (and maybe the first one) internally, and as a request to the backend they are both processed and parsed as `90s` but in the UI component, it will be parsed as `1m30s`, and this is because the logic favours the higher unit over the lower ones.

Regarding the suggestion to work with seconds or milliseconds instead of floating-point numbers, the floating point numbers are only being used to format the human readable text that shows up in the `RelativeDatepicker` component. This allows us to preserve the user's input format when possible. But as mentioned, we still send the timerange in milliseconds to the backend.

If you think there might be advantages to using milliseconds throughout the entire component, I'm open to discussing potential improvements or alternative implementations.

metalmatze: I don't have a strong opinion on how it gets represented internally. I was thinking that maybe there was an easier way. Seems like it's already pretty good, so please ignore my suggestion.

metalmatze: Feel free to merge this as is too.

PR #5007, opened 2024-08-22
pkg/query: Group by arbitrary labels

metalmatze:
Previously, Parca allowed to group by all `pprof_labels` _at once_ that are part of profiles. 
With these changes, Parca allows grouping by `labels`, which now includes the `pprof_labels` and the infrastructure `labels`.
This allows grouping by `comm` or `namespace` within an icicle graph. 
Additionally, grouping by only the `trace_id` but not any other labels like `thread_id` will be possible. 

## A simple example

Filter for `runtime.gcBgMarkWorker` in all of your infrastructure. Then group the icicle graph by `job` to see the garbage collection by `job`. 

### Before
before

### After
after

Notice the new rows in the second screenshot now with `job=prometheus` and `job=parca`. Beneath them, we pretty much have the same stack traces that were previously aggregated together. 

## UI 

Selecting the labels to group by looks like this now: 

We need to think about the Select component within the dropdown... Not ideal right now.


pkg/parca/parca_test.go
```diff
@@ -171,6 +171,8 @@ func MustReadAllGzip(t require.TestingT, filename string) []byte {
 }
 
 func TestConsistency(t *testing.T) {
+	t.Skipf("skipped, need to think how we want to bring back consistency exports without pprof_labels")
```
metalmatze, line 174: Given we don't store the pprof_labels anymore, it's impossible to export the same profile as pprof download from now on.

Issue #5106, opened 2024-09-17
area/ui: More actions menu is too close to the Icicle graph border


yomete:
![image](https://github.com/user-attachments/assets/f59e8ea6-9a11-4a4d-bbc2-3d3a32dcc354)

PR #5114, opened 2024-09-19
area/ui: Add a Color by functionality

yomete:
This PR introduces a color by feature such that users can now choose how they want their icicle graph to be colored e.g. binaries and filenames.

- Added a new "Color by" item to the "More Action" button to allow users color by either binary or filename.
image
- Also added a "Hidden Binaries" item so that if you filter using a binary and then switch the color by to be "Filename", you still get to see what binary you might have been filtering with previously.
image
- When there's just one binary, we automatically color by filename.

Also contains a fix for #5106


manojVivek (approved):
Looks good in terms of the code, but I'll defer to Frederic/Matthias on the functionality.

PR #5179, opened 2024-10-11
Fix resovling wrong symbol when symbol table is not continous

zmj64351508:
Check the size of symbol before return to avoid resovling wrong symbol.


brancz (approved):
Good catch, thank you!

PR #5186, opened 2024-10-15
pkg/parcacol: Query flame graph metadata labels

metalmatze:
Rather than querying all possible labels we want to query only for those label names that would exist within the flame graph being queried.

### DIFF
--- a/pkg/parcacol/querier.go
+++ b/pkg/parcacol/querier.go
@@ -1541,37 +1541,43 @@ func (q *Querier) GetProfileMetadataMappings(
 
 func (q *Querier) GetProfileMetadataLabels(
 	ctx context.Context,
-	start, end time.Time,
+	query string,
+	startTime, endTime time.Time,
 ) ([]string, error) {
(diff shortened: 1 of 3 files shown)


pkg/parcacol/querier.go
```diff
@@ -1541,37 +1541,43 @@ func (q *Querier) GetProfileMetadataMappings(
...
+	err = q.engine.ScanTable(q.tableName).
+		Filter(filterExpr).
+		Project(logicalplan.DynCol("labels")).
+		Execute(ctx, func(ctx context.Context, ar arrow.Record) error {
+			for i, field := range ar.Schema().Fields() {
+				nulls := ar.Column(i).NullN()
+				rows := int(ar.NumRows())
+				if nulls == rows {
+					// This column only has nulls.
+					// Therefore, it's not part of the label set to group by.
```
brancz, line 1576: good catch!

brancz (approved):

PR #5244, opened 2024-10-29
area/ui: Ensure all dates are in UTC

yomete:
We should be displaying all date and time in the UTC 24H format.
image

Replaces all usage of `lll` with `YYYY-MM-DD HH:mm:ss` as `lll` formats the date to be "Mar 20, 2024 11:30 AM" for example.
image


metalmatze (approved):
nice! 👍

PR #5245, opened 2024-10-29
area/ui: Fix bug of disappearing matchers

yomete:
Fixes a bug where when in the simple querying mode and you add a new set of label matchers, and then proceed to move the mouse away before adding a label name, operator or value., the label matcher set is removed entirely, which can be frustrating.

Also fixes the bug where the regex input field gets so small and you don't get to see what is being typed.


metalmatze (approved):
Very nice! Works like a charm now!

PR #5276, opened 2024-11-07
area/ui: rejig dashboard buttons

yomete:
This is a PR that restructures the toolbar buttons.

Summary:
- The "Sort by" button is moved within the context of the icicle graph.
- The "Table columns" button is moved within the context of the table component.
- The table component now has a column to display the color which can either be the filename or binary.
- Hiding binaries now works with the table component
- The color legend is now above both the icicle and table visualisations. it was previously located above the icicle viz only.


manojVivek (approved):
Lgtm!

PR #5288, opened 2024-11-11
area/ui: Enhance regex filtering in SimpleMatchers component

yomete:
This fixes the following comments:

- When clicking on a value in the dropdown I expect the selected value to be set as the value in the input of the dropdown. Further on, this is going to be helpful as I only need to adjust some parts of the values for the regex from prometheus-k8s-0 to prometheus-k8s-\\d rather than still having to type the entire thing.

- Something about the Add button still seems weird. It's still floating above the input text. (24s in the video). Can we have two vertical divs? One with the input and one below with the button? Then it'll always be below and never next or above the input.


yomete:
> Very, very nice! That is exactly what I imagined. 🥳 Love the detail that the copy into the input field only works on regex selectors. 💯 Ship it! 🎉

awesome!

Issue #5291, opened 2024-11-15
Failure to symbolize binary


danipozo:
Parca can't symbolize addresses for ClickHouse binaries.

### How to reproduce

First, download and decompress ClickHouse binary:
```
curl -LO https://s3.amazonaws.com/clickhouse-builds/24.10/2e054d555719f509b96e0cfea4d5f5819c69a61f/package_release/clickhouse
chmod +x clickhouse
./clickhouse --version
```

<details>

<summary>Normal flow reproducer</summary>
Then, execute ClickHouse server:

```
mkdir ch-data && cd ch-data
../clickhouse server
```
Try to profile it with Parca (you will need to compile parca-agent with the latest version of the upstream OTel eBPF profiler to get the fix for [this](https://github.com/open-telemetry/opentelemetry-ebpf-profiler/issues/230)). Also use `--debuginfo-upload-max-size=2244458216` to be able to upload the binary debuginfo. You will observe that stacks are missing symbols in the UI.
</details>

<details>

<summary>Smaller reproducer</summary>

Compile and run this as a binary in this repo (pardon my poor Go):

```go
package main

import (
	"context"
	"debug/elf"
	"os"

	"github.com/go-kit/log"
	"github.com/go-kit/log/level"
	"github.com/parca-dev/parca/pkg/symbol/addr2line"
	"github.com/parca-dev/parca/pkg/symbol/demangle"
)

func main() {
	ctx := context.Background()

  path := "clickhouse"
  f, err := elf.Open(path)
  if err != nil {
    print(err)
  }

  demangler := demangle.NewDemangler("simple", false)
  writer := log.NewSyncWriter(os.Stdout)
  logger := log.NewLogfmtLogger(writer)
  liner, err := addr2line.DWARF(logger, path, f, demangler)
  if err != nil {
    print(err)
  }

  lines, err := liner.PCToLines(ctx, 0x1192c860)
  if err != nil {
    level.Error(logger).Log("err", err)
  }
}
```

</details>

### Error logs

These come ultimately from `debug/dwarf/entry.go`, method `dwarf5Ranges`.

Parca:
```
level=debug name=parca ts=2024-11-15T12:39:31.37757694Z caller=symbolizer.go:172 component=symbolizer msg="failed to get lines" err="liner pctolines: failed to extract dwarf tree: invalid rnglist offset 409180350 (max 636512)"
```
Reproducer bin:
```
[...]

PR #5299, opened 2024-11-19
symbolizer: shell out to addr2line

danipozo:
Use `addr2line` instead of custom addr2line implementation which doesn't symbolize addresses properly for some binaries (see #5291). Even with a fix for https://github.com/go-delve/delve/issues/3861, which is the root cause of the failure in #5291, the current implementation can't symbolize ~50 % of the addresses in the stacks of the ClickHouse binary in my tests. Using [gimli-rs/addr2line](https://github.com/gimli-rs/addr2line) as the `addr2line` implementation works fast and well:
image

Please take this PR as a rough proposal, there are of course details to be fleshed out:
- Probably you want to keep the current implementation under a setting
- If not, probably some code would need to be deleted
- This introduces another dependency, docs and Docker images at least would have to be updated
- Also there are likely code issues, I'm far from a Go expert

### DIFF
--- a/pkg/symbolizer/symbolizer.go
+++ b/pkg/symbolizer/symbolizer.go
@@ -14,18 +14,23 @@
 package symbolizer
 
 import (
+	"bufio"
 	"context"
 	"debug/elf"
 	"errors"
 	"fmt"
 	"io"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"strconv"
+	"strings"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 
 	debuginfopb "github.com/parca-dev/parca/gen/proto/go/parca/debuginfo/v1alpha1"
+	profilepb "github.com/parca-dev/parca/gen/proto/go/parca/metastore/v1alpha1"
 	"github.com/parca-dev/parca/pkg/debuginfo"
 	"github.com/parca-dev/parca/pkg/profile"
 	"github.com/parca-dev/parca/pkg/symbol/addr2line"
@@ -63,9 +68,10 @@ func WithDemangleMode(mode string) Option {
 type Symbolizer struct {
 	logger log.Logger
 
-	debuginfo DebuginfoFetcher
-	cache     SymbolizerCache
-	metadata  DebuginfoMetadata
+	debuginfo             DebuginfoFetcher
+	cache                 SymbolizerCache
+	metadata              DebuginfoMetadata
+	externalAddr2linePath string
 
 	demangler *demangle.Demangler
 
@@ -89,6 +95,7 @@ func New(
 	cache SymbolizerCache,
 	debuginfo DebuginfoFetcher,
 	tmpDir string,
+	externalAddr2linePath string,
 	opts ...Option,
 ) *Symbolizer {
 	const (
@@ -96,12 +103,13 @@ func New(
 	)
 
 	s := &Symbolizer{
-		logger:    log.With(logger, "component", "symbolizer"),
-		cache:     cache,
-		debuginfo: debuginfo,
-		tmpDir:    tmpDir,
-		metadata:  metadata,
-		demangler: demangle.NewDemangler(defaultDemangleMode, false),
+		logger:                log.With(logger, "component", "symbolizer"),
+		cache:                 cache,
+		debuginfo:             debuginfo,
+		externalAddr2linePath: externalAddr2linePath,
+		tmpDir:                tmpDir,
+		metadata:              metadata,
+		demangler:             demangle.NewDemangler(defaultDemangleMode, false),
 	}
 
 	for _, opt := range opts {
@@ -126,7 +134,7 @@ func (s *Symbolizer) Symbolize(
 	req SymbolizationRequest,
 ) error {
 	if err := s.symbolize(ctx, req); err != nil {
-		level.Debug(s.logger).Log("msg", "failed to symbolize", "err", err)
+		level.Debug(s.logger).Log("msg", fmt.Sprintf("failed to symbolize: BuildID %s", req.BuildID), "err", err)
 	}
 
 	return nil
@@ -272,17 +280,122 @@ func (s *Symbolizer) getDebuginfo(ctx context.Context, buildID string) (string, 
 }
 
 type cachedLiner struct {
-	logger    log.Logger
-	demangler *demangle.Demangler
-	filepath  string
-	f         *elf.File
-	quality   *debuginfopb.DebuginfoQuality
-	buildID   string
+	logger                log.Logger
+	demangler             *demangle.Demangler
+	filepath              string
+	f                     *elf.File
+	quality               *debuginfopb.DebuginfoQuality
+	buildID               string
+	externalAddr2linePath string
 
 	// this is the concrete liner
 	liner liner
 
 	cache SymbolizerCache
+}
+
+type externalExecutableLiner struct {
+	cmd       exec.Cmd
+	inputPipe io.WriteCloser
+	scanner   *bufio.Scanner
+
+	logger    log.Logger
+	demangler *demangle.Demangler
+}
+
+func (c *cachedLiner) newExternalExecutableLiner(
+	filepath string,
+) (liner, error) {
+	// Create cmd with pipes and scanner for reading output line by line
+	addr2lineCmd := exec.Command(c.externalAddr2linePath, "--exe", filepath, "-afiC")
+	addr2lineInput, err := addr2lineCmd.StdinPipe()
+	if err != nil {
+		return nil, fmt.Errorf("error creating input pipe for addr2line: %w", err)
+	}
+	addr2lineOutput, err := addr2lineCmd.StdoutPipe()
+	if err != nil {
+		return nil, fmt.Errorf("error creating output pipe for addr2line: %w", err)
+	}
+	scanner := bufio.NewScanner(addr2lineOutput)
+
+	// Start command to load executable info and wait for input
+	err = addr2lineCmd.Start()
+	if err != nil {
+		return nil, fmt.Errorf("error starting addr2line: %w", err)
+	}
+
+	return &externalExecutableLiner{
+		cmd:       *addr2lineCmd,
+		inputPipe: addr2lineInput,
+		scanner:   scanner,
+		logger:    c.logger,
+		demangler: c.demangler,
+	}, nil
+}
+
+func (l *externalExecutableLiner) Close() error {
+	err := l.inputPipe.Close()
+	if err != nil {
+		return err
+	}
+	err = l.cmd.Wait()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (l *externalExecutableLiner) PCToLines(ctx context.Context, pc uint64) ([]profile.LocationLine, error) {
+	// Write the address to addr2line stdin twice in order to get addresses as a proper delimiter
+	for i := 0; i < 2; i++ {
+		_, err := io.WriteString(l.inputPipe, fmt.Sprintf("0x%x\n", pc))
+		if err != nil {
+			return nil, fmt.Errorf("error writing to addr2line stdin: %w\n", err)
+		}
+	}
+
+	lines := []profile.LocationLine{}
+	// addr2line output consists of a variable number of pairs of lines per address
+	// The first line of each block of output is the address, if we have to skip additional lines
+	// it means the previous read loop left some output unread
+	l.scanner.Scan()
+	addressLine := l.scanner.Text()
+	for !strings.HasPrefix(addressLine, "0x") {
+		l.scanner.Scan()
+		addressLine = l.scanner.Text()
+	}
+
+	keepReading := true
+	l.scanner.Scan()
+	for keepReading {
+		line1 := l.scanner.Text()
+
+		l.scanner.Scan()
+		line2 := l.scanner.Text()
+
+		line1 = strings.TrimSuffix(line1, "\n")
+		line2 = strings.TrimSuffix(line2, "\n")
+
+		fileLocation := strings.Split(line2, ":")
+		lineNum, err := strconv.Atoi(fileLocation[1])
+		if err != nil {
+			lineNum = 0
+		}
+
+		lines = append(lines, profile.LocationLine{
+			Line: int64(lineNum),
+			Function: &profilepb.Function{
+				StartLine: int64(lineNum),
+				Filename:  fileLocation[0],
+				Name:      line1,
+			},
+		})
+
+		l.scanner.Scan()
+		keepReading = !strings.HasPrefix(l.scanner.Text(), "0x")
+	}
+
+	return lines, nil
 }
 
 // newConcreteLiner creates a new liner for the given mapping and object file path.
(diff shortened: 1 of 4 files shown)


brancz:
I'm open to the idea of this though I would much rather actually fix the symbolizer, but I'd be ok with this being an escape hatch. It's definitely possible to write a correct symbolizer even with the bug in the delve code (we did in Polar Signals Colud but what we did doesn't work in a single-process context that we're constrained to in Parca).

danipozo:
Nice to read, I'll add a setting for this then and maybe some docs on how to use it if you face broken symbolization for your binary

danipozo:
I've added a setting to enable this functionality by specifying a path to the `addr2line` binary, and a new liner that uses this binary.  This liner can also be wrapped by the `cachedLiner` and therefore use the cache.

danipozo:
Hi, kindly asking for another round of review here @brancz

pkg/symbolizer/symbolizer.go
```diff
@@ -268,19 +276,103 @@ func (s *Symbolizer) getDebuginfo(ctx context.Context, buildID string) (string,
...
+	_, err := io.WriteString(l.inputPipe, fmt.Sprintf("0x%x\n", pc))
+	if err != nil {
+		return nil, fmt.Errorf("error writing to addr2line stdin: %w\n", err)
+	}
+	// addr2line output consists of two lines per address
+	line1, err := l.scanner.ReadString('\n')
+	if err != nil {
+		return nil, fmt.Errorf("error scanning addr2line output: %w\n", err)
+	}
+	line2, err := l.scanner.ReadString('\n')
```
brancz, line 351: I think the output can be more than this. There will be additional lines for each inlined function.

brancz (commented):
Very nice! I think we're really close, sorry for missing this before. Happy new year!

pkg/symbolizer/symbolizer.go
danipozo: Looking at the [man page](https://www.man7.org/linux/man-pages/man1/addr2line.1.html) for addr2line, when not using the `-i` flag, it won't display additional lines for inlined functions. This is interesting but I think it's not needed in this case, if I'm not wrong the agent already sends information for inlined functions and this liner will process them one by one. At least when profiling ClickHouse using this liner I get traces with `unalignedLoad` which is inlined:
image

brancz: That doesn't make sense to me. Fundamentally inlining means that at a single program counter, we've inlined one or more functions, so to reflect the correct stack trace we need to also resolve all inlined functions at that program counter.

danipozo: You're right, I checked and though `unalignedLoad` is inlined, it's only the innermost inlined call at that point, `addr2line` with `-i` returns three other functions that aren't shown in that stacktrace.

So it would be ideal to pass the `-i` flag and check all lines returned by `addr2line`, but I'm struggling to do so because I don't know beforehand how many lines will be output and I can't check them without blocking on reading from the process' stdout pipe (even calling `Peek()` blocks). I've seen the standard solution seems to involve using an auxiliary goroutine that reads from the pipe to then `select` on that, but that looks a bit messy to me, WDYT?

brancz: I would read the whole output and then just process the string without having to deal with the streaming nature.

danipozo: Hmm yeah that would be great but AFAICT there is no easy way to read the whole output without blocking to read potentially more output, as the pipe for the process standard output will still be open until the liner is closed

brancz: I would create the `os/exec.Command` with the `Stdin` field set to the input you want to pipe to it, and then call `.Output()` on it. That should give you all the output until the command exits, and it closes the stdin once the `io.Reader` passed in via the `Stdin` field is exhausted.

danipozo: The problem with that would be that I could not return lines information immediately after each call to `PCToLines()` [here](https://github.com/parca-dev/parca/pull/5299/files#diff-e23225ba333f3087a99b5816cbf26bb6fa3a0a1c040c66f04609be27c8b9c0c1R178).

I've tried creating an auxiliary channel to be able to read from the stdout of addr2line in a nonblocking way and that works *but* as the main routine might try to read from that output before it is available, it sometimes leads to skipping lines, so it is unreliable.

What I've done instead in the latest push is something that feels, and actually is, a bit of a hack but works great and gets complete stack traces with all inlined functions. I've added the `-a` flag to the process invocation, which causes it to output a line with the address before the rest of the output. That serves as a delimiter between blocks of output. Then, what I do is pass the program counter twice, and read the output until we find another address. The leftover lines after the second address will be skipped by the next read.

Screenshot with the proper stack trace, with the methods inlined before `unalignedLoad`:
image

brancz: I see now, you are keeping the addr2line process alive for the entire time (clever!). I think this might only work with llvm-addr2line, but with that we could use `llvm-addr2line --output-style=JSON`, then everything is in one line and easy to parse with it being json.

danipozo: Trying to keep this compatible with standard addr2line, that way it can be used with the standard, LLVM one, and also [gimli's](https://github.com/gimli-rs/addr2line) which is the one I'm using since it seems to have the best performance and completeness resolving addresses.

chaochaoxiaochao:
Hi, I used this MR locally for testing and found that there are still many issues?? The output of using addr2line-e libc.so.6 0xe7f9b is:
??:?
But the output of using addr2line-e libc.so.6-f 0xe7f9b is
__clone
??:?
I think this is what we want. It can directly display the function name and also serve as the fastest way to view and fix the problem

chaochaoxiaochao:
> Hi，我在本地用这个MR测试了一下，发现还是有很多问题？？使用addr2line -e libc.so.6 0xe7f9b输出的是： ??:? 但是使用addr2line -e libc.so.6 -f 0xe7f9b输出的是 __clone ??:? 我觉得这个才是我们想要的，既能直接显示函数名，又能作为查看和修复问题最快的方式

@danipozo Do you have any constructive suggestions? Thank you

chaochaoxiaochao:
> > Hi, I used this MR locally for testing and found that there are still many issues?? The output of using addr2line-e libc.so.6 0xe7f9b is: ??:? But the output of using addr2line-e libc.so.6-f 0xe7f9b is __clone ??:? I think this is what we want. It can directly display the function name and also serve as the fastest way to view and fix the problem
> 
> @danipozo Do you have any constructive suggestions? Thank you

And I followed the method in your MR:
addr2line --exe libstdc++.so.6.0.30  -afiC  0xce9db
The output is:
0x00000000000ce9db
std::error_code::default_error_condition() const
??:?
There is also output, but the actual UI does not display it

danipozo:
@chaochaoxiaochao can you add some more specific details about how you're using this and what issues you are finding? Along with proper formatting it would go a great length towards letting me understand your concerns.

> The output of using addr2line-e libc.so.6 0xe7f9b is:

> But the output of using addr2line-e libc.so.6-f 0xe7f9b is

In particular I don't follow what you're trying to say here since these two command invocations are the same.

Thanks!

danipozo:
> since these two command invocations are the same.

Ah now I see in one you're using the `-f` flag while in the other you aren't. So yes, in one case `addr2line` will try to show you the function name while in the other it won't, but I still don't understand your concern. In this PR I'm using `-afiC` to get function names, for all inlined functions at an address, and demangling those if needed. The `-a` flag is used to be able to read from the program output without blocking, as explained [in the thread](https://github.com/parca-dev/parca/pull/5299#discussion_r1919921146)

chaochaoxiaochao:
> > 因为这两个命令调用是相同的。
> 
> 啊，现在我看到在其中一个中你使用了`-f`标志，而在另一种中你没有使用。所以是的，在一种情况下`addr2line`会尝试向你显示函数名称，而在另一种情况下则不会，但我仍然不明白你的担忧。在这个 PR 中，我使用它`-afiC`来获取地址处所有内联函数的函数名称，并在需要时对其进行分解。该`-a`标志用于能够从程序输出中读取而不会阻塞，如[线程中所述](https://github.com/parca-dev/parca/pull/5299#discussion_r1919921146)

As I mentioned earlier, I attempted to output the function name locally using the - fiC 0xce9db method, but the actual UI did not display it. The UI displayed ????

chaochaoxiaochao:
> > > 因为这两个命令调用是相同的。
> > 
> > 
> > 啊，现在我看到其中一个中你使用了`-f`标志，而在另一种情况下你没有使用。所以是的，在一种情况下`addr2line`会尝试向你显示函数名称，而在另一种情况下则它不会，但我仍然不明白你的担忧。在这个 PR 中，我使用`-afiC`来获取地址处所有内联函数的函数名称，并在需要时对其进行排序。该`-a`标志用于能够从程序输出中读取而不会阻塞，如[线程中所述](https://github.com/parca-dev/parca/pull/5299#discussion_r1919921146)
> 
> 前面提到过，我尝试使用 -fiC 0xce9db 方法在本地输出函数名，但是实际 UI 并没有显示出来。UI 显示的????


image

chaochaoxiaochao:
> > > > 因为这两个命令调用是相同的。
> > > 
> > > 
> > > 啊，现在我看到其中一个中你使用了`-f`标志，而在另一种情况下你没有使用。所以是的，在一种情况下`addr2line`会尝试向你显示函数名称，而在另一种情况下则它不会，但我仍然不明白你的担忧。在这个 PR 中，我使用`-afiC`来获取地址处所有内联函数的函数名称，并在需要时对其进行排序。该`-a`标志用于能够从程序输出中读取而不会阻塞，如[线程中所述](https://github.com/parca-dev/parca/pull/5299#discussion_r1919921146)
> > 
> > 
> > 前面提到过，我尝试使用 -fiC 0xce9db 方法在本地输出函数名，但是实际 UI 并没有显示出来。UI 显示的????
> 
> image

I tried to check the terminal but there were no error messages. I opened it at debug level. It's embarrassing that I couldn't compile parca locally. Otherwise, I could investigate the reason

chaochaoxiaochao:
> > As I mentioned earlier, I attempted to output the function name locally using the - fiC 0xce9db method, but the actual UI did not display it. The UI displayed ????
> 
> As I mentioned earlier, I attempted to output the function name locally using the - fiC 0xce9db method, but the actual UI did not display it. The UI displayed ????

image

I tried to check the terminal but there were no error messages. I opened it at debug level. It's embarrassing that I couldn't compile parca locally. Otherwise, I could investigate the reason

chaochaoxiaochao:
> > > As I mentioned earlier, I attempted to output the function name locally using the - fiC 0xce9db method, but the actual UI did not display it. The UI displayed ????
> > 
> > 
> > As I mentioned earlier, I attempted to output the function name locally using the - fiC 0xce9db method, but the actual UI did not display it. The UI displayed ????
> 
> image
> 
> I tried to check the terminal but there were no error messages. I opened it at debug level. It's embarrassing that I couldn't compile parca locally. Otherwise, I could investigate the reason

The specific issue you mentioned earlier is actually that I can display the function name locally, but the UI shows that ??

danipozo:
@chaochaoxiaochao OK, I think I understand better now.

To debug this in your specific case, you would need to:
- Get the information that is shown on the tooltip when hovering one of the problematic addresses, specifically the binary name, build ID and address
- Make sure you are running `addr2line` against the same binary, you can get do so by checking the build ID using the `file` utility
- Make sure the `addr2line` version you're running and the one Parca is using are the same, different `addr2line` versions might have trouble with different binaries and addresses, although `llvm-addr2line` and gimli's should mostly work

Also please, can you try to stick to condensing the information in 1-2 messages in a row? This way it is much easier for me and probably for reviewers to follow, thanks!

chaochaoxiaochao:
> @chaochaoxiaochao好的，我想我现在明白了一点。
> 
> 为了在您的具体情况下调试这个问题，您需要：
> 
> * 获取将鼠标悬停在有问题的地址之一上时显示在工具提示上的信息，特别是二进制名称、构建 ID 和地址
> * 确保你正在运行`addr2line`相同的二进制文件，你可以通过使用`file`实用程序检查构建 ID 来实现这一点
> * 确保`addr2line`你运行的版本和 Parca 使用的版本相同，不同`addr2line`版本可能会因不同的二进制文件和地址而出现问题，尽管`llvm-addr2line`gimli 的版本应该基本可以正常工作
> 
> 另外，您能否尝试将信息浓缩为连续 1-2 条消息？这样我和审阅者就更容易理解了，谢谢！




> @chaochaoxiaochao OK, I think I understand better now.
> 
> To debug this in your specific case, you would need to:
> 
> * Get the information that is shown on the tooltip when hovering one of the problematic addresses, specifically the binary name, build ID and address
> * Make sure you are running `addr2line` against the same binary, you can get do so by checking the build ID using the `file` utility
> * Make sure the `addr2line` version you're running and the one Parca is using are the same, different `addr2line` versions might have trouble with different binaries and addresses, although `llvm-addr2line` and gimli's should mostly work
> 
> Also please, can you try to stick to condensing the information in 1-2 messages in a row? This way it is much easier for me and probably for reviewers to follow, thanks!

Sorry, I'll streamline my response：
1. 
企业微信截图_17376308021938
2. yes:
libc.so.6: ELF 64-bit LSB shared object, ARM aarch64, version 1 (GNU/Linux), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, BuildID[sha1]=09928b270aa19314161b21f565d1a9732c2c5332, for GNU/Linux 3.7.0, stripped
3. The gimli I am using is version 0.16.0
4. Please add the following scene:
The server runs on X86
Agent runs on Arm
Deploy separately on different machines

Thank you for supporting and viewing

chaochaoxiaochao:
@danipozo I tried to debug and compile libc locally before uploading debuginfo, but the result is still the same??

danipozo:
@chaochaoxiaochao I find it hard to keep debugging your use case, sorry. Please let's try to focus on code review here, I suggest that you open an issue to collect the details of your case and maybe find the cause at some point if/when this is used by more people or we find bugs that could be related.

@brancz the latest commit adds some error handling and a bigger buffer size to fix an edge case (~80 kiB output for a single address) that I found for my binary.

pkg/symbolizer/symbolizer.go
danipozo: Resolving this conversation, feel free to reopen if you consider it necessary

brancz:
If I understand correctly, we're saying that we're ok with not getting inlined frames? I think I'm generally ok with this since we'd like to ultimately just fix the symbolizer in the first place.

danipozo:
> If I understand correctly, we're saying that we're ok with not getting inlined frames? I think I'm generally ok with this since we'd like to ultimately just fix the symbolizer in the first place.

No, actually the latest solution described in [this comment](https://github.com/parca-dev/parca/pull/5299#discussion_r1919921146) lets us get inlined frames from `addr2line` output!

> Apologies (I accidentally misclicked and closed the PR but reopened immediately).

Misclicks happen :)

brancz (approved):
Ohhh awesome, I'm sorry I didn't understand before why that would work, but now makes sense!

Let's get this merged!

danipozo:
Great, thanks for taking the time to review!

brancz:
Of course, apologies it took so long, thank you very much for bearing with me!

PR #5306, opened 2024-11-20
remove unnecessary non-nil == nil check

vidurkhanal:
the header == nil check is unnecessary. the header was established to be non nil from couple of LOC before the switch statements.


brancz (approved):

brancz:
Thank you!

PR #5364, opened 2024-12-05
area/ui: Add a popover menu to the AreaGraph component

yomete:


ui/packages/shared/profile/src/MetricsGraphStrips/AreaGraph/Tooltip.tsx
```diff
@@ -0,0 +1,83 @@
...
+        left: `${tooltipPosition.x}px`,
+        top: `${tooltipPosition.y}px`,
+        pointerEvents: 'none',
+      }}
+    >
+      <div className="flex flex-col gap-1">
+        <div className="flex gap-1 items-center">
+          <div>Timestamp:</div>
+          <div className="text-gray-600 dark:text-gray-300">
+            {formatDate(timestamp, timePattern(timezone), timezone)}
```
manojVivek, line 70: I think it would make more sense to display the time with milliseconds precision here. 
When the timeframe is tight, we'll have multiple points within a second and milliseconds precision would really help.

manojVivek (approved):
One minor change, looks great otherwise.

PR #5381, opened 2024-12-11
area/ui: auto close the group by dropdown when there's an outside click

### LINKED ISSUE #5378: Group-by menu doesn't close as expected
When we open the "more" menu and click elsewhere, it disappears as expected. If we do the same with the group-by menu it doesn't.

https://github.com/user-attachments/assets/e587e355-214a-4a69-aa87-06d0d7fbf992

yomete:
Fixes #5378


manojVivek (approved):
Lgtm!

PR #5383, opened 2024-12-12
*: Icicle charts support

manojVivek:


pkg/query/flamegraph_arrow.go
```diff
@@ -194,6 +198,43 @@ func generateFlamegraphArrowRecord(ctx context.Context, mem memory.Allocator, tr
...
+			if fb.aggregationConfig.aggregateByTimestamp {
+				tsHasher.Reset()
+				tsStr := strconv.FormatInt(r.Timestamp.Value(i), 10)
+				durationStr := strconv.FormatInt(r.Duration.Value(i), 10)
+				_, _ = tsHasher.Write([]byte(tsStr))
+				_, _ = tsHasher.Write([]byte(durationStr))
+				tsHash = tsHasher.Sum64()
+
+				sampleTsRow := row
+				if row, ok := fb.rootsRow[tsHash]; ok {
```
brancz, line 210: for flamecharts having a root for a timestamp that already exists means we can't possibly build a correct flamechart, we should error on this case

pkg/query/flamegraph_arrow.go
```diff
@@ -284,6 +325,10 @@ func generateFlamegraphArrowRecord(ctx context.Context, mem memory.Allocator, tr
 					if fb.aggregationConfig.aggregateByLocationAddress {
 						key = hashCombine(key, r.Address.Value(j))
 					}
+					if fb.aggregationConfig.aggregateByTimestamp {
+						key = hashCombine(key, uint64(r.Timestamp.Value(i)))
+						key = hashCombine(key, uint64(r.Duration.Value(i)))
```
brancz, line 330: given the other comments, we shoudn't need the duration to be part of the hash

pkg/query/flamegraph_arrow.go
```diff
@@ -1330,9 +1398,29 @@ func (fb *flamegraphBuilder) appendRow(
...
+func appendGroupByMetadata(fb *flamegraphBuilder, r *profile.RecordReader, sampleRow int) {
+	fb.builderGroupByMetadata.Append(true)
+	for i := 0; i < fb.builderGroupByMetadata.NumField(); i++ {
+		n := fb.groupByMetadataFields[i].Name
+		b := fb.builderGroupByMetadata.FieldBuilder(i).(*array.BinaryBuilder)
+		switch n {
+		case profile.ColumnTimestamp:
+			ts := r.Timestamp.Value(sampleRow)
+			b.Append([]byte(fmt.Sprint(ts)))
+		case profile.ColumnDuration:
```
brancz, line 1415: same (we can use `value` to tell us how wide the frame should be)

brancz (approved):
If I'm understanding this right, we don't yet have the ability to merge stacks that are "next to each other", correct? I'm not sure yet whether we will need to re-think whether flamegraphs and flamecharts can be rendered with the same code since one cares about order and the other doesn't.

pkg/query/flamegraph_arrow.go
manojVivek: I'm sorry I don't understand why this has to be errored. 

So what I'm trying to here is convert the following records
```
[{
    ts: timestamp-1
    samples: ....
  },
  {
    ts: timestamp-2
    samples: ....
  },
  {
    ts: timestamp-1
    samples: .....
}]
```
to 
```
[{
    ts:timestamp-1
    samples: ....
  },
  {
    ts: tiemstamp-2
    samples: ....
}]
```

So that we'll have all samples with the same timestamp are grouped under a single root, that is easier for rendering in the UI. 
Isn't this similar to the group by label here:
Screenshot 2024-12-16 at 3 03 30 PM

brancz: There can't ever be two samples at the same timestamp for the same thread or CPU. Merging should happen when `end timestamp == next timestamp` (within some error of margin because clocks aren't perfectly precise, maybe 1-10 ms).

brancz: And `end timestamp = timestamp + value`

PR #5421, opened 2025-01-07
ui: Small improvements

metalmatze:
## Metrics Graph 

The MetricsGraph Tooltip now shows the "Per Second" in one line rather than having it split in two. All that was necessary was using a `&nbsp;`. 

Old (two lines)
old-tooltip

New (one line)
new-tooltip

## Query Selectors 

When multiple selectors are used in the query UI, the buttons and dropdowns used to move down. This look odd and somewhat broken.

old-selectors

This is now fixed. 

new-selectors


yomete (approved):
lgtm!

PR #5423, opened 2025-01-08
Parca UI improvements

yomete:
- When using the simple querying mode and in the regex mode, the cursor will now always be positioned at the end of the text.
- The border radius of the binary in the Legend and Table components is now uniform.
- The binary column in the Table component now shows the binary name (via a tooltip) when hovered on.

### DIFF
--- a/ui/packages/shared/profile/src/SimpleMatchers/Select.tsx
+++ b/ui/packages/shared/profile/src/SimpleMatchers/Select.tsx
@@ -183,6 +183,12 @@ const CustomSelect: React.FC<CustomSelectProps> = ({
     }
   };
 
+  const moveCaretToEnd = (e: React.FocusEvent<HTMLTextAreaElement>): void => {
+    const value = e.target.value;
+    e.target.value = '';
+    e.target.value = value;
+  };
+
   return (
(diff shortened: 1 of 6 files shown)


ui/packages/shared/profile/src/SimpleMatchers/Select.tsx
```diff
@@ -183,6 +183,12 @@ const CustomSelect: React.FC<CustomSelectProps> = ({
     }
   };
 
+  const moveCaretToEnd = (e: React.FocusEvent<HTMLTextAreaElement>): void => {
+    const value = e.target.value;
+    e.target.value = '';
+    e.target.value = value;
+  };
```
metalmatze, line 190: Very interesting.

metalmatze (approved):
flawless!

ui/packages/shared/profile/src/SimpleMatchers/Select.tsx
yomete: ikr, its apparently a common trick to force the cursor to move to the end of the text. When the value is set programmatically like this, the browser automatically places the cursor at the end of the new value.

PR #5467, opened 2025-01-29
Add `--otlp-insecure` flag

### LINKED ISSUE #5466: Enable sending traces to a tracing backend backed by TLS
I have a kubernetes cluster where `parca` and `tempo` are deployed. Tempo is running with TLS using self signed certificates and I want to configure `parca` to send its traces to that Tempo instance.

The current OTLP flag provided by parca: `--otlp-address` seem to not work when the tracing backend is running with TLS.

This is likely due to using `WithInsecure`:  https://github.com/parca-dev/parca/blob/03cff404aaaee69ab72a530636593345fd5b0900/pkg/tracer/tracer.go#L149
and not having a mechanism to pass TLSConfig,
we can expose another flag `--otlp-insecure` default to True for backwards compatibility and then check if its set to `true`, use `WithInsecure()` else don't set that flag.

michaeldmitry:
Currently, using `WithInsecure` for http and gRPC tracing exporters fails to export to tracing backends that are running with TLS config.
 

Fixes #5466

### DIFF
--- a/pkg/parca/parca.go
+++ b/pkg/parca/parca.go
@@ -139,6 +139,7 @@ type FlagsLogs struct {
 type FlagsOTLP struct {
 	Address  string `help:"The endpoint to send OTLP traces to."`
 	Exporter string `default:"grpc"                              enum:"grpc,http,stdout" help:"The OTLP exporter to use."`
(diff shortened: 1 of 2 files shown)


brancz (approved):
Thank you!

PR #5489, opened 2025-02-14
pkg/query: Sort samples before generating flame charts

metalmatze:
If the samples aren sorted we don't do anything. If there are multiple records we first sort them and then merge them into a single final sorted record.

The real world usage depends on changes from https://github.com/polarsignals/frostdb/pull/949


manojVivek (approved):
Lgtm!

PR #5539, opened 2025-02-25
area/ui: strip attributes from context menu

yomete:


yomete:
Merging this in to test upstream.

PR #5568, opened 2025-03-07
ui/profile/src/SimpleMatchers: Sort matches with levenshtein

metalmatze:
Previously `app_kubernetes_io_component` would have been at the top. But this isn't closest to the search term anymore so we show it last.

Screenshot 2025-03-07 at 12 14 29


yomete (approved):
lgtm

PR #5572, opened 2025-03-18
area/ui: Multiple metrics graph

yomete:


ui/packages/shared/profile/src/MetricsGraph/UtilizationMetrics/index.tsx
```diff
@@ -83,13 +84,26 @@ function transformToSeries(data: MetricSeries[]): Series[] {
...
+  switch (name) {
+    case 'GPU Utilization':
+      return 'percent';
+    case 'GPU Memory Utilization':
+      return 'percent';
+    case 'GPU Power':
+      return 'watts';
+    default:
+      return 'percent';
+  }
```
metalmatze, line 97: I'm not sure where the name variable comes from, but it feels like it should be based on the metric name and not on the axis label. Otherwise looks good to me.

metalmatze (approved):
One question. Otherwise looking great!

PR #5579, opened 2025-04-03
ui: Persist expanded iciclegraph state in URL

manojVivek:
This change persists the stack expansion state to the URL when a user interacts with the Iciclegraph, so that it can be easily shared.


brancz (approved):
Very nice!

PR #5589, opened 2025-04-17
ui: Unified Label Selector

manojVivek:
This PR includes changes that introduces additional capabilities to the existing `SimpleMatchers` to allow them to select both the profile labels and the utilization labels with a distinction in the UI to group the label names based on their type. 

Note:  This is only an optional feature and doesn't introduce any change in behavior to Parca as the component will behave as is.


yomete (commented):
lgtm!

yomete (approved):

PR #5628, opened 2025-05-01
send "group by" in export requests.

umanwizard:
Add the "group by" field to the query for pprof download and pprof.me export. This makes it so the selected label set will be available in the exported data.


umanwizard:
BTW, the labels still don't show up in pprof.me, but they do show up in the pprof you can download from pprof.me. So I suspect there is another issue on the pprof.me side that needs to be fixed.

brancz:
Just need to fix that lint, then this lgtm and we can have a separate look at pprof.me, but I suspect we just don't pass the group by labels through correctly in the pprof.me backend.

PR #5632, opened 2025-05-06
fix address for interpreter and kernel

zmj64351508:
Address is not shown for interpreter and kernel. Fix it by adding address when lines > 0.


brancz:
They don’t have addresses in the first place and how does this patch fix that?

zmj64351508:
In parca-agent, addresses are always set for each frame and sent to server. `insertKernelFrames` for kernel and `Symbolize/AppendFrameID` for each interpreter.

PR #5634, opened 2025-05-07
ui: Iciclechart improved error copy

manojVivek:
Includes:
- [x] Refinements to the previous error messages
- [x] Option to pass in additional help text through context

### DIFF
--- a/ui/packages/shared/profile/src/ProfileIcicleGraph/index.tsx
+++ b/ui/packages/shared/profile/src/ProfileIcicleGraph/index.tsx
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import React, {LegacyRef, useEffect, useMemo, useState} from 'react';
+import React, { LegacyRef, ReactNode, useEffect, useMemo, useState } from 'react';
 
 import {AnimatePresence, motion} from 'framer-motion';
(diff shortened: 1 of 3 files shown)


ui/packages/shared/profile/src/ProfileIcicleGraph/index.tsx
```diff
@@ -176,13 +180,41 @@ const ProfileIcicleGraph = function ProfileIcicleGraphNonMemo({
     // Do necessary checks to ensure that icicle chart can be rendered for this query.
     if (isInvalidIcicleChartQuery) {
       if (isNonDelta) {
-        return <ErrorContent errorMessage="Icicle chart is only available for delta profiles." />;
+        return (
+          <ErrorContent
+            errorMessage={
+              <>
+                <span>To use the Icicle chart, please switch to a Delta profile.</span>
```
brancz, line 187: Should we maybe just not allow people to select it in those cases in the first place?

brancz (approved):

ui/packages/shared/profile/src/ProfileIcicleGraph/index.tsx
manojVivek: UX wise I feel this would be better, just not allowing them to select could leave them puzzled on 'why?'.

PR #5640, opened 2025-05-13
feat: ui: Add support for dashed lines for limit metrics in graphs

metalmatze:
Filtered out virtual labels for unique key generation in colors. Introduced a `strokeDasharray` prop to allow dashed lines, applied to limit metrics for visual distinction.


metalmatze:
Will fix this bug, but the overall content of the PR is there, so please review. Thanks!

yomete (approved):
lgtm!

PR #5646, opened 2025-05-29
ui: Add "doesn't contain function" filtering to profile queries

manojVivek:
Screencast: 

https://github.com/user-attachments/assets/d2ebfdc8-26e2-4372-881b-56319d27645d


brancz (approved):
Very nice!

PR #5648, opened 2025-05-30
Revert "feat(debuginfo): implement ObjectSizer interface for UploadReader"

secfree:
This reverts commit b86663df0f51de0a54d726ac4acc94e34315f5d0.

#5641 uses "UploadReader.size" as "ObjectSize()", but "size" is not set when initializing, so "ObjectSize()" returns "0". Then it causes the uploaded "debuginfo" file to be size zero.

It seems currently the warning "could not guess file size for multipart upload; upload might be not optimized" has to be left.

### DIFF
--- a/pkg/debuginfo/reader.go
+++ b/pkg/debuginfo/reader.go
@@ -91,9 +91,3 @@ func contextError(ctx context.Context) error {
 		return nil
 	}
 }
-
-// ObjectSize implements the objstore.ObjectSizer interface to provide the total size of bytes read so far.
-// This is used by objstore.TryToGetSize to optimize multipart uploads by pre-allocating the correct buffer size.
-func (r *UploadReader) ObjectSize() (int64, error) {
-	return int64(r.size), nil
-}


secfree:
@brancz could you help review this PR? Thank you.

brancz:
Why not fix it by setting the `size`?

secfree:
> Why not fix it by setting the `size`?

I chose to revert the change instead of setting `size` for the following reasons:

- `UploadReader` is initialized at [this location](https://github.com/parca-dev/parca/blob/main/pkg/debuginfo/store.go#L368), where the only available parameter is `stream debuginfopb.DebuginfoService_UploadServer`. This stream does not provide any method to obtain the total size of the upload in advance.
- The `size` field in `UploadReader` is incremented as data is read via the `Read` method, so it only reflects the number of bytes that have been read so far — not the total object size. However, the `ObjectSizer.ObjectSize` interface is intended to return the complete size of the object in bytes. Using the current approach conflates these two concepts, which I believe is a design mismatch.

Thank you.

brancz (approved):

brancz:
Makes sense! I guess the right way to solve this would be to adjust the protocol and have the first message sent not contain any data, but just the total size, and only then start transferring data. I agree though that that needs a bit more work so reverting is the better way to go right now.

brancz:
(sorry misclicked to close and comment)

secfree:
Hi @brancz , can you help merge this one? Thanks.

brancz:
Thank you for the reminder!

PR #5651, opened 2025-05-30
pkg/query: Introduce value-offset

brancz:
Value offset allows each frame to be rendered in parallel by the frontend instead of recursively. Value offset determines the X offset, the depth of the frame the Y offset, and the cumulative the width.

This is the last piece missing that will allow the frontend to do this.

### DIFF
--- a/pkg/query/flamegraph_arrow.go
+++ b/pkg/query/flamegraph_arrow.go
@@ -61,8 +61,9 @@ const (
 	FlamegraphFieldFlat       = "flat"
 	FlamegraphFieldDiff       = "diff"
 
-	FlamegraphFieldTimestamp = "timestamp"
-	FlamegraphFieldDepth     = "depth"
+	FlamegraphFieldTimestamp   = "timestamp"
+	FlamegraphFieldDepth       = "depth"
+	FlamegraphFieldValueOffset = "value_offset"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/query/flamegraph_arrow.go
```diff
@@ -1788,11 +1811,17 @@ func (fb *flamegraphBuilder) trim(ctx context.Context, tracer trace.Tracer, thre
...
 			if v := fb.builderCumulative.Value(cr); v > int64(cumThreshold) {
 				// this row is above the threshold, so we need to keep it
 				// add this row to the queue to check its children.
-				trimmingQueue.push(trimmingElement{row: cr, parent: row})
+				trimmingQueue.push(trimmingElement{
+					row:         cr,
+					parent:      row,
+					valueOffset: valueOffset,
+				})
+				valueOffset += fb.builderCumulative.Value(cr)
```
metalmatze, line 1824: Ok. So the valueOffset is the offset from the other children relative to the parent's cumulative value, right? 
I just wanted to verify that we're not passing the pixel from the frontend and performing some x-y axis math in the backend. 
This makes sense to me.

brancz: correct it's the value relative to the total, the frontend can then use that to calculate the position based on the available pixel space

PR #5659, opened 2025-06-04
ui: Add the Icicle graph `sort by` dropdown back

yomete:


yomete:
Closing this for now as we need to implement sorting in the backend and then enable again in the frontend.

PR #5661, opened 2025-06-05
ui: Properly handle multiple encodings for query expressions

yomete:
This PR helps to handle query expressions that have been encoded multiple times and now breaking the app. It  safely decodes strings that may have multiple levels of URL encoding with the `decodeMultipleEncodings` function.

I've also added safegaurds to ensure that we're not unnecessarily encoding query expressions multiple times by doing all the encoding for query expressions in the `filterSuffix` function.


manojVivek (approved):
Lgtm!

PR #5703, opened 2025-06-18
ui: Fixed metrics graph alignment to the grid

manojVivek:
The metrics graph needed some padding to align it to the grid scale which in turn fixed the overflows on either side.


metalmatze (approved):
Great. Even zooming out on the page everything sits nice and flush. 👍

PR #5765, opened 2025-07-03
fix: add LICENSE file to proto directory for buf.build packages

### LINKED ISSUE #5764: LICENSE file is missing in buf.build for `/parca-dev/parca*`
Hello mates! 

As you might know, we are depending on two parca packages for the parca datasource in Grafana. 

```
buf.build/gen/go/parca-dev/parca/protocolbuffers/go
buf.build/gen/go/parca-dev/parca/connectrpc/go
```

The buf build packages does however not include a LICENSE file which triggers warnings in source code scanners that we are using. Having to configure overrides for these packages are painful and all scanners might not have good support for it either. 

cc @brancz @metalmatze 

```bash
pwd
/Users/carlbergquist/go/pkg/mod/buf.build/gen/go/parca-dev/parca/protocolbuffers/go@v1.36.6-20250618095748-e69367cffaca.1

ls
go.mod google parca

```

metalmatze:
## Summary
Adds LICENSE file to `proto/` directory so buf.build packages include license information.

Fixes #5764

### DIFF
--- a/proto/LICENSE
+++ b/proto/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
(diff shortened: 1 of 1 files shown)


metalmatze:
I thought maybe there's a fancy way, or we copy the LICENSE file. Claude also suggested copying, so that's what we're doing, being pragmatic.

asubiotto (approved):

PR #5792, opened 2025-07-09
ui: Rename icicle graph to be flame graph

yomete:
It has been clear on multiple occasions that users and potential users are confused by the current nomenclature of "icicle" graphs and "icicle" charts. This PR renames all mentions of "icicle" graphs and charts to "flame" graphs and charts, to be in accordance with industry standard naming. 

~Please note when reviewing this PR that the URL state referencing the flame graph in the dashboard has been changed from 'icicle' to 'flame' as well. If users had previously saved URLs, this may cause them to break. If this is deemed an issue, we should add some logic so that 'icicle' and 'flame' are synonymous in the URL state.~

We added logic to redirect previous links with `icicle` to be automatically `flame` which ensures we have backwards compatibility. This also works with `iciclechart` and `flamechart`.


metalmatze (approved):
Love it. The wording is so much cleaner now!

PR #5797, opened 2025-07-10
wip: Multiple filters feature

manojVivek:
Screencast: 

https://github.com/user-attachments/assets/0208db95-52b2-4dc8-96db-06823981166d

TODO
- [x] Fix the scroll position change when the filters are submitted
- [x] Fix the bug where enter key submits reset
- [x] Fix the right click exclude binary to work
- [x] Add form validation to not submit when the input is empty
- [x] Fix the broken group by dropdown


pkg/query/columnquery.go
```diff
@@ -368,15 +343,31 @@ func (q *ColumnQueryAPI) Query(ctx context.Context, req *pb.QueryRequest) (*pb.Q
...
+								Condition: &pb.StringCondition_Contains{
+									Contains: sandwichByFunction,
+								},
+							},
+						},
+					},
+				},
+			},
+		}
+		sandwichFilters := []*pb.Filter{sandwichFilter}
```
brancz, line 362: As a user I think I would expect sandwich filters to be the defined filters + the focused frame instead of overriding all of them.

brancz (approved):
I suspect we might have to do some follow ups for perf improvements, but we should probably only do any based on profiling data.

My guess is that we're going to want to create a `Filter` interface and build the filters once and then just execute all filters in the appropriate place so we don't need to do the if statements.

metalmatze (commented):
Incredible work already. Really only nit picks.

brancz (approved):

PR #5829, opened 2025-07-21
Move the data fetching for the Sandwich view to be in the ProfileView

yomete:
I previously wrote the sandwich flame graphs’ data fetching logic in the `Sandwich` component itself, which is problematic upstream in [pprof.me](http://pprof.me/). We get a “cannot read properties of undefined reading “query” because it’s expecting a queryClient.

Pasted Graphic 4 copy

This PR moves the data fetching logic back to the `ProfileViewWithData` component which already is responsible for data fetching for our other visualisations.

### DIFF
--- a/ui/packages/shared/profile/src/Sandwich/index.tsx
+++ b/ui/packages/shared/profile/src/Sandwich/index.tsx
@@ -11,59 +11,31 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import React, {useEffect, useMemo, useRef, useState} from 'react';
+import React, {useMemo, useRef, useState} from 'react';
 
-import {type Row as TableRow} from '@tanstack/table-core';
-import {tableFromIPC} from 'apache-arrow';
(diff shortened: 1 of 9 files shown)


ui/packages/shared/profile/src/Sandwich/index.tsx
```diff
@@ -11,173 +11,42 @@
...
-      const row = rows.find(row => {
-        return row.name.trim() === sandwichFunctionName.trim();
-      });
-
-      if (row != null) {
-        setSelectedRow(row as unknown as TableRow<Row>);
-      }
-    }
-  }, [sandwichFunctionName, rows, selectedRow]);
+  const callersFlamegraphData = useMemo(() => sandwichData.callers, [sandwichData.callers]);
```
manojVivek, line 48: I think these memos aren't going to do anything good and can be skipped.

manojVivek (approved):
Just a minor nit, looking good otherwise.

PR #5830, opened 2025-07-21
Flamegraph perf

yomete:
This PR implements a couple of performance improvements to our flame graph rendering.

We do viewport culling where the idea here is to not render rows that are not visible to users so as to speed up rendering. We do this on multiple levels i.e. vertically and horizontally.

We cull vertically and efficiently by grouping the rows (by their depth level) into buckets and this is done by pre-computing buckets i.e. we don't scan all rows, which results in much faster lookups.

Horizontal culling is done by checking horizontal bounds and only iterating through rows visible horizontally in the viewport.

Before
[Link to Profiler](https://share.firefox.dev/44Z8WfE)
image

After
[Link to Profiler](https://share.firefox.dev/46Yi4Ua)
image

As seen in the profiles shared above, the flame graph rendering time reduced from 8 seconds to 91ms by culling from 7,951 rendered nodes to just 91 visible nodes.

### DIFF
--- a/ui/packages/shared/profile/src/ProfileFlameGraph/FlameGraphArrow/useVisibleNodes.ts
+++ b/ui/packages/shared/profile/src/ProfileFlameGraph/FlameGraphArrow/useVisibleNodes.ts
@@ -1,0 +1,170 @@
+// Copyright 2022 The Parca Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {useMemo, useRef} from 'react';
+
+import {Table} from 'apache-arrow';
+
+import {RowHeight} from './FlameGraphNodes';
+import {FIELD_CUMULATIVE, FIELD_DEPTH, FIELD_VALUE_OFFSET} from './index';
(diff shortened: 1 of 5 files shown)


ui/packages/shared/profile/src/ProfileFlameGraph/FlameGraphArrow/useVisibleNodes.ts
```diff
@@ -0,0 +1,170 @@
...
+
+    // Find max depth
+    let maxDepth = 0;
+    for (let i = 0; i < table.numRows; i++) {
+      const depth = depthColumn.get(i) ?? 0;
+      if (depth > maxDepth) maxDepth = depth;
+    }
+
+    // Create buckets for each depth level
+    const buckets: number[][] = Array.from({length: maxDepth + 1}, () => []);
```
brancz, line 44: this worries me a little, I wonder if this grouping could be done on the backend maybe? happy to try this though, maybe it's an unwarranted worry

ui/packages/shared/profile/src/ProfileFlameGraph/FlameGraphArrow/useVisibleNodes.ts
```diff
@@ -0,0 +1,170 @@
...
+    if (table === undefined) return [];
+
+    const depthColumn = table.getChild(FIELD_DEPTH);
+    if (depthColumn === null) return [];
+
+    // Find max depth
+    let maxDepth = 0;
+    for (let i = 0; i < table.numRows; i++) {
+      const depth = depthColumn.get(i) ?? 0;
+      if (depth > maxDepth) maxDepth = depth;
```
brancz, line 40: I feel like we've had this be a performance issue previously, no? I think we wrote an optimized function to determine this.

yomete: yes we did, let me rewrite this.

brancz (approved):

PR #5860, opened 2025-07-31
Fix objstore operation duration metrics by upgrading thanos-io/objstore

secfree:
### Problem

The `objstore_bucket_operation_duration_seconds` metric was reporting extremely high and incorrect values, indicating severe performance issues that don't reflect actual operation times.

Before the fix:

```
objstore_bucket_operation_duration_seconds_sum{bucket="xxx",operation="get"} 1.0440258375572279e+10
objstore_bucket_operation_duration_seconds_count{bucket="xxx",operation="get"} 2068
```
- Average operation time: ~5,048,480 seconds (~58 days per operation)
- This clearly incorrect metric made it impossible to monitor actual object storage performance

### Root Cause

The issue was caused by a bug in the `github.com/thanos-io/objstore` library's metrics collection logic, where operation durations were being incorrectly calculated or accumulated.

### Solution

Updated `github.com/thanos-io/objstore` dependency from `v0.0.0-20240913165201-fd105025a2e5` to `v0.0.0-20241028145108-31c0873d5266` to include the fix from [objstore commit 31c0873d](https://github.com/thanos-io/objstore/commit/31c0873d5266a6c58fe8fb5d6de01b66dcd50f5f).

Code changes required:
- Updated dependency version in `go.mod` and `go.sum`
- Updated `client.NewBucket()` calls to accommodate API changes by adding the new fourth parameter (set to `nil`)

### Verification

After the fix:

```
objstore_bucket_operation_duration_seconds_sum{bucket="xxx",operation="get"} 318.43996360699987
objstore_bucket_operation_duration_seconds_count{bucket="xxx",operation="get"} 541
```
- Average operation time: ~0.59 seconds per operation
- Metrics now accurately reflect actual object storage performance

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -39,7 +39,7 @@ require (
 	github.com/prometheus/common v0.65.0
 	github.com/prometheus/prometheus v0.55.1
 	github.com/stretchr/testify v1.10.0
-	github.com/thanos-io/objstore v0.0.0-20240913165201-fd105025a2e5
+	github.com/thanos-io/objstore v0.0.0-20241028145108-31c0873d5266
 	github.com/zeebo/xxh3 v1.0.2
 	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.62.0
 	go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace v0.62.0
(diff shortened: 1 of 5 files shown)


secfree:
@brancz  can you also take a look at this one? Thank you.

PR #5872, opened 2025-08-14
ui: Add read-only mode to ProfileFilters component

yomete:
Introduces a readOnly prop to the ProfileFilters component, disabling filter editing and hiding interactive controls when enabled.


manojVivek (approved):
Lgtm!

PR #5900, opened 2025-08-21
ui: Added test ids to some of the user intractable elements in the UI

manojVivek:
This PR adds test ids to the user interacting elements in the UI that will be useful for UI automated testing.


yomete (approved):
lgtm!

PR #5921, opened 2025-08-26
feat(parcacol): use distinct for label discovery to reduce rows returned

secfree:
Replace `Project` with `Distinct` in Querier.Labels. This emits only unique label columns, significantly reducing the number of rows materialized and returned.

In my cluster, the query of "Labels" returns more than 100 million rows from the underlying database. This PR reduced it to less than 50,000.


secfree:
@brancz could you help review this PR? Thank you.

brancz:
thanks for the ping, I was on PTO and this pushed it to the top of my inbox

PR #5927, opened 2025-08-28
ui: Added test ids for more components

manojVivek:


manojVivek:
merging to try this, please add any comments and will address them.

PR #5935, opened 2025-09-02
ui: Fixed the time-range selection bug during the click-n-drag interaction

manojVivek:
The latest nanoseconds migration missed handling this one, the click and drag on metrics graph was setting the merged_from and merged_to in milliseconds which made the flame graph query to return "Profile has no samples" error.


yomete (approved):
lgtm

PR #5939, opened 2025-09-03
Refactors ProfileSelector and MetricsGraphSection to remove utilization-specific props and logic

yomete:
This PR refactors `ProfileSelector` and `MetricsGraphSection` to remove utilization-specific props and logic, consolidating query controls into `SharedQueryControls`.

`SharedQueryControls` is a reusable component that contains the Profile type selector, label matchers, sum by selector, time range picker, and search button. All the afore-mentioned components can then be customized to be displayed or not based on `props` passed.

A new `UtilizationMetricsSelector` component is now responsible for rendering the several utilization metric graphs and it works by using the new  `SharedQueryControls` to display only query controls that makes sense to the utilization metrics.

Also replaced the `UtilizationLabelsContext` with `UtilizationContext` which now handles the metrics and labels.


ui/packages/shared/profile/src/contexts/MatchersInputLabelsContext.tsx
```diff
@@ -54,7 +54,7 @@ export function LabelsProvider({
   end,
 }: LabelsProviderProps): JSX.Element {
   const [currentLabelName, setCurrentLabelName] = React.useState<string | null>(null);
-  const utilizationLabels = useUtilizationLabels();
+  const utilizationContext = useUtilization();
```
manojVivek, line 57: Maybe we should try an approach where we don't need to deal with both the labels and determine which to use here. 
IMO if there is a way we can make this generic to take only the labelNames and fetchLabelsNames handler  and not necessary to know the type of labels that would be great.

manojVivek (commented):
Overall definitely great improvement over what we have now, just some discussion points to see if there is a way we can make things more generic and remove all the utilization related stuff from Parca.

ui/packages/shared/profile/src/contexts/UtilizationContext.tsx
```diff
@@ -0,0 +1,73 @@
...
+  children: ReactNode;
+  value: UtilizationProps | undefined;
+}
+
+// The UtilizationContext is used to store the utilization label names and values, metrics and loading state. It also
+// contains the function utilizationFetchLabelValues to fetch the utilization label values.
+// This context was created so as to avoid props drilling.
+const UtilizationContext = createContext<UtilizationProps | undefined>(undefined);
+
+export function UtilizationProvider({children, value}: UtilizationProviderProps): JSX.Element {
```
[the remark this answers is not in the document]
yomete: so basically use these hooks to make a query against the upstream APIs to fetch the data needed?

ui/packages/shared/profile/src/contexts/MatchersInputLabelsContext.tsx
yomete: i think i have an idea for this🤔 i'll try it out and let you know

yomete:
Closed in favor of https://github.com/parca-dev/parca/pull/6025

PR #5943, opened 2025-09-08
ui: Compare mode test ids

manojVivek:


manojVivek:
Merging to try this, please add any comments and will be addressed

PR #5953, opened 2025-09-15
ui: Ensure Flame Graph is reset on all toolbar interactions

yomete:
This PR ensures that we now reset the Flame Graph on Filters and Level of Profiling changes. I thought it'd be a good idea to rename the `useResetStateOnNewSearch` hook to `useResetFlameGraphState`  since that reflects/explains more what happens in the hook.

### DIFF
--- a/ui/packages/app/web/CHANGELOG.md
+++ b/ui/packages/app/web/CHANGELOG.md
@@ -3,6 +3,12 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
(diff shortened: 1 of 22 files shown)


ui/packages/shared/profile/src/ProfileView/components/DashboardItems/index.tsx
```diff
@@ -42,8 +42,6 @@ interface GetDashboardItemProps {
   profileSource: ProfileSource;
   total: bigint;
   filtered: bigint;
-  curPath: string[];
```
yomete, line 45: Doing some cleanups here as we no longer need these props.

PR #5959, opened 2025-09-17
feat: use Distinct instead of Project for profile metadata queries

secfree:
Replace `Project()` with `Distinct()` in `GetProfileMetadataMappings` and 
`GetProfileMetadataLabels` methods to ensure unique results and improve 
query performance by eliminating duplicate stacktrace and label entries.

This change will not affect correctness as GetProfileMetadataMappings
and GetProfileMetadataLabels also have deduplicate logic.

This change helps reduce 100+ million rows from the underlying database, just like #5921


secfree:
"make go/lint" failed with "arrow.Record is deprecated" is not related to this change,
as the "main" branch also has the same issue.

brancz (approved):

PR #5963, opened 2025-09-24
ui: Cancel previous request when a new query starts

yomete:
PR uses React Query to cancel previous request if the request is exactly the same. Works for the Query, QueryRange and Labels requests.


manojVivek (approved):
Lgtm!

PR #5964, opened 2025-09-25
Add test IDs for Invert Call Stack button

yomete:
Also ensures flame graph state resets when group by labels are updated. (This is an actual bug detected by the E2E tests)


manojVivek (approved):
Lgtm!

PR #5987, opened 2025-10-07
Add Prometheus metrics for forwarded profile bytes in GRPCForwarder

metalmatze:
Works nicely. 
image


manojVivek (approved):
Lgtm!

metalmatze:
It turns out that we actually parse the profile with the scraper before forwarding. Closing this in favor of #5988

PR #5989, opened 2025-10-08
ui: Improve label name and values UI

yomete:
This PR contains improvements to how the `SimpleMatchers` and `MatchersInput` components work with label values.  It also contains logic to persist the "Color by" and "Left/Right align function names" user preferences.

For the `SimpleMatchers` and `MatchersInput` components:
- Stopped aggressively hiding label name and values when editing more than 3 label names and values in the `SimpleMatcher` component. Now when editing the label rows and you're working with a lot, you see all the label rows without having to always click on `Show more/Show less` all the time.
- Added a loading indicator to the label value selector.
- If a label name and value is selected we now show it instantly, regardless of loading or not loading the values from the API.
- Increased the max width of dropdown when the dropdown is open so as to see the full value (for most cases, extremely long values will still get line breaks)
- Reduced staleTime for label value requests to 2 minutes (previously 5 minutes).
- Added a "Refresh results" button to fetch new label values.

### DIFF
--- a/pkg/normalizer/normalizer.go
+++ b/pkg/normalizer/normalizer.go
@@ -512,6 +512,7 @@ func NormalizeWriteRawRequest(ctx context.Context, req *profilestorepb.WriteRawR
 				continue
 			}
 
+			//nolint:staticcheck // SA1019: Update when we actually use the latest Prometheus
 			if valid := model.LabelName(l.Name).IsValid(); !valid {
 				return NormalizedWriteRawRequest{}, status.Errorf(codes.InvalidArgument, "invalid label name: %v", l.Name)
 			}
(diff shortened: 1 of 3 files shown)


ui/packages/shared/profile/src/MatchersInput/index.tsx
```diff
@@ -115,12 +116,18 @@ export const useLabelValues = (
         profileType !== '' &&
         labelName !== undefined &&
         labelName !== '',
-      staleTime: 1000 * 60 * 5, // 5 minutes
+      staleTime: 1000 * 60 * 2, // 2 minutes
```
brancz, line 119: Why do we need staleness at all? Shouldn’t the timeframe and a success already be sufficient?

note the important thing is that we fetch label names and values based on the selected time (even before hitting search)

brancz (approved):

PR #6025, opened 2025-10-27
Refactor ProfileSelector and MetricsGraphSection components

yomete:
This PR refactors `ProfileSelector` and `MetricsGraphSection` to remove utilization-specific props and logic

There's also a new `LabelsQueryProvider` context that is responsible for fetching labels data from the Parca API
and making it available to child components and a `UnifiedLabelsProvider` context that consolidates raw label arrays into UI-friendly formats and passes it down to components like `SimpleMatchers` and `MatchersInput`

### DIFF
--- a/ui/packages/shared/profile/src/contexts/UnifiedLabelsContext.tsx
+++ b/ui/packages/shared/profile/src/contexts/UnifiedLabelsContext.tsx
@@ -57,6 +57,8 @@ interface UnifiedLabelsContextType {
   refetchLabelValues: () => Promise<void>;
   refetchLabelNames: () => Promise<void>;
   labelNameFromMatchers: string[];
+
+  suffix?: '_a' | '_b';
 }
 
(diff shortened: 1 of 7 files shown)


ui/packages/shared/profile/src/contexts/UnifiedLabelsContext.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+interface UnifiedLabelsProviderProps {
+  children: React.ReactNode;
+
+  queryClient: QueryServiceClient;
+  setMatchersString: (arg: string) => void;
+  runQuery: () => void;
+  currentQuery: Query;
+  profileType: string;
+  start?: number;
```
manojVivek: Same for these too, why not try getting these directly from the `useQueryState` hook?

yomete: nice, must have missed that useQueryState returns these now. I'll update this

PR #6028, opened 2025-10-29
Fix root cumulative value in flamegraph trimming

yomete:
Updates the flamegraph trimming logic to correctly set the root row's cumulative value as the sum of its direct children's cumulative values.

### DIFF
--- a/pkg/query/flamegraph_arrow.go
+++ b/pkg/query/flamegraph_arrow.go
@@ -357,7 +357,7 @@ func generateFlamegraphArrowRecord(ctx context.Context, mem memory.Allocator, tr
 		fb.trimmedLocationLine.AppendNull()
 		fb.trimmedFunctionStartLine = array.NewUint8Builder(fb.pool)
 		fb.trimmedFunctionStartLine.AppendNull()
-		fb.trimmedCumulative = array.NewUint8Builder(fb.pool)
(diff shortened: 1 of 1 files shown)


pkg/query/flamegraph_arrow_test.go
```diff
@@ -931,6 +931,86 @@ func TestGenerateFlamegraphArrowTrimming(t *testing.T) {
...
+	fa, cumulative, _, _, err := generateFlamegraphArrowRecord(ctx, mem, tracer, p, []string{FlamegraphFieldFunctionName}, float32(0.5))
+	require.NoError(t, err)
+	defer fa.Release()
+
+	cumulativeCol := fa.Column(fa.Schema().FieldIndices("cumulative")[0]).(*array.Int64)
+	rootCumulative := cumulativeCol.Value(0)
+
+	require.Equal(t, cumulative, rootCumulative, "Returned cumulative should match root row cumulative value")
+
+	require.Equal(t, int64(100), rootCumulative, "Root cumulative should be 100 (sum of direct child func1)")
```
metalmatze: lovely

brancz (approved):

metalmatze (commented):
This can be simpliefied. Seems like Claude made things too complicated not knowing about rootCumulativeTrimmedValue's semantics. Surprise. 😄

PR #6033, opened 2025-11-06
ui: Query selector state refactoring and batchUpdates support in URLState

manojVivek:
Contains:
- [x] Moved all profile selector state into useQueryState hook that uses `useURLState` hook for all URL manipulations instead of accessing the query params directly. 
- [x] Inorder to support the above changes, we needed `batchUpdates` in the URL State to push multiple params into the URL without creating any race conditions
- [x] Show nicer Profile type name for the value from the URL while the data is being fetched.


ui/packages/shared/components/src/hooks/URLState/index.test.tsx
metalmatze: Absolutely love the thorough tests! 💯

PR #6061, opened 2025-11-24
api: Validate the profile type before ingestion

manojVivek:


brancz (commented):
I like the “parse don’t validate” approach!

PR #6074, opened 2025-11-27
ui: Fix focusing on series in the metrics graph

yomete:
Using the right click menu to focus on a series only added the label matchers but doesn't run the query.

This fixes it by modifying the `MetricsGraphSection` and `ProfileSelector `to use the `commitDraft` function for applying label matcher filters essentially triggering a query search.

### DIFF
--- a/ui/packages/shared/profile/src/hooks/useQueryState.ts
+++ b/ui/packages/shared/profile/src/hooks/useQueryState.ts
@@ -38,7 +38,7 @@ interface UseQueryStateReturn {
   draftSelection: QuerySelection;
 
   // Draft setters (update local state only)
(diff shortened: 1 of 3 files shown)


ui/packages/shared/profile/src/hooks/useQueryState.ts
```diff
@@ -38,7 +38,7 @@ interface UseQueryStateReturn {
   draftSelection: QuerySelection;
 
   // Draft setters (update local state only)
-  setDraftExpression: (expression: string, commit?: boolean) => void;
+  setDraftExpression: (expression: string) => void;
```
yomete: we never use this `commit` boolean anywhere so removing it

metalmatze (approved):

PR #6075, opened 2025-11-27
ui: Add group_by param to default explorer values

yomete:
This ensures that when the flame graph is grouped by custom labels and there's a page reload, we no longer get the error below. 

image

The change simply ensures that the `group_by` param is treated as an array and not a string.


manojVivek (approved):
Lgtm!

PR #6087, opened 2025-12-04
ui: Flamechart improvements

yomete:
This pull request introduces an auto-configuration feature for the flame chart, providing users with a one-click way to optimize their view and an explanatory popover when auto-configuration occurs. 

**Flame chart auto-configuration and popover:**

- Added a new user preference `FLAMECHART_AUTO_CONFIG_POPOVER_DISMISSED` to allow users to dismiss the auto-configuration popover and remember their choice.
- Introduced a `FlameChartAutoConfigPopover` component that explains which settings were auto-configured for optimal viewing and allows users to dismiss it.
- Implemented the `useAutoConfigureFlamechart` hook, which updates URL state to set optimal flame chart parameters (time range, sum-by labels, etc.) and signals when auto-configuration occurs.

image

### DIFF
--- a/pkg/query/columnquery.go
+++ b/pkg/query/columnquery.go
@@ -982,33 +982,92 @@ func matchesFrameFilter(r *profile.RecordReader, locationIndex, lineIndex int, f
 	return true
 }
 
+// toLower converts ASCII byte to lowercase without allocation.
+// Non-ASCII bytes are left unchanged.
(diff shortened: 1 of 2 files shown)


brancz:
interesting .. should we maybe make flamecharts a top-level tab? I find the UX of the visualization selection changing the previous, further up selections quite unexpected

PR #6098, opened 2025-12-09
ui: Hide context menu after sandwich action

yomete:
Added calls to hideMenu() after setting the sandwich function name, ensuring the context menu closes immediately when the sandwich action is triggered.


yomete:
Merging this in for upstream usage.

PR #6128, opened 2025-12-25
Fix cumulative value calculation for recursive functions in Table view

secfree:
## Problem

In the Table view, recursive functions (functions that appear multiple times in the same call stack) may display Cumulative (%) values greater than 100%. For example (tested with parca 0.25.0):

image

## Root Cause

When building the table, the code iterates through each frame in a stacktrace and adds the sample value to that function's cumulative. When a function appears multiple times in the same stacktrace (due to recursion), its cumulative value gets counted multiple times.

Example:

```
Stacktrace: [A → B → A → D], value = 50

Before fix:
  Frame A: cumulative += 50  → 50
  Frame B: cumulative += 50  → 50
  Frame A: cumulative += 50  → 100  // Counted twice!
  Frame D: cumulative += 50  → 50

Function A Cumulative (%) = 100/50 = 200%  // Exceeds 100%
```

## Solution

Track which table rows have been counted for each sample using a `seenInSample` map. Only add to cumulative for the first occurrence of a function in each sample, while still maintaining caller/callee relationships for all occurrences.

```
After fix:
  Frame A: first seen, +50     → cumulative = 50
  Frame B: first seen, +50     → cumulative = 50
  Frame A: already seen, SKIP  → cumulative = 50 (unchanged)
  Frame D: first seen, +50     → cumulative = 50

Function A Cumulative (%) = 50/50 = 100% 
```

### DIFF
--- a/pkg/query/table.go
+++ b/pkg/query/table.go
@@ -135,6 +135,10 @@ func generateTableArrowRecord(
 
 		for sampleRow := 0; sampleRow < int(r.Record.NumRows()); sampleRow++ {
 			previousTableRow := -1
+			// Track which table rows have been counted for this sample to avoid
+			// double-counting cumulative values for recursive functions.
+			seenInSample := make(map[int]struct{})
+
 			lOffsetStart, lOffsetEnd := r.Locations.ValueOffsets(sampleRow)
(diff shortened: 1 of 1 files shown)


secfree:
@brancz can you help take a look? Thank you.

PR #6149, opened 2026-01-12
Update Go language version and frostdb to latest

manojVivek:
I was trying to update `frostdb` to the latest and that needed go lang version to be `1.25` so did that as well.


manojVivek:
Taken care in #6156

PR #6153, opened 2026-01-15
Refactor source report to use sparse line metrics

yomete:
The source report will no longer return an error if the source file is not present and we also added explicit line numbers to the Arrow schema that is returned. Now, only lines with actual profiling data are included in the response, rather than returning an array with one entry per line in the source file.

Changes:
Backend returns empty source field instead of error when source file is missing
Arrow schema now includes line_number column with explicit line numbers
Response only contains rows for lines with non-zero cumulative or flat values


pkg/query/sources.go
```diff
@@ -88,64 +88,80 @@ func generateSourceReportRecord(
...
+	defer lineNumArr.Release()
+	cumuArr := cumuBuilder.NewInt64Array()
+	defer cumuArr.Release()
+	flatArr := flatBuilder.NewInt64Array()
+	defer flatArr.Release()
 
 	return array.NewRecordBatch(
 		arrow.NewSchema(
 			[]arrow.Field{
+				{Name: "line_number", Type: arrow.PrimitiveTypes.Int64},
```
yomete: Add the line number as a column in the Arrow schema

PR #6326, opened 2026-04-13
fix(ui): self-host Google Fonts to eliminate third-party dependency

yomete:
Replaces the runtime `@import` of Google Fonts with self-hosted woff2 files served from `public/fonts/`. 

Relates to #6292


brancz:
Nice

PR #6360, opened 2026-05-14
ui: Cleanup the unused navigateTo props

manojVivek:


yomete (approved):
linter caught an error but otherwise lgtm

PR #6362, opened 2026-05-21
Update GPU profile types for PC sampling

gnurizen:


ui/packages/shared/profile/src/ProfileTypeSelector/index.tsx
```diff
@@ -111,6 +111,18 @@ export const wellKnownProfiles: WellKnownProfiles = {
     name: 'On-GPU',
     help: 'Time spent on the GPU.',
   },
+  'parca_agent:gpu_time:nanoseconds:gpu_time:nanoseconds:delta': {
+    name: 'On-GPU',
+    help: 'Time spent on the GPU. Combines kernel execution time and PC stall samples; the gpu_view label distinguishes the two views.',
+  },
+  'parca_agent:gpu_kernel_time:nanoseconds:gpu_kernel_time:nanoseconds:delta': {
+    name: 'On-GPU',
```
metalmatze, line 119: Maybe something like this here? 
```suggestion
    name: 'GPU Kernel',
```
Otherwise we have the same On-GPU twice.

metalmatze (approved):
🎉 

Feel free to adjust or ignore my comment.

ui/packages/shared/profile/src/ProfileTypeSelector/index.tsx
gnurizen: So currently you get one or the other depending on a parca-agent flag, gpu_time is this merged profile type that shows pc samples and gpu kernel times together.   gpu_kernel_time is the default where its just kernels and pc samples are in the gpu_stall_time profile.

brancz (approved):

ui/packages/shared/profile/src/ProfileTypeSelector/index.tsx
metalmatze: I see!

PR #6376, opened 2026-07-19
Fix pprof validation: nil period_type and off-by-one string-table bounds

arpitjain099:
While reading the pprof ingest path I hit two panics in ValidatePprofProfile (pkg/normalizer/validate.go). This validator runs on the WriteRaw gRPC ingest that every agent uses to push profiles, so the profile bytes are attacker-influenced, and both cases crash the validator (or the code just past it) rather than returning an error.

The first is a nil pointer deref. period_type is optional, so a profile that omits it arrives with a nil PeriodType. The rest of the function nil-checks its pointer fields, but the period_type branch dereferences p.PeriodType.Type directly, so merely leaving the field out panics inside the validator. I wrapped those two checks in a nil guard like everything else.

The second is an off-by-one in the string-table bounds checks. They use `> len(StringTable)`, but string indices are 0-based, so the valid range is 0..len-1. An index exactly equal to len slips through the check and is one past the end, and downstream code (MetaFromPprof, LabelsFromSample) then indexes it and panics with index out of range. I changed the eleven 0-based string-index checks to `>=`. I left the mapping/function/location id checks as `>` on purpose, since those ids are 1-based and len is a valid id.

Added a table-driven test in the normalizer package: an omitted period_type, and an index equal to the string-table length on both the sample-type and period-type paths, plus a valid control. The bad cases panic (or are wrongly accepted) before the change and return a clean error or validate fine after it. `go test ./pkg/normalizer/...` passes.

Happy to sign the CLA. Let me know if you would rather split the two fixes or send this against a release branch.

### DIFF
--- a/pkg/normalizer/validate.go
+++ b/pkg/normalizer/validate.go
@@ -37,10 +37,10 @@ func ValidatePprofProfile(p *pprofpb.Profile, ei []*profilestorepb.ExecutableInf
 		if m.Id != uint64(i+1) {
 			return fmt.Errorf("mapping id is not sequential")
 		}
-		if m.Filename != 0 && m.Filename > stringTableLen {
+		if m.Filename != 0 && m.Filename >= stringTableLen {
 			return fmt.Errorf("mapping (id: %d) has invalid filename index %d", m.Id, m.Filename)
 		}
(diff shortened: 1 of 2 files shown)


brancz (approved):

brancz:
Thank you very much!

PR #6401, opened 2026-09-01
Support compressed pprof scrape responses

alexon1234:
## Summary

- Decode gzip, LZ4, and Zstandard pprof responses at the scrape boundary.
- Preserve raw pprof support.
- Bound decompressed profile size to 100 MiB.
- Add focused format coverage.

This allows clients such as ddtrace 3.x, whose file exporter emits LZ4-framed pprof data, to be scraped directly by Parca without an application-side LZ4-to-gzip conversion.

## Test plan

- [x] `go test ./pkg/scrape`
- [x] `gofmt` on changed Go files


brancz (approved):
Very nice! Thank you!

PR #6403, opened 2026-09-18
pkg/query: handle null pprof metadata

tsenart:
## Problem

The pprof writer assumes that mapping and function dictionary fields are always populated.

- A null mapping filename or build ID indexes an empty Arrow dictionary and can panic with `index out of range [0] with length 0`.
- A null function name returns function ID `0`, but the writer still emits a line that refers to that ID. The resulting pprof is invalid.
- A function with one missing field loses other available function metadata.

We found these cases through Polar Signals Cloud's merged pprof export. Native gRPC and grpc-web showed the same results. The input profiles are valid before the Arrow round trip.

## Change

Read each optional mapping and function field only when it is present. Create a function from any available function metadata. When no function metadata exists, keep the location and address but omit the line, because pprof cannot represent a line without a function.

The table test covers each nullable mapping and function field with both an empty dictionary and a populated peer row. It checks sample values, labels, addresses, mappings, lines, and available symbols after serialization and parsing.

## Validation

- `go test ./pkg/profile ./pkg/query -count=1`
- `go vet ./pkg/query`
- `make go/lint` (`0 issues`)
- `git diff --check`

All pass.


tsenart:
@parca-dev/backend-maintainers Could a maintainer approve the fork workflows and review this fix? The local package tests, `go vet`, and `make go/lint` pass. The change comes from two reproducible failures in Polar Signals Cloud's merged pprof export.

metalmatze (approved):
Nice! The changes make sense if you ran into this.

